PR #7, opened 2019-11-05
feat: moved helper functions from commercetools-docs

davifantasia:


emmenko:
What we can do though is to move the helpers over here once we actually need them, so that we don't end up with unused functions. Does that make sense to you?

packages/gatsby-theme-docs/src/utils/helpers/generic.js
```diff
@@ -0,0 +1,40 @@
+import unified from 'unified';
+import markdown from 'remark-parse';
```
emmenko: Those 2 deps above do not seem to be defined in the package.json

packages/gatsby-theme-docs/src/utils/helpers/compute-path-with-path-prefix.js
```diff
@@ -0,0 +1,8 @@
+import { withPrefix } from 'gatsby-link';
```
emmenko:
```suggestion
import { withPrefix } from 'gatsby';
```

packages/gatsby-theme-docs/src/utils/helpers/compute-path-with-path-prefix.js
```diff
@@ -0,0 +1,8 @@
...
+
+export default function computePathWithPathPrefix(path) {
```
emmenko: I wondering if we really need this, I would expect to just do:

```js
withPrefix(path)
```

Then there is no reason for this helper function. Can you maybe explain what problems is this trying to solve and where is it supposed to be used?

packages/gatsby-theme-docs/src/utils/helpers/generic.js
```diff
@@ -0,0 +1,40 @@
...
+export const markdown2React = markdownString => {
+  const result = unified()
+    .use(markdown, { commonmark: true })
+    .use(remark2react)
+    .processSync(markdownString).contents;
+
+  return result;
+};
+
+export const computeType = value => {
```
emmenko: Can you explain what does this function do? Where is this supposed to be used?

packages/gatsby-theme-docs/src/utils/helpers/generic.js
```diff
@@ -0,0 +1,40 @@
...
+  const regex = /^\[object (\S+?)\]$/;
+  const matches = Object.prototype.toString.call(value).match(regex) || [];
+
+  return (matches[1] || 'undefined').toLowerCase();
+};
+
+export const generateAnchorString = str => {
```
emmenko: Same here, where is this supposed to be used?

packages/gatsby-theme-docs/src/utils/helpers/generic.js
```diff
@@ -0,0 +1,40 @@
...
+        .replace(/[^a-zA-Z0-9 ]+/g, '')
+        // replace all spaces with hyphen
+        .replace(/\s+/g, '-')
+    );
+  }
+
+  return '';
+};
+
+export const capitalizeFirstCharacter = str => {
```
emmenko: FYI: a one line version of this can be as following:

```js
return value.replace(/^\w/, char => char.toUpperCase());
```

packages/gatsby-theme-docs/src/utils/helpers/index.js
```diff
@@ -0,0 +1,10 @@
+import computePathWithPathPrefixInIndex from './compute-path-with-path-prefix';
+
+export {
+  markdown2React,
+  computeType,
+  generateAnchorString,
+  capitalizeFirstCharacter,
+} from './generic';
+
+export const computePathWithPathPrefix = computePathWithPathPrefixInIndex;
```
emmenko: Suggestion: in general I would move each helper function into its own file and have them all under `utils`, there is no real need to nest them into a subfolder or into the same file.

packages/gatsby-theme-docs/src/utils/helpers/generic.js
davifantasia: It's basically an efficient JavaScript type checker.

Please read about it here:

https://blog.logrocket.com/javascript-typeof-2511d53a1a62/

packages/gatsby-theme-docs/src/utils/helpers/compute-path-with-path-prefix.js
davifantasia: Each microsite was generated with path prefixes, e.g., for our test site the url would be:

`http://docs.commercetools.com/test`

Writers would write a path, say to getting started like so:

`/getting-started/`

Although GatsbyLink should generate the path prefix it did not work for some reason so we ended up with a link like `http://docs.commercetools.com/getting-started/`

Hence this function.

packages/gatsby-theme-docs/src/utils/helpers/compute-path-with-path-prefix.js
davifantasia: I'm not sure why the suggestion if it works as is.

packages/gatsby-theme-docs/src/utils/helpers/generic.js
davifantasia: Used to generate an ID for the page title.

The right/index nav's title is expected to be clickable and the page view should jump to the title - does this make sense?

packages/gatsby-theme-docs/src/utils/helpers/index.js
davifantasia: I think that makes sense, but so that the move can be as smooth as possible I suggest this to be done later.

packages/gatsby-theme-docs/src/utils/helpers/generic.js
emmenko: Sure but this already works, so I'm not sure where this function is supposed to be used 🤔

packages/gatsby-theme-docs/src/utils/helpers/generic.js
emmenko: Thanks for the link. In those cases please always add an explanation as comment next to the function and/or link to some resources where the implementation is take from.

However, I still don't understand where is this supposed to be used 😅

packages/gatsby-theme-docs/src/utils/helpers/compute-path-with-path-prefix.js
emmenko: Sure, but the signature of the `withPrefix` method accepts a path

image

This custom helper function is basically doing the same thing, that's why I'm asking why do we need it in the first place and not just use `withPrefix(path)` 🤔

packages/gatsby-theme-docs/src/utils/helpers/index.js
emmenko: Sorry but what's the problem with renaming a couple of files now?

PR #14, opened 2019-11-05
feat(theme): lint gatsby graphql queries

emmenko:
Ref: https://www.gatsbyjs.org/packages/gatsby-plugin-extract-schema/?=gatsby-plugin-extract-schema

Now the graphql queries used within the gatsby codebase will be linted against the generated schema.

> NOTE that the generated schema is in version control. This is usually fine (we do that in other repos as well) as it avoid linters to break if you checkout the repo without starting the gatsby website.


emmenko:
The RAML plugin is a separate package, so we can just keep the schema validation for the base theme.
It's not a super important feature to have, we can drop it if you feel uncomfortable. However, it's a simple setup and it provides an extra layer of safety almost for free.

nkuehn:
I generally find the Idea very good because the kinds of errors it catches are super annoying because they can come up at build / run time and can depend on content changes. 

Just wanted to raise awareness that it implicitly means that any custom GraphQL node generation (which you end up doing a lot in Gatsby, it's the best practice for customizations, you try to do stuff in the data node layer and not in React. ) needs to declare its schema then.  That's good!

emmenko:
@davifantasia any more feedback from your side? If it's fine for you, can you approve?

PR #16, opened 2019-11-06
docs(contributing): document dev tools

emmenko:


nkuehn:
LGTM, I'll have to learn a few things but all good practice and if it's the de facto coding standard inside commercetools that's a good idea in any case to encourage adoption. 

Two questions:
 1. I saw that there's a reusable eslint definition in some frontend repository that you can pull as a dependency.   Directly use that? 
 1.  we did have branch protection configured in the existing repos to allow a broader set of users in commercetools to get write permissions but still have control over the actual releases.  Configure that on master here, too?

emmenko:
> I saw that there's a reusable eslint definition in some frontend repository that you can pull as a dependency. Directly use that?

You mean this? https://github.com/commercetools/eslint-config
If so, it's actually not being maintained at all 🙃😅

> we did have branch protection configured

Sure, we can set it up, it should not be a problem.

nkuehn:
Well, for stuff like style definitions it's rather a good sign if nobody touches them for a year, so I wasn't suspicious when I saw the "Last modified" dates :-) 

Branch protection is not urgent, let's just not forget it before giving a broader set of people in the company write permissions.

emmenko:
> Well, for stuff like style definitions it's rather a good sign if nobody touches them for a year

Well it's not much about the rules, but about the version of the dependencies and the combination of the presets being used.

> Branch protection is not urgent

It's already set up 😉

PR #17, opened 2019-11-06
feat: move over docs-smoke-test website

emmenko:
I renamed the folder/website to `docs-smoke-test`, as `test` is kind of confusing with test folder or such.


nkuehn:
(feedback one of two):

Can we choose a name that is neither "Playground" nor "test"?  "Playground communicates that any developer can and should use it to play around to try stuff.   

But the core purpose of the site is an end-to-end smoketest to check 
 a) it builds without errors and 
 b) the VRT is green. 

If we end up with multiple themes (base docs theme and the theme extension for api documentation) we'll actually have to have multiple separate smoketest sites, too that define whether a change to the respective theme is OK. 

So I vote for a convention that picks up the name of the theme

 *  `gatsby-theme-docs` -> `docs-smoketest-site` (or whatever other pre/postfix you can agree on)
 *  `gatsby-theme-api-docs` -> `api-docs-smoketest-site`  (..)

nkuehn:
(feedback two of two)

Just to understand:  Why are the test site(s) not monorepo packages on their own?  They are complete JS projects with package.json,  potentially own component overrides (to test that overriding actually works) etc. 

I think the setup here should be the same as it is in the `commercetools-docs` repo so we practice the exact same setup including how builds are configured etc.

emmenko:
> Can we choose a name that is neither "Playground" nor "test"? "Playground communicates that any developer can and should use it to play around to try stuff.

Fine by me

> the core purpose of the site is an end-to-end smoketest to check

Yes, but we also need it for developing the theme etc. It does not make sense to have yet another website for "development", it should just be the same thing.

> Why are the test site(s) not monorepo packages on their own?

I'd like differentiate between packages that will/can be published (so everything in `packages`) and packages that are used only within the monorepo (e.g. for development, etc).

We can maybe have:
- `packages`
- `websites`

Does that make sense to you?

emmenko:
Alright I moved now the website into `test-websites/docs-smoke-test`.

emmenko:
Alright, I'll merge this and setup Now as the next step. We can do further improvements/renaming as follow up.

PR #24, opened 2019-11-06
chore: setup cypress

emmenko:
Besides setting up Cypress for running e2e tests, we additionally take Percy snapshots on the pages we visit.

Taking snapshots is more of an optional thing, but we have the chance of doing that for e2e tests as well.


packages/gatsby-theme-docs/src/components/content-pagination.js
```diff
@@ -10,6 +10,8 @@ import Card from './card';
 import Spacings from './spacings';
 import TextSmall from './text-small';
 
+const trimTrailingSlash = url => url.replace(/(\/?)$/, '');
```
nkuehn: It seems to be rather an artifact in this PR but  my experience is that it's a bad idea to change gatsby's default behavior concerning the URL structure.   Gatsby is assuming  and generating trailing slash URLs in many situations so I think it's not ideal to try to work against its defaults.   
Using "real" folders  is also more interoperable with various hosting options because the real output of the static builds are folders with index.html files in them. 

Opinionated topic, I know..

emmenko: This is just to properly match the URL (slug) with the one defined manually in e.g. the `navigation.yaml`. Yes, Gatsby adds the trailing slash but the navigation links are maintained manually, which is error prone.

emmenko:
> Can you give a short explanation of why one more testing tool is necessary? I did understand you that you promoted react-testing-library for behavioral test and I am not sure about the role of cypress here.

The reason is that they address different concerns. To put in short, E2E tests run against the production build whereas RTL tests run in an isolated test environment.
Given that there are subtle differences between the dev/test and prod environment, we should test on both ends.

With RTL we can test parts of the pages/components, most of the time we still need to e.g. mock graphql queries. This is fine as it focuses more on the behavior of the components.
With E2E we can focus more on the general interactions, like navigations etc.

Hope it makes sense.

PS: the fact that we can also take Percy snapshots in Cypress is just a nice thing to have, it's not a must for having the E2E tests.

PR #25, opened 2019-11-06
feat: writing style linter with commercetools specific configuration

nkuehn:
Proposed package that allows to install a pre-packaged and preconfigured vale language style linter globally or into a writing project. 

I have not been able to fully test the experience "as if it were installed from NPM" but propose to start working with it to work out the practical details e.g. with the various Editor Plugins. 

It is usable for linting on CI as it it here and returns a non-zero return code on "errror" rule violations but not on warnings or suggestions.


packages/writing-style/README.md
```diff
@@ -0,0 +1,36 @@
...
+1.  Install the command globally via `npm install -g @commercetools-docs/writing-style`
+1.  Use it by calling `commercetools-vale ./path/to/your/content`
+
+In addition, you can install a matching editor plugin to get immediate feedback while authoring, for example [for VSCode](https://marketplace.visualstudio.com/items?itemName=testthedocs.vale) - don't use the `vale-server` one but the one for the command line.
+
+**If your plugin allows, reconfigure the command from `vale` to `commercetools-vale`**. Otherwise, for example in VSCode, you have to uninstall your global vale installation and create an alias via `echo "alias vale=commercetools-vale" >>~/.bash_profile` (or ending `.zsh` if you are on a fresh Mac OS X catalina) (no insurance provided, suggestions for a cleaner method welcome).
+
+Please don't forget to frequently update the package to get the latest styles.
+
+## Linter Usage: CI and command line in Javascript Projects
```
emmenko: I see 2 problems at the moment:
- the vale binaries that we ship with the package are not versioned, meaning that it's not clear which version of vale is running under the hood
- the vscode plugin does not really work at the moment as it assumes the binary is installed on the global `PATH`. So if we install the package in a repository, we also need to install it globally in order to use the editor plugin.
  I can look into the vscode plugin and open a PR to make the PATH configurable, so that we can point it to the workspace installation.

emmenko: This is what we should aim for in my opinion (regarding vscode): https://github.com/microsoft/vscode/blob/2aa183fdc89d93bbb592183925ec3c797fc46ac7/.vscode/settings.json#L40

nkuehn: concerning 1:   I can rename them to a versioned file name if that makes it more transparent. In any case you can just call  it with `--version` and get the version.  Not sure about the relevance though. 

concerning 2: Yes, I suspected that, too.  The plugin uses execFile to call vale, so the setting the shell environment does not work because execFile does not spawn a shell to call the binary. 
The call is hardwired here: https://github.com/testthedocs/vscode-vale/blob/master/src/extension.ts#L156   Ideal would be to be able to configure a different executable path per workspace (not even going through the PATH env).  I've never done VSCode plugins, would have to learn how to declare a configuration flag.

nkuehn: PS: I created an issue with a proposal:  https://github.com/testthedocs/vscode-vale/issues/7

emmenko: Thanks, I can also spin up a PR on their repo today. It should be pretty straightforward to add the setting.

nkuehn: I addressed both issues now.  The README does not promise any more that the VSCode plugin already works and the binaries have the version number in the file name now.

packages/writing-style/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+The base style is the [Google Developer Documentation Style Guide](https://github.com/errata-ai/Google), which is included as a copy (MIT licensed, too) and modified and enriched with commercetools specific rules.
+
+The configuration defaults to checking `.md`, `.mdx`, `.txt`, and `.html` files, which can be overridden by calling it with a command-line parameter `commercetools-vale --glob='*.{js,md}' ./path/to/content` passing a list of [supported file format extensions](https://errata-ai.github.io/vale/formats/#formats).
+
+> As of Nov 2019, basing off Google's styles was a pragmatic choice. Other styles like Microsoft's are attractive starting points, too.
+
+## Linter Usage: Standalone command-line
+
+1.  Install the command globally via `npm install -g @commercetools-docs/writing-style`
```
emmenko:
```suggestion
1.  Install the command globally via `yarn global add @commercetools-docs/writing-style` or `npm install -g @commercetools-docs/writing-style`
```

packages/writing-style/commercetools-vale.js
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+const { spawnSync } = require('child_process');
+const path = require('path');
+
+const valeBinary = path.join(__dirname, `/bin/vale-1.7.1.${process.platform}`);
```
emmenko:
```suggestion
const version = `1.7.1`;
const valeBinary = path.join(__dirname, `/bin/vale-${version}.${process.platform}`);
```

emmenko:
We can merge this if you want, ignore the percy snapshot (the codeblocks colors are somehow flaky)

PR #29, opened 2019-11-07
refactor(theme): to support different scenarios in Link component

emmenko:
This PR moves the logic of the Link component. I refactored a bit the cide and "merged" the 2 implementations.

I still need to move the tests, but so far looks good.


packages/gatsby-theme-docs/src/components/link.js
```diff
@@ -0,0 +1,237 @@
...
+import { css } from '@emotion/core';
+import styled from '@emotion/styled';
+import { OutboundLink } from 'gatsby-plugin-google-analytics';
+import ExternalLinkIcon from '../icons/external-link-icon.svg';
+import { colors, dimensions } from '../design-system';
+import { useSiteData } from '../hooks/use-site-data';
+
+const isProduction = process.env.NODE_ENV === 'production';
+
+const productionHostname = 'docs.commercetools.com';
```
emmenko: @nkuehn should we make this configurable? Or do you want to keep it hardcoded?

packages/gatsby-theme-docs/src/components/link.js
```diff
@@ -0,0 +1,237 @@
...
+    css={linkStyles}
+    target="_blank"
+    rel="noopener noreferrer"
+  />
+);
+const AnchorLink = styled.a`
+  ${linkStyles}
+`;
+const InternalSiteLink = styled(AnchorLink)``;
+const HistoryLink = styled(GatsbyLink)`
```
emmenko: I named those components after their contextual usage:
- `ExternalSiteLink`
- `InternalSiteLink`
- `AnchorLink`
- `HistoryLink`

packages/gatsby-theme-docs/src/components/link.js
```diff
@@ -0,0 +1,237 @@
...
+`;
+
+/**
+ * The <Link> component handles different scenarios of how links are written in the website.
+ * Below are the recommended convention for providing a link `href`. However, to be
+ * as compatible as possible, the component also tries to support other possible scenarios,
+ * including relative filesystem paths, etc.
+ *
+ * ---
+ * # Links conventions
```
emmenko: This is the convention that we agreed on. There will be a proper document as follow up to instruct the writers how to write the links properly.

packages/gatsby-theme-docs/src/layouts/content.js
```diff
@@ -38,16 +39,19 @@ const ResizableGrid = styled.div`
...
           title
+          description
+          author
         }
       }
     }
   `);
   return (
-    <React.Fragment>
+    <SiteDataContext.Provider value={data.site}>
```
emmenko: We put the query result into a context, so that components can easily access it.

emmenko:
I still need to add some e2e tests, you can start reviewing the PR though.

packages/gatsby-theme-docs/src/components/link.spec.js
```diff
@@ -0,0 +1,134 @@
...
+const renderLink = (ui, initialPath) => {
+  const source = createMemorySource(initialPath);
+  const history = createHistory(source);
+  return render(<LocationProvider history={history}>{ui}</LocationProvider>);
+};
+
+const pathPrefix = '/prefix';
+const withPrefix = path => [pathPrefix, path].join('/').replace(/\/\/+/g, '/');
+
+describe('rendering', () => {
```
emmenko: image

packages/gatsby-theme-docs/src/components/link.js
nkuehn: That's a good general question - we might need the official domain e.g. for the canonical links, too and we likely also need it for RSS feeds later on.  

We could force setting it in the site metadata in the config and have a default in the theme. I would not hardwire it inside the link implementation though.   

BTW that's a typical case where you may need a static query in a component and think about testability.

packages/gatsby-theme-docs/src/components/link.js
nkuehn: to be honest all three name choices (InternalSiteLink, HistoryLink and InlineLink) do not feel logical to me.   

"InternalSite" is just for "on the same page" but not for "on the same site".  

"History" is a different concern to me than the linking itself.  You can choose to make  a GatsbyLink triggering a history or not but that's a different angle.   

And "Inline" rather sounds like inline vs. block to me but in the implementation it is used for external links.

nkuehn:
Sorry, I did individual comments because I got used to GitHub defaulting to a review now. 

While grooming the backlog I found another use case that needs to be caught:  If a link points to a target window that is not the same window (esp. _blank  ) it needs to be a plain HTML link to work properly.  I might be that it's fixed in gatsbyJS in the meanwhile but we should do at least a manual test. 

I will review the tests next week.

packages/gatsby-theme-docs/src/components/link.js
emmenko: Inline is not one of the 4 main components listed above, it’s just for styling.

We can improve the naming if you want, the point is that we have those 4 different link types 😉

packages/gatsby-theme-docs/src/components/link.js
emmenko: Sounds good to me to put it in the metadata

emmenko:
The `_blank` case is actually kind of handled already, and it’s rendered as an external link. Let’s have a deeper discussion about it next week.

packages/gatsby-theme-docs/src/components/link.spec.js
```diff
@@ -0,0 +1,134 @@
...
+      location: { pathname: '/page-1/' },
+      expected: { role: 'image-link' },
+    },
+    {
+      title: 'external link',
+      props: {
+        href: 'https://github.com/commercetools/commercetools-docs-kit',
+      },
+      location: { pathname: '/page-1/' },
+      expected: { role: 'external-link' },
```
nkuehn: I would prefer explicitly testing the href output in all test cases and explicitly for prod vs. dev.  
It's also for making sure that all links actually write a href so that  opening in a new tab always works. 

And I tend to not trust test implementations that do too much magic in own code to generate assertions for cases that are not actually "linear" in the sense that all cases are handled the same.

packages/gatsby-theme-docs/src/components/link.spec.js
```diff
@@ -0,0 +1,134 @@
...
+    },
+    {
+      title: 'anchor link',
+      props: { href: '#title-1' },
+      location: { pathname: '/page-1/' },
+      expected: { role: 'anchor-link' },
+    },
+    {
+      title: 'anchor link to same page',
```
nkuehn: Not sure about the naming convention that has the name of the expected behavior ("anchor") as the main keyword, but it's not a blocker.

packages/gatsby-theme-docs/src/components/link.js
nkuehn: OK, i suggest to think about a better name for `HistoryLink` only.  I would prefer to stick with GatsbyLink becuase it's key property is that we delegate the handling to gatsbyJS to do whatever it thinks it should do with it.   If you want it less implementation-specific it's  a sort of  In-App Link.  

Other links also write history and the GatsbyLink can choose to not write history depending on its configuration and parameters, that's why I am not convinced of "History".

packages/gatsby-theme-docs/src/components/link.spec.js
```diff
@@ -0,0 +1,134 @@
...
+      title: 'internal link with hash',
+      props: { href: 'https://docs.commercetools.com/site-key/page-1#title' },
+      location: { pathname: '/page-1/' },
+      expected: { role: 'internal-link', href: '/site-key/page-1#title' },
+    },
+  ];
+
+  const scenariosInProdMode = scenarios.map(scenario => ({
+    ...scenario,
+    location: { pathname: withPrefix(scenario.location.pathname) },
```
nkuehn: if you keep the assertion generation magic I think we need to cover the special case we spoke about (a link to a different microsite would be kept as an absolute URL in dev mode but we would remove the host for production because otherwise we can't test refactors that span across multiple sites)

packages/gatsby-theme-docs/src/components/link.spec.js
```diff
@@ -0,0 +1,134 @@
...
+      title: 'history link to another page',
+      props: { href: '/page-2' },
+      location: { pathname: '/page-1/' },
+      expected: { role: 'history-link' },
+    },
+    {
+      title: 'history link to a sub-page',
+      props: { href: 'getting-started' },
+      location: { pathname: '/page-1/' },
+      expected: { role: 'history-link', href: '/page-1/getting-started' },
```
nkuehn: Do we want to automatically append trailing slashes in history links or is that too magic?

packages/gatsby-theme-docs/src/components/link.spec.js
```diff
@@ -0,0 +1,134 @@
...
+      title: 'history link to another page with hash',
+      props: { href: '/page-2#title' },
+      location: { pathname: '/page-1/' },
+      expected: { role: 'history-link', href: '/page-2#title' },
+    },
+    {
+      title: 'internal link',
+      props: { href: 'https://docs.commercetools.com/site-key/page-1' },
+      location: { pathname: '/page-1/' },
+      expected: { role: 'internal-link', href: '/site-key/page-1' },
```
nkuehn: Since we chose to make this the default and recommended way we could choose to remove the alert / popup behavior and actually let people test it locally by navigating.   

It's a tradeoff - most authoring work is to create new content and link to existing places so the experience of doing (and locally testing) that is more important than the experience of refactoring  the URL structure or section naming.   Refactoring must mostly work at all through the link checker in the builds.

packages/gatsby-theme-docs/src/components/link.spec.js
```diff
@@ -0,0 +1,134 @@
...
+    },
+    {
+      title: 'internal link using "/../" notation',
+      props: { href: '/../site-key/page-1' },
+      location: { pathname: '/page-1/' },
+      expected: { role: 'internal-link', href: '/site-key/page-1' },
```
nkuehn: If we choose to not show a popup / alert warning any more we could (should?) make this situation an absolute link to https://docs.commercetools.com  in dev mode.

packages/gatsby-theme-docs/src/components/link.js
emmenko: Technically "history" is the correct term, as it refers to the [browser history](https://developer.mozilla.org/en-US/docs/Web/API/History) which is what Gatsby uses through `@reach/router`.

My intent with the naming was about that and making it clear how the link behaves. If you prefer to keep GatsbyLink it's fine, I don't want to argue much about the naming.
I just wanted to clarify the intention behind those choices from a technical perspective.

packages/gatsby-theme-docs/src/components/link.spec.js
emmenko: The prod links will be tested as e2e tests (which is what is currently missing in this PR).

packages/gatsby-theme-docs/src/components/link.spec.js
emmenko: Ah right, I forgot about the dev/prod mode difference

packages/gatsby-theme-docs/src/components/link.spec.js
emmenko: Sure, we can do that

packages/gatsby-theme-docs/src/components/link.spec.js
emmenko: I'm not sure I get the point. In dev mode you still get the alert when you click on the link.

emmenko:
While testing the links I found a strange behavior regarding some of the links, which differs from dev/prod.
The strange thing is about the `href` value that we get as a `prop`, which makes it kind even more strange because it's not something we can control.

Given the link `[Link]()`, in dev mode we get the following props:

```
{href: "/smoke-tests/", children: "Link"}
```

which results in the link pointing to the parent directory.
However, in **prod mode the link works fine** as it points to `/smoke-tests/links/`.

The same problem appears for the following relative links:

```
[Link](../)
=> {href: "/", children: "Link"}

[Link](link-target)
=> {href: "/smoke-tests/link-target/", children: "Link"}

[Link](link-target/)
=> {href: "/smoke-tests/link-target/", children: "Link"}
```

@nkuehn have you noticed this before?

emmenko:
Thinking about it, it could be that the href that we get from mdx is resolved from the file system, which in dev mode is just a file compared to prod mode where it’s an index.html inside a folder 🙃

emmenko:
So given the assumption above, the behavior is correct, we just need to change/fix the links in the test page as it’s misleading.

emmenko:
Alright I managed to solve the issue by using the `gatsby-remark-rewrite-relative-links` plugin, which resolves relative paths based on the file system.

Now we get the correct `href` from the props:

```
[Link](../images/image)
=> {href: "/images/image/", children: "Link"}

[Link](links/link-target)
=> {href: "/smoke-tests/links/link-target/", children: "Link"}
```

emmenko:
@nkuehn @davifantasia do you want to have a final look? This is good to go from my side now.

nkuehn:
For the record:  @emmenko and me just discussed the question whether the authors need to write relative urls  relative to the file system location or relative to the effective URL (which by default has a trailing slash). 

High-level finding is that the fact that both URLs with and without trailing slashes can be written and both work will always lead to potential confusion.  

We'll reconfigure all defaults  and gatsby itself to not use trailing slashes,  which leads to the actual URL in production and development matching the path depth of the file system the content is originally authored in.

emmenko:
🚀

PR #31, opened 2019-11-08
feat: styling global reset

davifantasia:


packages/gatsby-theme-docs/src/layouts/internals/reset.js
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        body {
+          margin: 0;
+          padding: 0;
+          border: 0;
+          font-style: normal;
+          font-variant: normal;
+          font-stretch: normal;
+          font-family: 'Roboto', -apple-system, system-ui, BlinkMacSystemFont,
```
emmenko: We can probably remove the font styling here, since we define it in the `Global` component already.
We should only have one source of truth.

packages/gatsby-theme-docs/src/layouts/internals/reset.js
```diff
@@ -0,0 +1,358 @@
...
+            'Segoe UI', 'Helvetica Neue', Arial, sans-serif;
+          font-weight: ${typography.fontWeights.regular};
+          font-size: ${typography.fontSizes.body};
+          line-height: ${typography.lineHeights.body};
```
emmenko: Same here

packages/gatsby-theme-docs/src/layouts/internals/reset.js
```diff
@@ -0,0 +1,358 @@
...
+          text-align: left;
+          vertical-align: baseline;
+          color: ${colors.light.primary};
```
emmenko: Same here

packages/gatsby-theme-docs/src/layouts/internals/reset.js
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        /* See https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible
+          and https://developer.paciellogroup.com/blog/2018/03/focus-visible-and-backwards-compatibility/
+          and the boostrap reboot source code */
+        [tabindex='-1']:focus:not(:focus-visible) {
+          outline: 0 !important;
+        }
+
+        /* Default typography for typical  monospace elements */
```
emmenko: We can move this to the `Global` component, since we are explicitly defining some values here.

packages/gatsby-theme-docs/src/layouts/internals/reset.js
```diff
@@ -0,0 +1,358 @@
...
+        code,
+        kbd,
+        samp,
+        pre {
+          font-family: 'Roboto Mono', Menlo, Monaco, Consolas, 'Liberation Mono',
+            'Courier New', monospace;
+          font-size: ${typography.fontSizes.small};
+        }
+
+        /* default typography to bold (Roboto 700) weight for typically bold formatted elements */
```
emmenko: Same for those other things

packages/gatsby-theme-docs/src/layouts/internals/reset.js
```diff
@@ -0,0 +1,358 @@
...
+        strong {
+          font-weight: ${typography.fontWeights.bold};
+        }
+
+        /* make small small */
+        small {
+          font-size: ${typography.fontSizes.small};
+        }
+
+        /* default to non-underlined, commercetools colored links. */
```
emmenko: I think we can remove this, links are already styled explicitly

emmenko:
Please check the VRTs to see if there are broken things or not.

PR #36, opened 2019-11-11
refactor(e2e): remove unnecessary snapshots

emmenko:
Just to avoid taking too many unnecessary snapshots, as we are paying for them. Those snapshots were just added mostly for demo purposes.


emmenko:
We should have at least one percy snapshot, so I added a test to take one for small screen when the main nav is opened.

PR #50, opened 2019-11-12
refactor(code-block): no line numbers, wrap line, bash prompt

emmenko:
image


websites/docs-smoke-test/src/content/code-samples/code-block.mdx
```diff
@@ -9,6 +9,15 @@ The following block-level code samples must be
 - have a copy button
 - do not have a header bar because they don't have a title and are single-language
 
+### Terminal/Console
+
+```console{outputLines: 2-4}
```
emmenko: This is how it works, by default all lines have the "prompt" line and you can "hide it" by providing a line or range of line.

In this case only the first line has the prompt.

emmenko: We should also use `console`, as otherwise `bash`, `shell` etc should be used for scripting.

packages/gatsby-theme-docs/src/components/markdown.js
```diff
@@ -221,14 +222,15 @@ const CodeBlock = props => {
     props.children.props && props.children.props.children
       ? props.children.props.children
       : props.children;
-  const { splitLanguage, highlightLines } = codeBlockParseOptions(language);
+  const { highlightLines, outputLines } = codeBlockParseOptions(
+    languageToken + props.children.props.metastring
```
emmenko: This is what we get as props

image

emmenko:
@nkuehn please have another look, I refactored a bit the parsing implementation to make it simpler. I also documented the usage of the code block syntax.

PS: I had to force push so your changes are gone. Next time please let me know if you want to change something.
Also please try to use `git rebase` and `git push --force-with-lease`.

packages/gatsby-theme-docs/src/components/markdown.js
emmenko: This changed, not relevant anymore

nkuehn:
just in case we're considering to switch back to pure custom CSS  line numbers and/or terminal formats, here's the current implementation:  https://github.com/commercetools/docs.commercetools.com/blob/master/_assets/stylesheets/content.scss#L224-L250

PR #52, opened 2019-11-12
feat: content notifications (used in MDX files)

davifantasia:


packages/gatsby-theme-docs/src/components/notifications.js
```diff
@@ -0,0 +1,114 @@
...
+import createStyledIcon from '../utils/create-styled-icon';
+import markdown2React from '../utils/markdown-2-react';
+import { pxToRem, colors, tokens, dimensions } from '../design-system';
+
+const Container = styled.div`
+  display: flex;
+  width: 100%;
+  padding: ${dimensions.spacings.m};
+  margin: ${dimensions.spacings.s} 0;
+  border: ${pxToRem('1px')} solid ${colors.light.borderInfo};
```
emmenko: The border should just stay to 1px, at least that's what we do at the moment. If we were to change that, we should do that in a separate PR. Also the value should be defined in the design-system, so that we don't need to use `pxToRem` directly.

```suggestion
  border: 1px solid ${colors.light.borderInfo};
```

packages/gatsby-theme-docs/src/components/notifications.js
```diff
@@ -0,0 +1,114 @@
...
+  border-radius: ${tokens.borderRadius6};
+`;
+
+const IconWrapper = styled.div`
+  flex-shrink: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: ${pxToRem('24px')};
```
emmenko: It would make sense to define icon sizes in the design-system, then use those variables (so no direct use of `pxToRem`)

packages/gatsby-theme-docs/src/components/notifications.js
```diff
@@ -0,0 +1,114 @@
...
+  height: ${pxToRem('24px')};
+`;
+
+const InfoTextWrapper = styled.div`
+  margin-left: ${dimensions.spacings.s};
+`;
+
+function Box(props) {
+  return (
+    <Container css={props.containerCustomStyle}>
```
emmenko: Since we can use the `css` prop, it will be injected as a `className`, so we can just use that:

```suggestion
    <Container className={props.className}>
```

packages/gatsby-theme-docs/src/utils/markdown-2-react.spec.js
```diff
@@ -0,0 +1,13 @@
...
+import markdown2React from './markdown-2-react';
+
+describe('markdown2React', () => {
+  it('should return a react component from a markdown string', async () => {
+    const markdownString = '# This is a markdown title';
+    const expectedReactElement = React.createElement('h1', { key: 'h-1' }, [
+      'This is a markdown title',
+    ]);
+
+    expect(markdown2React(markdownString)).toEqual(expectedReactElement);
```
emmenko: I think it would be enough to just check that the return value is a react element:

```suggestion
    expect(React.isValidElement(markdown2React(markdownString))).toBeTruthy();
```

This avoids the need to test the implementation of the parsing library, which is not useful to us

packages/gatsby-theme-docs/src/components/notifications.js
```diff
@@ -0,0 +1,114 @@
...
+};
+
+export function Info(props) {
+  const containerCustomStyle = css`
+    border-color: ${colors.light.iconInfo};
+    background-color: ${colors.light.surfaceInfo2};
+  `;
+
+  return (
+    <Box containerCustomStyle={containerCustomStyle} icon="default">
```
emmenko:
```suggestion
    <Box css={containerCustomStyle} icon="default">
```

However, what we can also do is to pick the proper styles within the `Container` component, based on the given `icon` prop. This avoids to define the styles in every component.

```jsx
const Container = styled.div`
  border-color: ${props => {
    switch (props.icon) {
      case 'warning'
        return colors.light.iconWarning;
      case 'error'
        return colors.light.iconError;
      default
        return colors.light.iconInfo;
    }
  }};
  background-color: ${props => {
    switch (props.icon) {
      case 'warning'
        return colors.light.surfaceWarning;
      case 'error'
        return colors.light.surfaceError;
      default
        return colors.light.surfaceInfo2;
    }
  }};
`
```

packages/gatsby-theme-docs/src/utils/markdown-2-react.js
```diff
@@ -0,0 +1,12 @@
+import unified from 'unified';
+import markdown from 'remark-parse';
+import remark2react from 'remark-react';
+
+export default function markdown2React(markdownString) {
+  const result = unified()
+    .use(markdown, { commonmark: true })
+    .use(remark2react)
```
emmenko: We need to pass the option `remarkReactComponents` to map the markdown elements to our custom markdown components.
https://github.com/remarkjs/remark-react#optionsremarkreactcomponents

We can either map a few of them (e.g. link, paragraph, etc) or all of them.
In any case, we should pass the options as a second argument of the `markdown2React` function, to avoid coupling the utility function with react components.

/cc @nkuehn

packages/gatsby-theme-docs/src/design-system.js
```diff
@@ -39,6 +39,9 @@ export const colors = {
     surfaceCodeHighlight: uikitColors.colorAccent40,
     surfaceInfo: uikitColors.colorInfo95,
     surfaceQuote: uikitColors.colorAccent98,
+    surfaceInfo2: '#eef4f7',
+    surfaceWarning: '#fef0e7',
```
emmenko: This is the `colorWarning95` from uikit

packages/gatsby-theme-docs/src/design-system.js
```diff
@@ -39,6 +39,9 @@ export const colors = {
...
+    surfaceError: '#ffe5ee',
```
emmenko: This is the `colorError95` from uikit

packages/gatsby-theme-docs/src/design-system.js
```diff
@@ -55,6 +58,10 @@ export const colors = {
     link: uikitColors.colorPrimary25,
     linkHover: uikitColors.colorPrimary,
     linkNavigation: uikitColors.colorInfo,
+    // icons
+    iconInfo: uikitColors.colorInfo,
+    iconWarning: '#f16d0e',
```
emmenko: This is the `colorWarning` from uikit

packages/gatsby-theme-docs/src/design-system.js
```diff
@@ -39,6 +39,9 @@ export const colors = {
     surfaceCodeHighlight: uikitColors.colorAccent40,
     surfaceInfo: uikitColors.colorInfo95,
     surfaceQuote: uikitColors.colorAccent98,
+    surfaceInfo2: '#eef4f7',
```
emmenko: Are you sure this is correct? It should be `colorInfo95` (also for consistency with how the other colors are picked), which is already mapped as `surfaceInfo`

packages/gatsby-theme-docs/src/components/notifications.js
```diff
@@ -0,0 +1,128 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import { ClassNames } from '@emotion/core';
```
emmenko: This should be 

```suggestion
import { css } from '@emotion/core';
```

packages/gatsby-theme-docs/src/components/notifications.js
```diff
@@ -0,0 +1,128 @@
...
+import createStyledIcon from '../utils/create-styled-icon';
+import markdown2React from '../utils/markdown-2-react';
+import { colors, tokens, dimensions } from '../design-system';
+
+const Container = styled.div`
+  display: flex;
+  width: 100%;
+  padding: ${dimensions.spacings.m};
+  margin: ${dimensions.spacings.s} 0;
+  border: 1px solid;
```
emmenko: Isn't the color missing? 🤔

packages/gatsby-theme-docs/src/components/notifications.js
```diff
@@ -0,0 +1,128 @@
...
+function Box(props) {
+  return (
+    <Container className={props.className}>
+      <IconWrapper>{renderIcon(props.icon)}</IconWrapper>
+      <InfoTextWrapper>{markdown2React(props.children)}</InfoTextWrapper>
+    </Container>
+  );
+
+  function renderIcon(icon) {
+    const InformationIcon = createStyledIcon(UnstyledInformationIcon);
```
emmenko: Please define those variables after the imports, not within the render function

emmenko:
@davifantasia PS: please try to avoid those merge commits 🙏

image

packages/gatsby-theme-docs/src/components/content-notifications.js
```diff
@@ -0,0 +1,139 @@
...
+          <Info>
+            This *is* **markdown**!
+          </Info>
+
+        Example (it works):
+          <Info>
+
+            This *is* **markdown**!
+
+          </Info>
```
emmenko: /cc @nkuehn this works 👆I think we should just go with that syntax, as it's following the spec.

I haven't found another way around that without having to do the parsing on our own.

nkuehn: OK, interesting.  I

I still think it's too implicit from an author's perspective, but we can opt to try it with real users and see whether they are going mad or not.  It's charming nevertheless a) because it's official and b) because an author can control in the content whether he/she wants the body to be parsed as MDX or not. 

If we find the syntax to be too implicit I don't see a problem in parsing on our own, I just opt for then re-using as much as possible from the central MDX configuration in the MDX Provider (e.g. the components).   

We do need the helper for parsing on our own in any case in the RAML generated content.  The RAML descriptions are per specification markdown and we have to render that programmatically.

PR #55, opened 2019-11-12
[WIP] Add RAML parser and validator plugin

nkuehn:
NEW STATUS: WIP - this will have to provide the a similar output structure as the one we agreeed with @jenschude to provide from the RMF implementation.   Pausing this for now. 

Due to the varying problems with the varying RAML parsers we bias towards a setup that decouples the parser, validator and resolver implementation of the API specifications from the code that provides the individual types, endpoints, security schemes etc on the gatsbyJS graphQL in a more granular manner. 

This is the first step, an implementation using the "new official" RAML parser  which turned out to be not fast enough on the (huge) projects API but a good choice for smaller APIs. 

It still lacks some features of our existing implementation that's based on the "old" parser but is good to start with now.   E.g. It does not hot-reload when included files are changed and some necessary metadata are lost, especially the orginally declared type name.


packages/gatsby-transformer-webapi/gatsby-node.js
```diff
@@ -0,0 +1,58 @@
...
+  const parser = webapi.WebApiParser;
+  const model = await parser.raml10.parse(`file://${node.absolutePath}`);
+
+  if (validate) {
+    const validationReport = await parser.raml10.validate(model);
+    if (!validationReport.conforms) {
+      reporter.warn(`validation errors in API "${apiKey}" :`);
+      reporter.warn(validationReport.toString());
+    }
+  }
```
davifantasia: This can be extracted into a validation function

packages/gatsby-transformer-webapi/gatsby-node.js
```diff
@@ -0,0 +1,58 @@
...
+    parent: node.id,
+    internal: {
+      content: resolvedRamlString,
+      contentDigest: createContentDigest(resolvedRamlString),
+      mediaType: node.internal.mediaType,
+      type: 'ResolvedWebapiRaml',
+    },
+  };
+  createNode(apiNode);
+  createParentChildLink({ parent: node, child: apiNode });
```
davifantasia: Not needed at the moment but as more processing needs to be done on the resolved spec, this would need to move into its own function, a node creation function for example.

packages/gatsby-transformer-webapi/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "keywords": [
+    "gatsby",
+    "gatsby-plugin",
+    "raml",
+    "amf"
+  ],
+  "scripts": {
+    "clean": ": Nothing to clean in gatsby-transformer-webapi",
+    "test": "jest",
+    "build": ": No standalone build of the WebAPI transformer plugin in production."
```
davifantasia: I don't think these scripts are needed for now. And we already setup and run jest from the root folder.

packages/gatsby-transformer-webapi/src/helpers.js
```diff
@@ -0,0 +1,37 @@
+// unless we find a more explicit way to detect the phase (initial vs. incremental build),
+// we rely on the last modification date to assume hot reload vs. initial load:
+const wasNodeJustChanged = node => {
+  if (node.ctimeMs && node.absolutePath) {
+    const sinceLastChanged = Date.now() - node.ctimeMs;
+    return sinceLastChanged < 1000; // times vary significantly, sometimes up to 600ms
+  }
+  return false;
+};
```
davifantasia: This function should be included since it's not called anywhere.

packages/gatsby-transformer-webapi/gatsby-node.js
nkuehn: I prefer leaving it in the simplest possible state because we are will likely not follow up with that parser for now

packages/gatsby-transformer-webapi/gatsby-node.js
nkuehn: I prefer leaving it in the simplest possible state because we are will likely not follow up with that parser for now

packages/gatsby-transformer-webapi/scripts/dump.debug.js
```diff
@@ -0,0 +1,59 @@
+const webapi = require('webapi-parser');
```
davifantasia: I don't think this file should be included in this merge, for similar reasons as the files it actually dumps.

websites/api-docs-smoke-test/src/api-specs/.gitignore
```diff
@@ -0,0 +1,4 @@
+*.amfgraph.raml
```
davifantasia: is it possible to just not include the files in the merge?

websites/api-docs-smoke-test/src/api-specs/test/api.raml.validation.txt
```diff
@@ -0,0 +1,4 @@
+Model: file:///Users/nkuehn/GitHub/commercetools-docs-kit/websites/api-docs-smoke-test/src/api-specs/test/api.raml
```
davifantasia: This should not be included in the merge.

packages/gatsby-transformer-webapi/scripts/dump.debug.js
nkuehn: Please propose a better way to manage debugging and development helper code that is necessary to develop on a piece of software. 

It's not practically possible to run the full gatsby in debug mode to interactively explore the data stuctures of the internal API provided by the parser (it's very very slow and does not behave stable).  And the parser's API is very badly documented. 

IMHO  `/scripts` is the conventional place for helper scripts that you need in the development workflow.   Tell me a better place to manage such code an I'm all fine.  But deleting or not sharing it in the team is not an option.

websites/api-docs-smoke-test/src/api-specs/.gitignore
nkuehn: I think it's not a good idea to rely on all team members to just not forget to commit, especially with so much automation and CI in place. 
Our  .gitignores  often contain a broad set of ignores that cover all potential developer situations to avoid accidental commits.  

if we remove the script though, then the ignore has to go, too - that's right.

websites/api-docs-smoke-test/src/api-specs/test/api.raml.validation.txt
nkuehn: I must have messed up something during testing the .gitignores and file deletions.  I'll check, thanks !

websites/api-docs-smoke-test/src/api-specs/.gitignore
davifantasia: Yes, just removing the scripts works too and we won't need to ignore them :)

packages/gatsby-transformer-webapi/scripts/dump.debug.js
davifantasia: For me, if I'm not using the IDE's debugger then I'm logging.

I also gave up on vs code debug configs because of its inconsistencies, but I found that the single file visual studio code debugger (just pressing f5) is the most consistent so I use that quite often.

nkuehn: Let's talk about the best debug setup tomorrow in our call and document / commit the best practice - in this kind of RAML parser / data structure debugging code it's such a massive productivity difference to not have to work with logging.  I did spend ages on the old parser until I wrote myself a script like this which works reliably in the  debugger.  About 5* as fast to read the data structures out or the parser API then because you interactively browse all its function calls and directly see the return values.

nkuehn:
Status update: @jenschude  has started working on a generator CLI based on the [RMF code generator](https://github.com/vrapio/rmf-codegen)  that writes any sort of RAML API into individual RAML Fragment Files that we can directly consume here without fully parsing / expanding them. 

We'll restart work here once that work is ready to test.  It will be a Java CLI at first to get moving.

nkuehn:
supersede by #119

PR #60, opened 2019-11-13
feat(theme): setup GA and make it configurable

### LINKED ISSUE #43: add google analytics plugin and make the id configurable per site
default should be an empty or no id configured in a way that does not track but does not throw errors either.

emmenko:
Closes #43


packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -2,7 +2,7 @@
 
 const path = require('path');
 
-module.exports = {
+module.exports = (themeOptions = {}) => ({
```
nkuehn: It like the pattern to define all allowed options with defaults, so it's much safer to track whether the documentation of the options is correct and matches the implementation. 

`themeOptions = { gaTrackingId: false }`  ?

emmenko: Sure, makes sense

emmenko: I added this here: https://github.com/commercetools/commercetools-docs-kit/pull/62/files#diff-829e90677cff64de530aac85e640cabcR20-R21

PR #61, opened 2019-11-13
fix(theme): prefer overflow: auto

### LINKED ISSUE #59: theme:  main container always has scrollbar
The main container y  overflow is not set to auto, so when on windows or having a mouse attached the scrollbar is visible independently of whether it's needed. 

From a windows user perspective that's perceived as a bug because it deviates from the 99% convention how websites behave.

emmenko:
Fixes #59

### DIFF
--- a/packages/gatsby-theme-docs/src/layouts/internals/globals.js
+++ b/packages/gatsby-theme-docs/src/layouts/internals/globals.js
@@ -80,8 +80,7 @@ const Globals = () => (
       }
       .gatsby-highlight > code,
       .gatsby-highlight code[class*='language-'],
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-docs/src/layouts/internals/globals.js
```diff
@@ -90,23 +89,10 @@ const Globals = () => (
         margin: 0;
         padding: 0;
       }
-      .gatsby-highlight pre[class*='language-'].line-numbers {
```
emmenko: Unrelated cleanup changes...

PR #67, opened 2019-11-14
refactor: restructure test pages, snapshot only component pages

emmenko:
The idea with this PR is to clean up and restructure a bit the test pages. We also don't need to snapshot all pages, I think it's enough for now to do that for the component pages.

Let me know if this makes sense to you.


websites/docs-smoke-test/src/data/navigation.yaml
```diff
@@ -1,71 +1,31 @@
...
+      path: /components/content-notifications
+    - title: Images
+      path: /components/images
+    - title: Links
+      path: /components/links
 - chapter-title: Chapter 9999 With a Longer Multi-Word Name that Wraps
   pages:
     - title: Page with some very long title too much for our innocent API to handle
-      path: /#foobar
+      path: /not-found-1
```
nkuehn: This will break the builds once the link checker is merged (it checks the test sites for broken links then).  That was the reason the paths were just an anchor on the index page.    They need to be actual anchors on the index page.

websites/docs-smoke-test/src/content/components/images.mdx
```diff
@@ -0,0 +1,67 @@
+---
+title: Images
+---
+
+# With caption
+
+Images can have a caption like the above. It's authored by setting the `title` using the markdown notation.
```
nkuehn: Do you mean "below" instead of "above"?

websites/docs-smoke-test/src/data/navigation.yaml
```diff
@@ -1,71 +1,31 @@
...
-      path: /smoke-tests/lists/
-    - title: Definition Lists
-      path: /smoke-tests/definition-lists/
-    - title: Empty Page
-      path: /smoke-tests/empty-page/
-    - title: Page with little content
-      path: /smoke-tests/page-with-little-content/
-- chapter-title: Custom Elements
+    - title: Empty
+      path: /views/empty
```
nkuehn: FYI: the history of the empty page was that it was intended to be a VRT because there were layout bugs when there was no content or extremely short content (less than a line length).  I added it to find regressions over the positioning in big screen widths.

websites/docs-smoke-test/src/data/navigation.yaml
emmenko: Ah right

websites/docs-smoke-test/src/data/navigation.yaml
emmenko: Yeah, it’s currently tested by the e2e tests

emmenko:
I'm still not 100% happy with how the snapshots are taken...I'll do some more investigation

emmenko:
@nkuehn please have a final look, this is now ready to go from my side.

packages/gatsby-theme-docs/src/layouts/internals/layout-page-navigation.js
```diff
@@ -134,7 +134,7 @@ const LayoutPageNavigation = props => {
         <SlidingContainer>{navigationContainer}</SlidingContainer>
       </ContainerOverlay>
 
-      <GridContainer>
+      <GridContainer role="page-navigation">
```
nkuehn: I'm not aware of an ARIA role "page-navigation".   There's only "navigation" if you are not using the `<nav>` tag.  I think what you want here is an aria-label "Page Navigation".

emmenko: Right, let me change it to `navigation` and provide a `aria-label` for differentiate between the 2 navs

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role#Labeling_landmarks

PR #72, opened 2019-11-15
chore: use vale local bin for vscode

emmenko:
So our [PR](https://github.com/testthedocs/vscode-vale/pull/8) for make the extension configurable got merged and released yesterday. I tried this out now and it works!

image

However, the `mdx` files are [being ignored by the extension](https://github.com/testthedocs/vscode-vale/blob/de4ab0bab73774f01d1ee3114bc84a806db0c3c5/src/extension.ts#L275). I'll open a new PR to add the `mdx` extension and maybe to also make it configurable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,8 @@ public
 
 # Editor/IDE
 .idea
+-.vscode
+-!.vscode/settings.json
 
 # Cypress
 cypress/screenshots
--- a/.gitignore
+++ b/.gitignore
@@ -73,8 +73,6 @@ public
 
 # Editor/IDE
 .idea
-.vscode
-!.vscode/settings.json
 
 # Cypress
 cypress/screenshots
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "vscode-vale.path": "node_modules/.bin/commercetools-vale"
+}


nkuehn:
P.S.:  Although I have the latest version of the plugin installed (0.8.4) VSCode complains that the setting is not known (greyed out).  Maybe it's not properly declared yet in the Plugin or I have a local issue. 

image

emmenko:
> VSCode complains that the setting is not known (greyed out)

I have the same, not sure exactly what the problem is.

> default prettier as the formatter

Hmm let's maybe have a look first at your setup, I'm not sure this is something to add to the workspace settings 🤔

.gitignore
```diff
@@ -73,8 +73,8 @@ public
 
 # Editor/IDE
 .idea
-.vscode
-!.vscode/settings.json
+-.vscode
```
nkuehn: I think / guess you actually want to achieve this:  (.vscode folder not tracked except the settings file)
```suggestion
.vscode/*
.vscode/settings.json
```

git doesn't even list subfolders so unless you have a more specific exclude using a wildcard you can't re-include content in a previously excluded folder.

nkuehn:
forget the comment about prettier, I need this only in commecetools-docs  to make sure authors don't get frustrated over formatting changes.

.gitignore
emmenko: Hmm are you sure? This works fine for me

image

davifantasia: These don't work for me:
```
-.vscode
-!.vscode/settings.json
```

nkuehn: @davifantasia  this branch is already merged, please open an issue if the current master does not behave as expected in your setup. Thanks !

PR #90, opened 2019-11-22
docs(theme): document main theme features and how to use them

emmenko:


emmenko:
@nkuehn can we merge this? We can always improve things as we go...

packages/gatsby-theme-docs/README.md
```diff
@@ -1,3 +1,168 @@
...
+          websiteKey: 'change-me',
+        },
+      },
+    ],
+  };
+  ```
+
+  Available options for the theme plugin are:
+
+  - `websiteKey` (**required**): the identifier of the website, without whitespaces. Usually this value would be the same as the `pathPrefix` without the leading slash.
```
nkuehn:
```suggestion
  - `websiteKey` (**required**): the identifier of the website, used for error reporting and similar concerns. Usually this value would be the same as the `pathPrefix` without the leading slash and without without whitespace.
```

packages/gatsby-theme-docs/README.md
```diff
@@ -1,3 +1,168 @@
...
+
+  - `gaTrackingId` (**required**): this is the Google Analytics tracking ID. The recommended ID to be used is: `UA-38285631-3`.
```
nkuehn:
```suggestion
  - `gaTrackingId` (**required**): this is the Google Analytics tracking ID. For all sites hosted on the `docs.commercetools.com` domain the ID must be: `UA-38285631-3`.  
```

I thought it's not required?

packages/gatsby-theme-docs/README.md
```diff
@@ -1,3 +1,168 @@
...
+    ```
+
+- `src/content`: this is where you would put your content pages as `*.mdx` files (_see [Writing content pages](#writing-content-pages)_).
+
+- `src/data/navigation.yaml`: this contains the website main navigation links. The structure of the file is a _list of chapters_ as following:
+
+  ```yaml
+  - chapter-title: This is the title
+    beta: false # (optional): will show the beta flag next to the chapter title
+    pagination: false # (optional) will remove this page from the prev/next content pagination
```
nkuehn:
```suggestion
    pagination: false # (optional) hides the prev/next content pagination at the bottom of the pages in this chapter. Use for non-linear content like reference documentation. 
```

packages/gatsby-theme-docs/README.md
```diff
@@ -1,3 +1,168 @@
...
+### Folder structure
+
+The project structure should contain at least the following files and folders:
+
+```
+├── .eslintrc.yml
+├── gatsby-config.js
+├── package.json
+└── src
+    ├── content
```
nkuehn: I thought these are autogenerated on first startup, but that maybe be something that's not ported from the last implementation yet

packages/gatsby-theme-docs/README.md
```diff
@@ -1,3 +1,168 @@
...
+    │   ├── files
+    │   └── index.mdx
+    ├── data
+    │   └── navigation.yaml
+    └── images
+```
+
+- `.eslintrc.yaml`: in case you're using a monorepository, you need to provide this file with an empty object `{}`
```
nkuehn:
```suggestion
- `.eslintrc.yaml`: in case you're using a monorepository, you need to provide this file with an empty object `{}`, otherwise provide a valid ESLint configuration. 
```

packages/gatsby-theme-docs/README.md
```diff
@@ -1,3 +1,168 @@
...
+
+- `gatsby-config.js`: this is required for a Gatsby website and should contain the website specific configuration. At the very least, the core Gatsby theme must be listed in the plugins section. You are free to provide more plugins as you need for your specific website.
```
nkuehn:
```suggestion
- `gatsby-config.js`: this is required for a Gatsby website and should contain the website specific configuration. At the very least, the commercetools Gatsby theme must be listed in the plugins section. You are free to provide more plugins as you need for your specific website.
```

packages/gatsby-theme-docs/README.md
```diff
@@ -1,3 +1,168 @@
...
+
+  Available options for the theme plugin are:
+
+  - `websiteKey` (**required**): the identifier of the website, without whitespaces. Usually this value would be the same as the `pathPrefix` without the leading slash.
+
+  - `gaTrackingId` (**required**): this is the Google Analytics tracking ID. The recommended ID to be used is: `UA-38285631-3`.
+
+    > For test websites the field should not be set.
+
+  - `createNodeSlug` (_optional_): in case you need to have more control over the creation of the page slugs, you can implement this function. This is useful if for example your website has content files in other file system locations and you want to provide a more meaningful URL path.
```
nkuehn: I don't think this should be documented so prominently.  If you want a more meaningful path, provide more meaningful folder and file structures.   I'm really struggling with refactoring of the existing site because there's such an unpredictable mix of manual and conventional URLs.

packages/gatsby-theme-docs/README.md
```diff
@@ -1,3 +1,168 @@
...
+- `beta`: to indicate if the _beta_ info message should be displayed or not.
+
+### Available elements within markdown files
+
+Besides the standard markdown syntax, the theme provides some extra elements that can be used within the `*.mdx` files.
+
+The elements should be rendered as XML tags, like HTML elements. For example:
+
+```mdx
+<MyComponent>
```
nkuehn: I suggest to only mention components that actually exist.  "MyComponent" implies understanding what a "component" is and how it comes about and how you develop it etc.  If you want to keep it generic, rather user "MyElement" because the rest of the text refers to elements.   I would prefer to stick with "Component" instead of "Element" to ease developer<->writer communication

```
<Warning>
```

packages/gatsby-theme-docs/src/overrides/README.md
```diff
@@ -1,7 +1,57 @@
-Overrides files using theme shadowing, in order to inject specific functionalities to parts of the theme.
+Overrides files using [Gatsby theme shadowing](https://www.gatsbyjs.org/docs/themes/shadowing/), to inject functionalities to specific parts of the theme.
+
+## Usage
+
+To use this feature, the website should mimic the folder structure of the Gatsby theme and list the files in the `overrides` folder that should be overridden.
```
nkuehn:
```suggestion
To use this feature, the website must mimic the folder structure of the Gatsby theme and list the files in the `overrides` folder that should be overridden.
```

packages/gatsby-theme-docs/src/overrides/README.md
```diff
@@ -1,7 +1,57 @@
...
+src
+└── @commercetools-docs
+    └── gatsby-theme-docs
+        └── overrides
+            ├── markdown-components.js
+            ├── page-header-side.js
+            └── use-additional-site-data.js
+```
+
+> Note that Gatsby shadowing works with any file in the theme. However, we want to provide an explicit extension point that consumers of the theme should use, instead of relying on internal file system paths that can potentially change any time.
```
nkuehn:
```suggestion
> Note that Gatsby shadowing works with any file in the theme. However, the theme provides an explicit stable extension point that is the `overrides` folder. Users of the theme should use this extension point only instead of relying on theme-internal file system paths that can potentially change any time.
```

packages/gatsby-theme-docs/README.md
emmenko: They are, here we only want to document what they are for. Maybe we can mention that some of them are generated on first gatsby start.

packages/gatsby-theme-docs/README.md
emmenko: >  I don't think this should be documented so prominently.

How do you suggest to document it then?

packages/gatsby-theme-docs/README.md
emmenko: Yeah good point. I wasn't sure to use "component" as maybe "element" might be more understandable for non technical people. Should we just go with the wording "component" then?

packages/gatsby-theme-docs/README.md
```diff
@@ -1,3 +1,168 @@
...
+
+Content inside the element.
+
+</MyComponent>
```
emmenko:
```suggestion
</Subtitle>
```

packages/gatsby-theme-docs/README.md
nkuehn: Since I don't see a need for the feature I would not document it at all for now.  But if you have a good use case example we can keep it and then document it here where you have it.

packages/gatsby-theme-docs/README.md
nkuehn: Yes, I would generally stick with component.  In a separate piece of documentation or training we can explain the relation to HTML and the whole idea behind MDX.   
"Element" could technically spoken also be a custom HTML element and would then have to be lowercased.

packages/gatsby-theme-docs/README.md
emmenko: It's already used in app-kit:

https://github.com/commercetools/merchant-center-application-kit/blob/bca85524b7a24329ca4c786265a90c224c9e2ad1/website/gatsby-config.js#L35-L46

That's why we have it and why It's documented 😅

packages/gatsby-theme-docs/README.md
emmenko: https://github.com/commercetools/commercetools-docs-kit/pull/90/commits/5ecd9303d365bce0f3044ac30b3f419fedde1182

Issue #70, opened 2019-11-14
Possible to declare all allowed mdx frontmatter as explicit schema ?


nkuehn:
(absolute nice-to-have just for robustness) 

Currently, if a site's content never uses a frontmatter key that is queried by our implementation, the site build fails because the queried field is not on the graphQL schema. 

The "beta"  frontmatter is such a case.  

I don't know whether it is possible at all to change the graphQL surface of other plugins (gatsbyjs is shielding plugins from each other pretty good), so this might well simply not work.

emmenko:
TL;DR: it does not work out of the box using the gatsby schema extension hooks/points.

However, I found a workaround that we could use to circumvent that problem. I'll open a PR and we can talk about it there.

PR #92, opened 2019-11-22
feat(theme): allow to configure robots noindex

### LINKED ISSUE #88: Configurable search engine include/exclude (site-wide with page override)
To be able to execute go-lives of new products and websites in a structured way it needs to be possible to configure the search indexing on a site-wide level  and override it on a per-page level. 

The site wide default should be configurable in the site's gatsby-config.js , per-page overrides should be MDX frontmatter (or use react helmet in pure JS pages)

Whether the implementation uses `robots.txt` or the HTML meta attributes or a combination is open to implementation as long as overriding works in both directions (override a page to be visible in an otherwise excluded site, override a page to be hidden in an otherwise indexed site). 

References:
 * https://support.google.com/webmasters/answer/93710?hl=en

emmenko:
Closes #88 

By default the option `excludeFromSearchIndex` is `false`, which means websites can be crawled. You need to explicitly opt-in to disable the crawling on the entire site or on a page level (only via the navigation.yaml)

image

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-node.js
+++ b/packages/gatsby-theme-docs/gatsby-node.js
@@ -33,14 +33,13 @@ exports.sourceNodes = ({ actions }) => {
       chapterTitle: String! @proxy(from: "chapter-title")
       beta: Boolean
       pagination: Boolean
-      pages: [ChapterPage!]
+      pages: [NavigationPage!]
     }
 
-    type ChapterPage {
+    type NavigationPage {
       title: String!
       path: String!
       beta: Boolean
-      excludeFromSearchIndex: Boolean
     }
   `);
 };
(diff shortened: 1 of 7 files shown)


packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -117,6 +123,9 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
         shortTitle: matchingNavigationPage
           ? matchingNavigationPage.title
           : undefined,
+        excludeFromSearchIndex:
+          excludeFromSearchIndexFromChapterPage ||
```
emmenko: The chapter page takes precedence over the plugin config.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -33,13 +33,14 @@ exports.sourceNodes = ({ actions }) => {
       chapterTitle: String! @proxy(from: "chapter-title")
       beta: Boolean
       pagination: Boolean
-      pages: [Entry!]
+      pages: [ChapterPage!]
     }
 
-    type Entry {
+    type ChapterPage {
```
emmenko: > Not sure about the "ChapterPage" name. Wouldn't it suffice to make it a "Page" ?

It's not really a page, if any it should be `NavigationChapterPage`. It's always better to be as specific as possible when naming types.

nkuehn: Right, `NavigationPage`  would be correct.  But there are no pages outside a chapter, so I think "NavigationChapterPage"  could be overspecific unless you want to futureproof the naming.

emmenko:
> I am not 100% sure but I think we decided to want all page-specific metadata as frontmatter in the MDX and not separated over navigation.yaml and frontmatter. Plus, my main use case in the current docs is to exclude a page from search that is for the same reason also not part of the navigation. So nav doesn't really help there.

Sure we can talk about it. I actually went with that option first, but bumped into issue when trying to extend graphql types from the ones generated by mdx (e.g. `MdxFrontmatter`).
This is a current limitation with Gatsby themes and plugins, there are a bunch of issues about that.
I do have a workaround though, I can show it to you tomorrow.

packages/gatsby-theme-docs/gatsby-node.js
emmenko: I still think that "chapter" needs to be in the name because the `navigation.yaml` is a list of chapters (which is also how we documented it so far, both technically and visually). Therefore, it feels more appropriate to use that.

Having long names in types is not necessarily a bad thing. Also, it does not really matter at the end because we don't explicitly use the type in our code, it's more for documentation purposes and to have a more unique name within the graphql schema.

nkuehn: OK, thanks for the explanation!

nkuehn:
@emmenko frontmatter schema predictability is a general topic, I have already created  #70 for that question. Plugins are strongly shielded in gatsby and that's actually a great thing.

PR #98, opened 2019-11-26
refactor(theme): use left-aligned layout

emmenko:
Just to get a deploy link for now...


emmenko:
FYI: we're going with this for now until we reach a solution for the "centered" layout

PR #100, opened 2019-11-26
refactor(theme): to add favicon for light/dark theme based on media query

emmenko:
Support favicons for light/dark theme


packages/gatsby-theme-docs/gatsby-ssr.js
```diff
@@ -22,4 +32,19 @@ export const onRenderBody = ({ setPostBodyComponents }) => {
         onLoad="function OptanonWrapper() {};"
       />,
     ]);
+
+  setHeadComponents([
```
emmenko: The logic here is based on the implementation of the gatsby-plugin-manifest: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-manifest/src/gatsby-ssr.js

emmenko:
> But it does not seem to work. I can only guess - can you provide static content from a theme at all?

I had a wrong path, it should work now.

PR #105, opened 2019-11-29
fix(writing-styles): disable exclamation mark, improve brand mappings

emmenko:


packages/writing-style/vale-styles/commercetools/Acronyms.yml
```diff
@@ -8,6 +8,7 @@ second: '(?:\b[A-Z][a-z]+ )+\(([A-Z]{3,5})\)'
 # ... with the exception of these:
 exceptions:
   - API
+  - AWS
```
emmenko: Is that ok for you?

packages/writing-style/vale-styles/commercetools/Brand.yml
```diff
@@ -15,3 +15,12 @@ swap:
...
   sphere.io: the commercetools platform
+  MC: Merchant Center
+  merchant center: Merchant Center
+  Merchant center: Merchant Center
+  merchant Center: Merchant Center
+  Custom Applications: custom applications
+  Custom applications: custom applications
+  custom Applications: custom applications
+  custom apps: custom applications
+  Custom Apps: custom applications
```
emmenko: Let me know if these make sense to you. Not sure if "custom applications" should also be "Custom Applications" or better lowercase.

packages/writing-style/vale-styles/commercetools/Acronyms.yml
nkuehn: Yes!  It's derived from the Google style so it's no surprise AWS is not an allowed word :-)

packages/writing-style/vale-styles/commercetools/Brand.yml
nkuehn: Good Question, It's a branding policy decision.  Technically it's correct to lowercase it unless we decide to declare them a product with an own name. 

ask @coolmilo ?

nkuehn: Plus, I suggest to also discourage "MC apps"  in this list.

packages/writing-style/vale-styles/commercetools/WordList.yml
```diff
@@ -22,7 +22,8 @@ swap:
   Ajax: AJAX
   Android device: Android-powered device
   android: Android
-  application: app
+  app: application
+  apps: applications
```
nkuehn: since you disallow "MC apps", "custom apps" in our product specific terminology, I suggest to allow "app" and "apps" as standalone terms - if we talk about iOS apps and Android apps in our documentation (eg. in the Swift SDK) that's perfectly fine.

packages/writing-style/vale-styles/commercetools/Brand.yml
emmenko: Ah thanks

emmenko:
@nkuehn please try to avoid those merge commits 😉

image

nkuehn:
@emmenko  that's what happens if you update the branch from the GitHub PR UI when it blocks merging.  

If you dislike that I suggest to think about the GitHub config.

PR #106, opened 2019-11-29
feat: plugin for introspection of mdx component placement across the site

nkuehn:
This plugin is required by the RAML driven API spec generator components to be able to generate links to other types without having to manually maintain maps of what type is documented on what page.


packages/gatsby-transformer-mdx-introspection/gatsby-node.js
```diff
@@ -0,0 +1,88 @@
...
+      let jsxId = 0; // the exact same JSX could be on one page twice, use a counter to differentiate
+      children.forEach(child => {
+        if (child.type === 'jsx') {
+          createComponentInMdxNode({
+            node,
+            createNodeId,
+            createContentDigest,
+            actions,
+            jsxString: child.value,
+            jsxId,
```
emmenko: Nitpick: use the for-each index, you don't need the counter variable `jsxId`

```suggestion
            jsxId: index + 1,
```

packages/gatsby-transformer-mdx-introspection/gatsby-node.js
```diff
@@ -0,0 +1,88 @@
...
+async function introspectMdx({
+  node,
+  createNodeId,
+  createContentDigest,
+  actions,
+}) {
+  await remark()
+    .use(mdx)
+    .use(() => tree => {
+      const { children } = tree;
```
emmenko: Nitpick: unnecessary destructure. Simply use `tree.children.forEach`

packages/gatsby-transformer-mdx-introspection/gatsby-node.js
nkuehn: Right, good hint. 
I was only superficially refactoring this piece from a for .. of  statement to find peace with ESLint.

packages/gatsby-transformer-mdx-introspection/gatsby-node.js
nkuehn: right, that's a leftover from an earlier iteration that needed more data from the tree.

PR #118, opened 2019-12-02
feat(theme): algolia search

### LINKED ISSUE #89: Algolia Docsearch box in top bar
* whether a given site has a search box should be configurable site-wide
   * necessary configuration (no secrets, these are openly visible in the website afterwards):  `apiKey` and `indexName`.  We can also choose to simply allow mixing any allowed docsearch config into a default config. 
 * We can choose to implement the `/` keyboard shortcut and taking over the current selection in a separate step. The docsearch widget has an own `bindKeyboardShortcuts` setting but it's badly implemented and worked only on US keyboards the last time we tried.  

Keep in mind that this is the special free Algolia _docsearch_  offering, not the Algolia product sold for money. 

The generic Algolia has an own set of React components, but docsearch has a pre-built widget for the docsearch offering that has to be used on docsearch sites. 

This is the Docsearch widget: https://github.com/algolia/docsearch

Here's how Facebook's docosaurus includes it in React.js:  https://github.com/facebook/Docusaurus/blob/master/packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.js - not officially open source but you can take it as an inspiration. 
 
This GatsbyJS documentation is therefore *not* applicable to this case: https://www.gatsbyjs.org/docs/adding-search-with-algolia/

emmenko:
Closes #89


nkuehn:
FYI here's the current website's integration code which implements a few details we could carry over:

https://github.com/commercetools/docs.commercetools.com/blob/master/_assets/javascripts/main_script.js#L40-L87 

 * react on "/" keyboard and take over the selected text
 * don't send to https://docs.commercetools.com  but to the current local domain (could be reconsidered)
 * disable algolias own keybinding which works only in US keyboards. 
 * maybe more..

PR #129, opened 2019-12-07
fix(theme): SSR style tags messing with nth selectors

### LINKED ISSUE #121: Flash of unstyled content (regression)
The current master has a flash of unstyled content (tested on a windows 10 machine in Firefox and Chrome). 

The vertical spacing both in the left hand nav and the body content is initially higher and gets reduced after page load.

emmenko:
Fixes #121 

SSR (Server Side Rendering) works out-of-the-box in Emotion, however the [default approach](https://emotion.sh/docs/ssr#default-approach) is to insert a `<style>` tag above each element with styles.

This has a problem though, as pointed out in the Emotion docs:

> Warning: This approach can interfere with nth child and similar selectors as it is inserts style tags directly into your markup. You will get a warning if you use such selectors when using this approach.

This is exactly the root of the issue we're seeing, as we make use of the [lobotomized owl selector](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/).

To fix this problem, Emotion [suggests to implement SSR on our own](https://emotion.sh/docs/ssr#advanced-approach), following certain guidelines, with the goal of avoiding the `<style>` tags to be injected next to the elements.

### DIFF
--- a/packages/gatsby-theme-docs/create-emotion-cache.js
+++ b/packages/gatsby-theme-docs/create-emotion-cache.js
@@ -1,0 +1,8 @@
+// See https://emotion.sh/docs/ssr#gatsby
+import createCache from '@emotion/cache';
+
+// To be used in gatsby-ssr.js
+export const createDocsCache = () => createCache();
+
+// To be used in gatsby-browser.js
+export const docsCache = createDocsCache();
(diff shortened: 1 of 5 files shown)


emmenko:
> I am not able to evaluate the code quality or whether this is the most effective approach

I've been following the emotion + gatsby documentation basically, nothing more.

PR #131, opened 2019-12-09
refactor(writing-styles): add postinstall script to download vale binary based on OS platform

emmenko:
This PR further improves the `writing-styles` package.

I noticed that the `vale` binary files takes up a few megabytes each, causing the package to be around 25Mb. This is a bit annoying when publishing and installing the package, as it just takes longer.

We don't need to ship the binary with the package, instead we can use a `postinstall` script to download the binary "on demand" upon package installation, according to the OS platform. This also ensures that we only download the proper binary and not all of them.


packages/writing-style/.gitignore
```diff
@@ -1 +1,3 @@
-*.zip
\ No newline at end of file
+*.zip
+*.tar.gz
+bin/vale
```
emmenko: We don't put it in version control anymore

packages/writing-style/scripts/download-vale.js
```diff
@@ -0,0 +1,37 @@
...
+        `Cannot download vale binary, unsupported platform: ${process.platform}.`
+      );
+  }
+})();
+
+const archivePath = path.join(__dirname, `../${binaryFileName}`);
+const destPath = path.join(__dirname, '../bin');
+
+console.log('[writing-styles] Verifying vale installation...');
+if (fs.existsSync(path.join(destPath, 'vale'))) {
```
emmenko: We don't download it again if a binary file is already there.

packages/writing-style/scripts/download-vale.js
```diff
@@ -0,0 +1,37 @@
...
+  console.log(
+    '[writing-styles] Vale binary already installed, skipping installation...'
+  );
+} else {
+  shelljs.exec(
+    `curl -sLO https://github.com/errata-ai/vale/releases/download/v${version}/${binaryFileName}`
```
nkuehn: This is a linux & mac os specific implementation, the curl command is not available OOTB on a windows installation.  Rather use node-fetch or direct node.js  HTTPS?

packages/writing-style/scripts/download-vale.js
```diff
@@ -0,0 +1,37 @@
...
+  );
+  shelljs.exec(`tar -xzf ${archivePath} -C ${destPath} vale`);
```
nkuehn: Same here, you'll have to switch by platform.  
Unfortunately there's no OOTB unzip on Windows. There are pure Node implementations like  https://www.npmjs.com/package/extract-zip   or crazy Powershell based command lines that work from Windows 8 upwards but.

packages/writing-style/scripts/download-vale.js
emmenko: Hmm good point. I’ll see if there is a cross platform package to do that, otherwise we can keep things as they are for now

emmenko:
@nkuehn alright please have another look, I re-wrote the script to be cross-platform compatible.

packages/writing-style/scripts/download-vale.js
emmenko: https://github.com/commercetools/commercetools-docs-kit/pull/131/commits/924649dfe205b85489a71b8b7a3c82cc020746e9

packages/writing-style/scripts/download-vale.js
```diff
@@ -0,0 +1,89 @@
+const os = require('os');
+const fs = require('fs');
+const path = require('path');
+const fetch = require('node-fetch');
+const shelljs = require('shelljs');
+const { path7za } = require('7zip-bin');
```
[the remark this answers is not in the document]
emmenko: > we're doing a refactor to get rid of 8-10MB sized binaries committed and to be able to do that we pull a package for unzipping them that is by itself a 9MB binary committed into NPM

Yes it's a bit unfortunate but it's still better then before for the following reasons:
- `7zip-bin` is an external dependency, so we don't need to "deal" with the package size directly
- our current package is `46.3`MB as opposed to the `7zip-bin` which is only `8.84`MB

image

image

- our package now is just a few kb, which makes publishing faster (and less error prone)

So I still think there is a good benefit of doing this.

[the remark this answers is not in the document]
nkuehn: OK, it's not worth investing another hour to make it completely.  It's odd-ish that uncompressing is so cumbersome in node, didn't expect that.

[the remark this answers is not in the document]
emmenko: The problem is not much about node, but uncompressing in windows

PR #135, opened 2019-12-10
refactor(broken-link-checker): allow to pass excluded-keywords option

emmenko:
We need this option for the migration phase, as some links are detected as broken.

For example, a link to https://docs.commercetools.com/http-api-authorization is detected as broken, as we assume that this is an "internal" link and therefore remove the domain origin. However, because of the "legacy" URL structures, this is a valid link that will only work on the live domain.

https://github.com/stevenvachon/broken-link-checker#excludedkeywords


emmenko:
As a follow we can also allow to configure the link checker using a config file, then we will barely need to touch the internals every time we want to add a new option.

PR #145, opened 2019-12-11
feat: migrate base theme for api docs

emmenko:


packages/gatsby-transformer-mdx-introspection/gatsby-node.js
```diff
@@ -54,7 +49,12 @@ async function introspectMdx({
         }
       });
     })
-    .process(node.rawBody);
+    .process(node.rawBody, (error, file) => {
+      if (error) {
+        console.log(String(file));
+        throw error;
```
emmenko: We need to handle this error otherwise the error it's swallowed and the process exits without any log.

packages/gatsby-transformer-mdx-introspection/src/generate-component-node-object.js
```diff
@@ -4,6 +4,14 @@ const generateComponentNodeObject = jsxString => {
   const jsxAsDom = parse5.parseFragment(jsxString, {
     scriptingEnabled: false,
   });
+
+  // In case we don't get any childNodes, we return early.
+  // Example:
+  //   { nodeName: '#document-fragment', childNodes: [] }
+  if (jsxAsDom.childNodes.length === 0) {
```
emmenko: This happens for example for `</SubTitle>`. For now, we simply return and don't create any node.

cc @nkuehn @davifantasia

PR #150, opened 2019-12-12
feat: extract ui components into a uikit package

emmenko:


packages/ui-kit/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "prism-themes": "1.3.0",
+    "prismjs": "1.17.1",
+    "prop-types": "15.7.2",
+    "tiny-invariant": "1.0.6"
+  },
+  "devDependencies": {
+    "@commercetools-uikit/design-system": "10.13.0",
+    "@commercetools-uikit/icons": "10.13.0",
+    "@commercetools-uikit/spacings-inline": "10.13.0",
+    "gatsby-plugin-google-analytics": "2.1.31"
```
nkuehn: I will try to test whether it works.   The original reason why I proposed a separate ui kit was to have a package that is 100% safe to not pull in gatsby plugins or gatsby because gatsby seems to run its build magic over all the dependencies and not just the plugins that are actually configured in gatsby-node.js

emmenko: Yeah I'm also not 100% sure about this. I guess we can leave the `Link` into the theme package for now? 🤔

emmenko: On a second thought, it’s a bit tricky here.
The Link component is mapped to the mdx provider and the list of mapped components is also exposed from the kit. This would be useful for the “playground editor”.

The Gatsby plugin for GA here is only used because of the OutboundLink component, so there is nothing about Gatsby going on here.
Maybe we can just extract the GA logic into our own Link component and remove the plugin dep.

nkuehn: If it's worth the effort I think the consequential way would be to place a plain styled non-gatsby "dumb"  Link component  into the ui-kit (providing the css style as an export too) and replace it with the full Link component in the theme. 

The Link component is very opinionated and assumes a path prefixed URL structure with multipe sites and more.  That's not the idea of a UI kit to me.  

If the playground editor uses a simpler Link that does not track outbound links and does not do so much "magic" that's no problem to me.  Explaining the URL conventions doesn't really work as a playground anyways because the resulting links look all the same anyways (and even the DOM looks near identical)

emmenko: Alright, I’ll refactor this part then.

emmenko:
@nkuehn @davifantasia please have a another look, thanks

PR #154, opened 2019-12-12
refactor(theme): do not render search bar when excludeFromSearchIndex is true

### LINKED ISSUE #149: Default to not showing the on-site search box if the whole site is configured "Noindex"
If a whole microsite is configured to be excluded from search engines, its content will not be indexed in the on-site search, too. 
To avoid user confusion, such sites should not have the on-site search box because such a box suggests that the topic area you are currently reading is indexed.  

Independently, we can consider configuring whether a top menu is shown or not.   In total, it's about a "Private Beta" mode that has the following effects:
 - beta on site level
 - noindex on site level
 - no search box 
 - no top menu  (optional but we can hardwire it as a convention, too)

emmenko:
Fixes #149

### DIFF
--- a/packages/gatsby-theme-docs/src/layouts/content.js
+++ b/packages/gatsby-theme-docs/src/layouts/content.js
@@ -22,7 +22,10 @@ const LayoutContent = props => {
   const siteData = useSiteData();
   return (
     <LayoutApplication>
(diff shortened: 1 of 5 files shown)


emmenko:
Yeah ok, will take a look at it tomorrow

emmenko:
Alright, it should be good to go now.

PR #155, opened 2019-12-12
refactor(theme-api): no hardcoded transformer raml options, instead forward options

### LINKED ISSUE #148: Move hardwired API names from api-docs theme to per-site config
gatsby-theme-api-docs  configures the legacy raml parser with our default standards, which is good.  But one option should be configured on site level, that's the `includeApis` whitelist. 

https://github.com/commercetools/commercetools-docs-kit/blob/master/packages/gatsby-theme-api-docs/gatsby-config.js#L17

emmenko:
Fixes #148


packages/gatsby-transformer-raml-legacy/gatsby-node.js
```diff
@@ -11,6 +11,7 @@ const defineSchema = require('./src/schema/define-schema');
 
 async function onCreateNode(
   { node, actions, createNodeId, createContentDigest },
+  // Plugin options passed through `{ transformerRaml: {} }`
```
davifantasia: The plugin should not care how the option is passed to it, so I think think comment can be confusing.

emmenko:
> It appears you forgot to pass the raml options from the api-docs-smoke-test

Ah yeah, should we pass all the options as before? Or some can be omitted?

Feel free to push a commit since you have more knowledge of how the options work.

packages/gatsby-transformer-raml-legacy/gatsby-node.js
emmenko: Right, we can remove it then.

PR #156, opened 2019-12-12
feat(broken-link-checker): warn on insecure links

### LINKED ISSUE #74: Broken Link Checker: Warn or throw errors on insecure links (`http://*`)
To gain more independence of Netlify's analysis features it would help if the broken link checker would warn on unencrypted URLs.  It's checking all source links of images, scripts etc, too so that's a relatively broad coverage.

emmenko:
Fixes #74

### DIFF
--- a/packages/broken-link-checker/src/run.js
+++ b/packages/broken-link-checker/src/run.js
@@ -15,12 +15,9 @@ const run = ({ server, port, config, checkerOptions }) => {
     html: (/* tree, robots, response, pageUrl, customData */) => {},
     junk: (/* link, customData */) => {},
     link: (link /* , customData */) => {
-      if (!link.internal) {
-        const { protocol } = new URL(link.url.original);
-        if (protocol === 'http:') {
-          console.warn(`Unsecure external link : ${link.url.original}`);
-          console.warn(`   | used on: ${link.base.parsed.pathname}`);
-        }
+      if (!link.internal && link.url.parsed.protocol === 'http:') {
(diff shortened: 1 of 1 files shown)


packages/broken-link-checker/src/run.js
```diff
@@ -15,9 +15,16 @@ const run = ({ server, port, config, checkerOptions }) => {
     html: (/* tree, robots, response, pageUrl, customData */) => {},
     junk: (/* link, customData */) => {},
     link: (link /* , customData */) => {
+      if (!link.internal) {
+        const { protocol } = new URL(link.url.original);
```
nkuehn: I think you don't need to re-parse here - not tested it now but I remember `${link.url.parsed.protocol}` is available (`link.url.parsed` is already an URL object or at least like one).

packages/broken-link-checker/src/run.js
```diff
@@ -15,9 +15,16 @@ const run = ({ server, port, config, checkerOptions }) => {
...
+        if (protocol === 'http:') {
+          console.warn(`Unsecure external link : ${link.url.original}`);
```
nkuehn:
```suggestion
          console.warn(`Insecure external link : ${link.url.original}`);
```
One of the illogical parts of english: There is no "unsecure", but there is "unsecured" and there there  is "insecure" but no "insecured".   Don't ask, it's as explainable as most of german :-)

emmenko: Ah right, thanks :)

PR #162, opened 2019-12-13
fix(ui-kit): bundle package using rollup

emmenko:
I forgot that the ui-kit library should be bundled, otherwise imports will break.

The setup uses Rollup, and is very similar to what we do in app/ui-kit.


emmenko:
What do you mean? This is how you usually set up the tooling in a monorepository

PR #170, opened 2019-12-17
refactor(theme): to change layout to a full (content) page scroll

emmenko:


emmenko:
Agreed with Nikolaus that we can merge this as this is the direction we want to go. This way we don't have a pending PR for weeks.
We'll work on the UI details with the design team once they have some proposals.

PR #174, opened 2019-12-18
feat(ui-kit): export only ContentNotification

davifantasia:
I would like to use the `ContentNotification` in a couple of the components on the `gatsby-theme-api-docs`. To avoid potential bloat caused by importing multiple components, exporting only the `ContentNotification` is useful.

### DIFF
--- a/packages/ui-kit/src/components/content-notifications.js
+++ b/packages/ui-kit/src/components/content-notifications.js
@@ -56,7 +56,7 @@ const Container = styled.div`
   min-width: 100%;
 `;
 
-const ContentNotification = props => {
+export const ContentNotification = props => {
   const Icon = getIconByType(props.type);
   const iconColor = getIconColorByType(props.type);
   return (
--- a/packages/ui-kit/src/index.js
+++ b/packages/ui-kit/src/index.js
@@ -3,8 +3,10 @@ import * as Markdown from './components/markdown';
 
 // components
 export { designSystem, Markdown };
-export { ContentNotification } from './components/content-notifications';
-export { default as ContentNotifications } from './components/content-notifications';
+export {
+  default as ContentNotifications,
+  ContentNotification,
+} from './components/content-notifications';
 export { default as Globals } from './components/globals';
 export { default as Link } from './components/link';
 export { default as Reset } from './components/reset';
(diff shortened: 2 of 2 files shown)


davifantasia:
@emmenko , @nkuehn , at the moment when you do this in gatsby-theme-api-docs:

`import { ContentNotifications } from '@commercetools-docs/ui-kit'`

Three instances of `ContentNotification` is being imported into that package, for my use case I need only ContentNotification.

Generally speaking, I guess the question from me is do we want to stick with a pattern of exporting multiple components (or instances of components) even if others are not need by the requesting application. To me, this looks like a recipe for bloat, unless that is what we want, then I'm fine with it.

davifantasia:
> There's not bundle size advantage either because the API docs theme does require the docs theme and that uses the full docs UI kit anyways.

I'm not so good with Webpack but I expect the bundling of each theme to cause a separation of the number of components, and eventually, size of all components in each theme.

emmenko:
The bundle size is not a problem here, there is no duplication.

Can you explain though why you need the base component and how you intend to use it?

emmenko:
It might only be less ideal if you only intend to use one component in your app, but since we use all of them there is no size penalty

davifantasia:
> Can you explain though why you need the base component and how you intend to use it?

Please see these files:
- https://github.com/commercetools/commercetools-docs-kit/blob/rmf-codegen/packages/gatsby-theme-api-docs/src/components/api/type/type.js (line 21)
- https://github.com/commercetools/commercetools-docs-kit/blob/rmf-codegen/packages/gatsby-theme-api-docs/src/components/api/resource-method.js (line 21)
- https://github.com/commercetools/commercetools-docs-kit/blob/rmf-codegen/packages/gatsby-theme-api-docs/src/components/api/resource/resource.js (line 23)

emmenko:
So from what I can see you don't need the base component. You were just concerned with the "bundle size"?

davifantasia:
I need only the base component.

emmenko:
What do you need it for? In the links you posted you use the `ContentNotifications.Error`.

davifantasia:
> What do you need it for? In the links you posted you use the `ContentNotifications.Error`.

I don't know how else to explain. But I can just close this PR if you think it's not necessary.

PR #177, opened 2019-12-19
fix: corrected expected order of properties

davifantasia:
Corrected the descriptions of the expected order, as "Should display 3rd" was missing.

I also included `OutOfOrderPropertiesTestType.properties.version`.

### DIFF
--- a/websites/api-docs-smoke-test/src/api-specs/test/api.raml
+++ b/websites/api-docs-smoke-test/src/api-specs/test/api.raml
@@ -302,51 +302,54 @@ types:
     properties:
       eigthProperty:
         type: string
-        description: Should display 7th (1st in RAML)
+        description: Should display 8th (1st in RAML)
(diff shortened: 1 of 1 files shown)


davifantasia:
> This is changing the test assertion but not improving the test quality. (internal) Ticket DOCS-199 defined the ordering convention as follows:
> 
> * id, version, key, createdAt, lastModified  are ALWAYS in front in this order. This even allows for custom visual layout of this block.
> * "custom" is always last.
> * retain the order for the rest.
> 
> To me it looks like the real issue is that the `version` property is not part of the test model..

I'm not sure what you mean, I did not change the order, I just corrected the text descriptions, as "Should display 3rd" was missing.

Also I don't see version included in the list of properties, of course I can add that if it's supposed to be there.

nkuehn:
Great, now it matches the defined behavior!

PR #178, opened 2019-12-19
feat: updated auto generated raml from rmf-codegen

davifantasia:
This contains:

- fixes for missing data in a type
- resources files are now annotated for identification and each file starting with `# Resource`


websites/api-docs-smoke-test/src/generated-api-specs/test/resources/ByProjectKey.raml
```diff
@@ -1,4 +1,4 @@
-#ByProjectKey
+(resourceName): ByProjectKey
```
nkuehn: It works, but I think a custom comment line would be the cleaner way to do this than assuming a specific property being on top although object semantics don't guarantee property order.

websites/api-docs-smoke-test/src/generated-api-specs/test/types/EnumPropertiesType.raml
```diff
@@ -6,23 +6,35 @@ description: |-
...
-    format: int16
+    format: float
     type: number
     (builtinType): number
     description: |-
       An int with enum values
+    enum:
+    - 1
+    - 2
+    - 3
```
nkuehn: Something wrong here?  The test case's name is explicitly "an int with enum values" but the generated RAML declares a number with float format.

davifantasia:
@jenschude, please see Nikolaus' comments.

websites/api-docs-smoke-test/src/generated-api-specs/test/types/EnumPropertiesType.raml
jenschude: The issue is coming from the input raml.

The type is defined as number without any format. And number defaults to float. So either declare the type to be integer or add a format.

websites/api-docs-smoke-test/src/generated-api-specs/test/resources/ByProjectKey.raml
jenschude: When you read this with yaml, the property will always be there. And you can directly access it. Accessing the comment may be more tricky. Cause the yaml reader will not put comments in the object.

websites/api-docs-smoke-test/src/generated-api-specs/test/types/EnumPropertiesType.raml
nkuehn: Test RAML bug number 10 or so.  It's scary how easy it is / was to write wrong RAML without noticing.

websites/api-docs-smoke-test/src/generated-api-specs/test/resources/ByProjectKey.raml
nkuehn: @jenschude  We're filtering the RAML files that have to be read with a low-level  "read the first line from this file" implementation that doesn't even read the full file into memory.  

That is / was necessary to not handle the tons of untyped fragments and includes in the original RAML, parsing them just to decide that it's not a root API definition was too slow. 

That's why the implementation requires literally the first line of a file to be read analogous to shebangs and other "Magic numbers" in the first bytes of a file.

jenschude: So for the generated RAML files I can guarantee that this will be also at the first line. But I could also just duplicate it in a comment. Doesn't matter for me. ;)

websites/api-docs-smoke-test/src/generated-api-specs/test/types/EnumPropertiesType.raml
jenschude: Nah any integer is actually a valid float. So it's not necessarily wrong ;)

websites/api-docs-smoke-test/src/generated-api-specs/test/resources/ByProjectKey.raml
nkuehn: thanks!  It's more consistent parsing and a better understandable codebase for us if everything is "comment in first line" based

websites/api-docs-smoke-test/src/generated-api-specs/test/types/EnumPropertiesType.raml
nkuehn: Let's not start that discussion here  :-)

PR #179, opened 2019-12-21
 refactor(theme): to render search results in an overlay, aligned with the page content

emmenko:


emmenko:
Alright, Sven said it's ok now. @nkuehn would you like to have a final look?

nkuehn:
@emmenko  I really like the way you interact with it, thanks!   The only minor complaint I find is that it has a slight height flicker when activating the search box for the first time.  I guess that's the algolia widget initializing itself empty.  Maybe reserve a minimum height?

emmenko:
@nkuehn alright, I changed a bit how we load the search component. We lazy load it and use React.Suspense to determine if a loading state should be shown or not.

Below you can see it with an artificial delay of 2 sec:

emmenko:
I'll merge once CI passes

PR #183, opened 2019-12-23
feat(introspection): utilize Babel parser to better parse MDX into component nodes

jazevedo620:
## Overview

This pull request is essentially a re-write of the `gatsby-transformer-mdx-introspection` plugin that modifies the approach from using a `remark`/`parse5` HTML parsing approach to a fully featured MDX/JSX parsing pipeline via `@mdx-js/mdx` and `@babel/parser`. At a high level, the plugin now:

1. Compiles the raw MDX files and creates an intermediate JSX form (the same that is queried via `Mdx.body`)
2. Parses the JSX into an AST using Babel
3. Locates the root component of the main MDX content (always `<MDXLayout>`)
4. Reduces the verbose AST into a simpler component tree where each node is a JSX/HTML element (or a text node)
5. Exposes this component tree via graphql using a backwards-compatible schema

Because of this, now **all** components and HTML elements (other than those configured to be excluded) will result in a `ComponentInMdx` node that can be queried: not just those at the top level. Additionally, attributes are exposed in a way that tries to use their actual JSX value if simple enough.

### This PR Contains

- [x] JSDoc comments on added functions
- [x] Re-written unit and integration tests
- [x] Updated `README.md`
- [x] Linter-passing code

### Querying Sub-Components

Another area this PR seeks to improve is querying sub-components in the component tree. The GraphQL Node type includes two fields: `childComponentInMdx` and `childrenComponentInMdx` that support filtering/sorting over the children (and, optionally, descendants if `deep` is set to `true`) of the current node. This allows for advanced use cases to be serviced, such as getting all link descendants of header elements.

### Performance Notes

Since this change makes the process of parsing MDX and generating component nodes much more heavy-weight than it was originally, there are a few solutions I have used to lessen the impact:

- Caching of generated component tree on a per-file basis using Gatsby's plugin cache API
- Set of HTML elements/JSX components to *exclude from generating Gatsby `ComponentInMdx` nodes for*, which is configurable via `excludeTags` in the plugin options. This is because certain elements, like `p` or `tr/td` get generated implicitly by the markdown->JSX conversion and would otherwise result in a high volume of `ComponentInMdx` nodes
- Predicate function in plugin config at `shouldIndexNode` designed to be used as an escape hatch in case only a small subset of total MDX pages need to have their components indexed. Defaults to `true` for every node

## Breaking Changes

This PR tries to keep the same functionality that was present in the original version (it includes legacy integration tests to ensure this). However, one key area in which it is different is the **casing of JSX attributes/tags**. This is due to dropping the HTML parsing engine, which automatically converted the case. Now, all components/attributes will use the source casing from the compiled MDX, which is usually Pascal/camelCase. The original behavior can be restored by setting `lowercaseIdentifiers` to true in the plugin options. **This was already done for `gatsby-theme-api-docs`**, which was the only dependent I could find in the repo.


nkuehn:
Hi @jazevedo620 , happy new year and congratulations to being the first contributor outside the team that maintains this repository!

You have invested quite significant effort, so to avoid misunderstandings when reviewing it would be great to hear a bit of background - what was the use case that motivated so much work?  ( I can guess it from the example in the README, but would feel better to hear from you). 

I'm asking because the initial design consideration behind the MDX introspection plugin was that the functionality would actually belong directly into [gatsby-plugin-mdx](https://www.gatsbyjs.org/packages/gatsby-plugin-mdx/) as an optional or dynamically evaluated GraphQL feature but I could not invest the necessary effort to contribute and therefore chose the simple-but sufficient-for-me approach building a minimal plugin using a DOM parser.

jazevedo620:
Yeah, sure. I appreciate you taking the time to review this PR, and I don't mind giving a bit more information on my motivating use case. The primary reason behind the changes was to enable using MDX components/shorcodes as a way of embedding structured content that can later be queried at a more granular level than just "all of the content" (which is what the official MDX plugin normally does).

## Use Case

One of the projects I work on is a (Gatsby-based) documentation site for a small project ([architus/docs.archit.us](https://github.com/architus/docs.archit.us/)), where we document our project and its components using MDX. In developing and documenting our application, we wanted to use MDX to detail our API routes, but were torn between choosing to either:

1. Embed API route information in longer-form copy that describes its purpose and fit within the larger project/ecosystem (overall, the routes would be split between multiple pages)
2. Provide an API route index page that includes all available API routes

### API Documentation Schema

In addition, we were unsure how to encode our API routes: we could have used an API documentation language, but we haven't incorporated that into our normal workflow so it would only be used for the docs site. Alternatively, we could have used some structured markup language like YAML or JSON, but that would make it difficult to use Markdown/shortcodes inside fields like route descriptions, where writing something like `see [link](/internal/path)` is much easier than the longform `see <a href="/internal/path">link</a>`.

Instead of either of these options, we desired a system that combines:

1. The ability to be structured like YAML or JSON
2. The ability to include markdown content to make writing copy easier

From this, it was suggested that we try using MDX to solve the task, which resulted in an example schema that we came up with. This allowed us to bring together our two desired qualities, and had the added benefit of being homogeneous with the language we already write our documentation in.

<details><summary>Example Schema</summary>
<p>

```jsx
<Route method="post" path="/session/refresh" category="Session" auth>
  <Route.Name>Refresh Session</Route.Name>
  <Route.Description>

This route is a part of the [session management cycle](/internal/session) and is designed to be called to refresh the user's current session after the `refreshAfter` field has elapsed since last session issuance/refresh.

  </Route.Description>
  <Route.Responses>
  <Response code={200} label="Successful refresh" type="json">
  <Json.Object>
  <Json.Scalar key="refreshAfter" type="Integer" optional={false} description="Time (in seconds) to send another refresh after" />
  <Json.Scalar key="expiresIn" type="Integer">

Session-related *refresh time*

  </Json.Scalar>
  </Json.Object>
  </Response>
  <Response code={401} label="Unauthorized" type="text">
  Not authorized
  </Response>
  </Route.Responses>
  <Route.Examples>
  <Example>
  <Example.Request method="post" path="/session/refresh" />
  <Example.Response code={200} note="Successful refresh">
  { { refreshAfter: 0, expiresIn: 0 } }
  </Example.Response>
  </Example>
  </Route.Examples>
</Route>
```

</p>
</details>

#### Linking & Referencing Content

Another added benefit of this approach would be to enable a system of linking and referencing content that is commonplace in typesetting languages like `LaTeX`. This would become especially useful for allowing us to have both API route documentation systems simultaneously: we could have API routes split between docs pages in copy, with custom Gatsby code that would find all API routes and catalog them in a single page.

### Researched Solutions

To accomplish the above schema and allow for parsing/duplication of content, we looked at a few different solutions along the way.

1. The first-party [gatsby-plugin-mdx](https://www.gatsbyjs.org/packages/gatsby-plugin-mdx/) has a feature to query for the MDX AST it generates internally, which looked promising at first. However, upon closer inspection, the AST it provides via the `Mdx.mdxAST` field doesn't parse inline JSX, and instead leaves them as `type: "jsx"` nodes in the AST. Even worse, it handles interleaving markdown and JSX very poorly, opting to create separate nodes with the opening and closing tags and having a third node for the markdown content in MDX code like the snippet below:

    ```jsx
    <Alert type="info">

    **Informative** content here, such as general tips or bits of info

    </Alert>
    ```
2. We also found this plugin as an option, which could actually parse some JSX elements and find their attributes without breaking on interleaved markdown. However, there were a few features missing that we needed to fit our use case:
    - Ability to query nested nodes that are children of other defined tags, such as getting the children of `<Route>` in the example schema above
    - Ability to get attributes that were specified as simple JSX expressions, such as numbers, booleans (or their shorthand syntax), and JSON objects
    - Ability to retrieve the contents (as plaintext) to retrieve values such as `<Route.Name>` in the above schema
    - The transformation of attributes/tags to lowercase was also jarring, and while it could be worked around, it wasn't ideal
3. Lastly, we considered making a new plugin to implement our desired features, but decided that it's feature set would overlap heavily with the features of this plugin, so I implemented the features in this plugin instead.

## Implementation Details

With our desired use case in mind, I set out to modify the internal parsing of the plugin to use a full JSX parser to accommodate the extended syntax JSX supports when compared to HTML (like non-string attribute values).

### MDX Parsing

Early on, I decided to re-use the MDX parser, which has a few added benefits:

1. I didn't have to worry about the difference between markdown content/syntax like `see [link](/internal/path)` vs traiditional JSX syntax like `see <a href="/internal/path">link</a>`. This meant I could focus only on parsing/transforming JSX, and ignore the fact that it originally came from markdown/MDX
2. The plugin is less sensitive to changes in the MDX language specification, as it only ever examines JSX (with one exception, frontmatter, which I talk about later). This means that as long as the MDX parser package outputs raw JSX code, the plugin should still be able to parse its structure.

    The one brittle part of this process is the need for the JSX AST (which contains all the beginning `import`s and the function declaration) to be traversed until it finds the JSX in the function body, which is in an `<MDXLayout />` component. The main assumption is that the entire content should appear as children of this node, which could change in the future (as it's an internal implementation detail of the MDX -> JSX transformation)

The resultant JSX code is then piped directly into the Babel parser (with JSX enabled), which produces a Babel AST that includes all of the content nodes.

### Reducing the JSX AST

The one problem with the Babel output is that it is very verbose and complicated to use. Since Babel outputs a literal AST, all of the peculiarities of the way components are declared, indented, and used are reflected in the output, when the desired output is much more high-level. With this in mind, the other main functionality of the plugin is to transform/**reduce** the Babel JSX AST into a simpler *reduced component tree*, which contains information about the structure of the component tree, akin to the output of the HTML parser used in the original plugin.

#### Notable Details

- As much as possible, the plugin attempts to deserialize expressions into their direct javascript values, with basic support for
  - Primitive literals literals (string, number, boolean, null, undefined)
  - Object literals (with their structure recursively traversed to try to parse as many simple key-value pairs as possible)
  - Spread JSX attribute syntax - I wasn't originally going to implement this (as it is generally pretty niche and not used with the simple literal values that the parser supports anyways), but I noticed that a lot of the emitted JSX uses it in components/tags sourced from block or inline markdown elements, so I added support for basic uses
  - Shorthand boolean attributes
- For expressions it cannot reasonably deserialize into their proper representations, the plugin will instead pull the original JSX code from the emitted document and use that string as the value
  - This results in function attribute values having their values in the reduced component tree be literal strings of their source code
  - For shorthand object method syntax like `{ method() { } }`, the emitted attribute becomes `{ name: "method", value: "function() { }" }` to be consistent
- The MDX parser automatically inserts certain attributes, such as `mdxType` and `parentName`, which are automtically removed by the plugin to prevent confusion. This behavior can be disabled in the options, though
- I noticed while smoke-testing the plugin against my own set of MDX files that the frontmatter would get mis-parsed as a thematic break, a series of text, and then a header as the last entry if it was passed into the MDX parser. To prevent this, I included `graymatter` as a dependency to remove the frontmatter from the raw MDX before it is piped into the MDX parser.
- It is possible for JSX to contain what I called "detached heads" of JSX components, where JSX is embedded inside a complex JSX expression:
  ```jsx
  <div>
    {/* IIFE example */}
    {(function (){
        return <h1>I'm inside an IIFE!</h1>;
    })()}
  </div>
  ```
  While this could be ignored, it would be parsed anyways by the Babel, so I figured I would include these components in the output to support other use cases. These appear as additional reduced component trees, where the root is the first encountered JSX element inside a javascript expression. This is why many of the variables call the output a "**reduced forest**", because the output could potentially have multiple detached trees in addition to the primary component tree.

  ##### Resultant output for above snippet

  ```js
  "childrenComponentInMdx": [
    {
      "component": "div",
      "attributes": [],
      "tree": [
        "function () {\n return <h1>I'm inside an IIFE!</h1>;\n }()"
      ]
    },
    {
      "component": "h1",
      "attributes": [],
      "tree": [
        "I'm inside an IIFE!"
      ]
    }
  ]
  ```

### Gatsby Interface

With the reduced component tree created, the last step was to expose it via Gatsby's GraphQL API to plugin users, and I decided to leverage the first-party support that the data store has for tree-like data structures to expose the components in component trees (and their children). The final implementation included the following fields on each `ComponentInMdx` node:

- `component` - Same as before, though now uses the original casing
- `attributes` - Also same as before, but the values can potentially include simple javascript expressions (or JSX substrings)
- `jsx` - Original JSX code that the component comprises (uses Babel's start/end positions to accomplish this)
- `isRoot` - Boolean flag used to flag whether components are at the root level of the document, or are the roots of their own detached trees
- `tree` - Direct component tree children, used as an "escape hatch" out of the Gatsby node model in case users want to examine the children component nodes directly
- `ast` - Original Babel AST node that caused this component to be emitted. *In order to reduce the heap cost of keeping this in the data store (and in the cache), I made this field opt-in by setting `attachAST` to `true` in the plugin options*
- `mdx` - "Parent" MDX file node (not the same as the `parent` in case the current node is a nested component, in which case the `parent` is another `ComponentInMdx` node)
- `content` - Direct plaintext children of the current node, which is lazily evaluated. This tries to replicate the behavior of the final browser output (with formatting removed), though the way whitespace is handled in the browser is likely different
- `childComponentInMdx` and `childrenComponentInMdx` - I originally, wasn't going to make these (and have the main way of introspecting children be `tree`), but it seemed like a waste to not leverage the Gatsby node system, so these are filterable/sortable fields that basically replicate the behavior of `allComponentInMdx`, but only look at a component's children. In addition, the search can be made to look at all indirect children/descendents too, by setting the `deep` flag to `true` as an input to the field.

### Other Notes

- As I mentioned above, whitespace is something that the plugin tries to collpase as much as possible. To do this, it collapses consecutive horizontal whitespace characters (tabs, spaces, etc.) into a single space, and separately collapses consecutive vertical whitespace characters (line feeds and carriage returns) into a single newline/line feed `\n`. In addition, it left trims the first text child of a node, and right trims the last text child. Since this might mangle intentional spacing in the source document, this behavior is configurable in the plugin options.

## Expanded Use Cases

With the above implementation in mind, there are a few additional use cases that these changes support:

- Referencing features like those in `LaTeX` or something similar: duplicating entire figures automtically by finding their originals and exposing their contents to the referencing components using something like context. This could also be used for specifying an entity or figure by some global ID, which could resolve to a cross-page (or cross-site) link to that entity.
- Since the plugin doesn't discriminate between links that are results of markdown syntax or JSX syntax, the plugin could be used to get all links contained in an MDX document, and do something with them. A really useful functionality that this enables (and which I plan on implementing in my own project) is the ability to use this "index" of all internal and external links to perform broken link checking at build time.

**P.S.** In writing this rationale comment, I realized that the `detached head` functionality was missing from the tests, so I am going to quickly add that, and that there is also an error that gets outputted upon plugin load (which I thought I had removed in an earlier commit, but oh well), so I'm going to remove that quickly as well:

```
warn On types with the `@dontInfer` directive, or with the `infer` extension set to `false`, automatically adding fields for children types is
deprecated.
In Gatsby v3, only children fields explicitly set with the `childOf` extension will be added.
For example, in Gatsby v3, `ComponentInMdx` will not get a `childrenComponentInMdx` field.
```

nkuehn:
@jazevedo620 wow, this is amazing.  I think this is literally the most complete description and rationale I ever read in an open source PR. 

Your basic use case is a pretty interesting approach to documentation of APIs but also beyond.  I would summarize it as instead of having to originally author in a data-structured way you rather extract structural data from the free-form authoring done in MDX+JSX.  

We've decided (in the team maintaining this repo) that it would be great to have your rewrite in because some of its features could in fact become necessary for us, too in the future  and your description made clearer that it's not as good a fit into the actual gatsby mdx plugin as I had originally thought. We'd appreciate a lot if you help out maintaining the plugin if user issues and upgrade work comes up since we're not using it as intesely as you to. 

I'll do a sort of feature-level review in a minute, see it as suggestions, not as hard prerequisites to merge.   I'd like to find out to what degree the complexity and configurability can be reduced to ease later maintenance.

packages/gatsby-transformer-mdx-introspection/README.md
```diff
@@ -40,31 +47,122 @@ query MyQuery {
...
+        ...
+      ]
+    }
+  }
+}
+```
+
+## Options
+
+**`lowercaseIdentifiers`** [boolean] (optional, defaults to false)
```
nkuehn: I suggest to drop that legacy support.  This plugin is in a pre-1.0 alpha release phase and the lowercasing was never actually a wanted feature.

packages/gatsby-transformer-mdx-introspection/README.md
```diff
@@ -40,31 +47,122 @@ query MyQuery {
...
+
+Whether to have the `component` field (and `attribute.name`) be all lower case (legacy)
+
+**`trimWhitespace`** [boolean] (optional, defaults to true)
```
nkuehn: Did you have a use case for keeping the whitespace yourself? Otherwise i suggest to not have that configuration option, too.   The intent of the plugin is to make the meaning of the JSX explorable and not the syntax, so I don't think a use case that needs this fits the intent of the plugin. 

Every config option multiplies the testing and code complexity and users will bring up enough other configurability needs later on top.

packages/gatsby-transformer-mdx-introspection/README.md
```diff
@@ -40,31 +47,122 @@ query MyQuery {
...
+
+Whether to trim leading/trailing whitespace in JSX snippets and string literals
+
+**`collapseWhitespace`** [boolean] (optional, defaults to true)
```
nkuehn: same here?  Did you have own use cases for not collapsing the whitespace?

packages/gatsby-transformer-mdx-introspection/README.md
```diff
@@ -40,31 +47,122 @@ query MyQuery {
...
+
+Whether to collapse consecutive whitespace in JSX snippets and string literals
+
+**`attachAST`** [boolean] (optional, defaults to false)
```
nkuehn: Just to be sure: Your text sounded like you aded this option for a potential user.  Was there an own use case in your site implementation for digging into the full AST and doing custom things?  

Not being able to trigger a potentially bad performing call is the best way to not have trouble with implementations that use it too liberally, hence I tend to not have it.

packages/gatsby-transformer-mdx-introspection/README.md
```diff
@@ -40,31 +47,122 @@ query MyQuery {
...
+
+**`removeMdxCompilationArtifacts`** [boolean] (optional, defaults to true)
+
+Whether to remove attributes that are (usually) artifacts of MDX compilation (`mdxType` and `parentName`)
+
+**`shouldIndexNode`** [(node) ⇒ boolean] (optional, defaults to () ⇒ true)
+
+Predicate function used as a performance escape hatch to filter MDX files that get parsed/indexed. Use if not all MDX files need to be indexed and build times are prohibitive
+
+**`excludeTags`** [array\<string>] (optional, defaults to [...])
```
nkuehn: Since the defaults are inherently an assumption given specific sorts of markdown (which can contain arbitrary HTML) I wondered whether forcing an explicit whitelist wouldn't be better for the uses case we have in mind here (these all look for specific components that are hardwired in other parts of the implementation). 

I find whitelisting interesting because we have performance worries anyways and know use cases all look at a specific list of components or tags only.

packages/gatsby-transformer-mdx-introspection/README.md
jazevedo620: I'm fine with dropping support for this; as you said it was intended entirely as an option to preserve legacy behavior.

packages/gatsby-transformer-mdx-introspection/README.md
jazevedo620: The whitespace options are interesting, just because I intended to add them as an escape hatch for potential users to opt _out_ of the very simplistic algorithm I implemented. This was mostly to make it clear that the whitespace **is not** getting trimmed/collapsed according to the [standard algorithm that browsers uses](https://medium.com/@patrickbrosset/when-does-white-space-matter-in-html-b90e8a7cdd33), which takes into account HTML semantics and CSS properties. Instead, the behavior just collapses consecutive whitespace naiively, which may produce undesired effects if he contents need to be strictly parsed (which wasn't one of my specific requirements, but doesn't seem too unlikely).

I added this because the "text"/JSX contents **are** exposed via Gatsby's node model, so users would notice that the formatting might be different. However, I know that the MDX compiler already prettifies JSX itself, so source formatting may be lost anyways. With this in mind, I'm fine with removing the option, though I would also suggest combining both of the whitespace options to add an escape hatch out of the string processing (which would be simple enough for users to re-implement if they needed to tweak the way it works).

packages/gatsby-transformer-mdx-introspection/README.md
jazevedo620: I suggested combining these two, just because the default behavior might be undesired in certain circumstances. I don't have a use case myself for it, though.

packages/gatsby-transformer-mdx-introspection/README.md
jazevedo620: I haven't actually gotten to implementing the algorithm I need for my own use case, but the main reason I added this was that I wanted to allow potential users to query the underlying Babel JSX AST. However, now that I consider it, since there is the `tree` field exposed for manual traversal of the reduced tree, I'm not very attached to keeping this option. The best option is probably to remove it for now and introduce it (or a more performant solution) later on if it is needed.

packages/gatsby-transformer-mdx-introspection/README.md
jazevedo620: The whitelist approach is a good idea, and I think it serves most of the use cases I can think of, too. I would probably want to implement it using some `string | regular expression` syntax so that users could effectively implement their own blacklist with negative look-arounds.

jazevedo620:
I will work on implementing the changes discussed in the comments soon, though I am busy with school at the moment.

I'll try to get it done within a few days. Additionally, I will revert the change I made to the config of one of the docs sites' gatsby theme.

jazevedo620:
Okay, I have made most of the requested modifications @nkuehn; let me know what you think about the `cleanWhitespace` option. Also, like you suggested, I'm fine with helping maintain the plugin in the future since this PR incorporates so many changes.

Additionally, these changes will break the `gatsby-theme-api-docs` package, as it includes the plugin in its config and relies on the legacy behavior (and also doesn't specify a whitelist). I can try to fix these to work with the new plugin, but I'm not familiar with the codebase.

packages/gatsby-transformer-mdx-introspection/src/jsx-ast-utils.js
```diff
@@ -0,0 +1,253 @@
...
+function getChildren(node) {
+  let children = [];
+  if (node != null) {
+    // Examine each attribute of the object
+    Object.values(node).forEach(value => {
+      if (value != null && typeof value === 'object') {
+        if (Array.isArray(value)) {
+          // Check if it is a node array
+          if (value.every(isAstNode)) {
+            children = children.concat(value);
```
emmenko: What's the reason for the reassignment?

packages/gatsby-transformer-mdx-introspection/src/jsx-ast-utils.js
```diff
@@ -0,0 +1,253 @@
...
+}
+
+/**
+ * Cleans up a JSX snippet string, removing leading/trailing space and a
+ * trailing semicolon, if applicable
+ * @param {string} raw Raw JSX snippet string
+ */
+function cleanJsxSnippet(raw) {
+  const trimmed = raw.trim();
+  if (trimmed.charAt(trimmed.length - 1) === ';') return trimmed.slice(0, -1);
```
emmenko:
```suggestion
  return trimmed.replace(/(.*);$/, '$1')
```

packages/gatsby-transformer-mdx-introspection/src/jsx-ast-utils.js
```diff
@@ -0,0 +1,253 @@
...
+  return trimmed;
```
emmenko:
```suggestion
```

packages/gatsby-transformer-mdx-introspection/src/jsx-ast-utils.js
```diff
@@ -0,0 +1,253 @@
...
+}
+
+/**
+ * Finds all JSX element children of the current node
+ * @param {object} node Babel JSX AST
+ */
+function getJsxChildren(node) {
+  // Start searching at each child
+  return getChildren(node).reduce(
+    (accum, child) => [...accum, ...findNode(child, isJsxElement, false)],
```
emmenko:
```suggestion
    (children, child) => [...children, ...findNode(child, isJsxElement, false)],
```

packages/gatsby-transformer-mdx-introspection/src/reduce-jsx-ast.js
```diff
@@ -0,0 +1,307 @@
...
+ * nodes
+ * @param {object} context Context object
+ */
+function serializeAndSearch(node, collapse, detachedHeads, context) {
+  const { cleanWhitespace, jsx } = context;
+  const value = reduceNode(node, cleanWhitespace && collapse, jsx);
+  const subJsx = getJsxChildren(node);
+  if (subJsx.length > 0) {
+    // The value is (or contains) one or more JSX tree heads; add to list
+    detachedHeads.push(...subJsx);
```
emmenko: Not sure I like this side effect. I would prefer to pass a callback function instead and let the consumer do the array.push outside of this method.

```js
function serializeAndSearch(node, collapse, onJsxChildren, context) {
  // ...
  if (subJsx.length > 0) {
    onJsxChildren(subJsx);
  }
  return value;
}
```

packages/gatsby-transformer-mdx-introspection/src/reduce-jsx-ast.js
```diff
@@ -0,0 +1,307 @@
...
+ * Reduces component node children by removing whitespace-only text nodes
+ * and collapsing whitespace in all text nodes as neccessary (depending on
+ * options)
+ * @param {array} children Array of reduced component nodes
+ * @param {object} context Context object
+ */
+function reduceChildren(children, context) {
+  const { cleanWhitespace } = context;
+  const reduced = [];
+  children.forEach((child, i) => {
```
emmenko: Nitpick: use a `reduce` function?

packages/gatsby-transformer-mdx-introspection/src/transform-mdx.js
```diff
@@ -0,0 +1,35 @@
...
+  const jsx = await mdx(mdxBody);
+  const ast = babelParser.parse(jsx, {
+    sourceType: 'module',
+    plugins: ['jsx'],
+  });
+
+  // Find content root and reduce AST into component forest
+  const contentRoot = jsxAstUtils.findContentRoot(ast);
+
+  const reducedForest = reduceJsxAst(contentRoot, jsx, options);
```
emmenko:
```suggestion
  return reduceJsxAst(contentRoot, jsx, options);
```

packages/gatsby-transformer-mdx-introspection/src/transform-mdx.js
```diff
@@ -0,0 +1,35 @@
...
+  return reducedForest;
```
emmenko:
```suggestion
```

packages/gatsby-transformer-mdx-introspection/README.md
```diff
@@ -40,31 +45,108 @@ query MyQuery {
...
+
+**`removeMdxCompilationArtifacts`** [boolean] (optional, defaults to true)
+
+Whether to remove attributes that are (usually) artifacts of MDX compilation (`mdxType` and `parentName`)
+
+**`shouldIndexNode`** [(node) ⇒ boolean] (optional, defaults to () ⇒ true)
+
+Predicate function used as a performance escape hatch to filter MDX files that get parsed/indexed. Use if not all MDX files need to be indexed and build times are prohibitive
+
+**`tagWhitelist`** [array\<string | RegExp>] (optional, defaults to ['a', /^h[1-6]\$/)
```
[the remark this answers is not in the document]
jazevedo620: Fixed, along with moving JS array expression into code block

packages/gatsby-transformer-mdx-introspection/__tests__/integration.js
```diff
@@ -0,0 +1,379 @@
...
+const transformMdx = require('../src/transform-mdx');
+
+// Contains overall integration tests to test plugin functionality
+
+// Utility function to mock default options
+const mockOptions = () => ({
+  cleanWhitespace: true,
+  removeMdxCompilationArtifacts: true,
+  shouldIndexNode: () => true,
+  tagWhitelist: [/^(?!(?:MDXLayout)|(?:p)|(?:span)|(?:tr)|(?:th)|(?:td)).*$/],
```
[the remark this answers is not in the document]
jazevedo620: This regex was what the old default was when I had a blacklist-based method, so I just transferred it to here and forgot. I cleaned it up a bit in the latest commit, and explained how the negative lookarounds emulate a blacklist.

packages/gatsby-transformer-mdx-introspection/__tests__/integration.js
```diff
@@ -0,0 +1,379 @@
...
+
+${mdx1}
+`.trim();
+
+    const result = await introspectMdx(mdx, mockOptions());
+    expect(result.children).toMatchObject(tree1);
+  });
+
+  it('properly parses complex mdx documents to component trees', async () => {
+    const mdx = await fs.promises.readFile(
```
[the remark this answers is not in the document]
jazevedo620: No; I just made them async because the tests were already asynchronous; I'll replace them with synchronous methods

packages/gatsby-transformer-mdx-introspection/__tests__/jsx-ast-utils.js
```diff
@@ -0,0 +1,282 @@
...
+    expect(jsxAstUtils.findNode(parent, match('p'), false)[0]).toEqual(parent);
+  });
+});
+
+describe('isJsxElement', () => {
+  it('returns false for non-jsx elements', () => {
+    expect(jsxAstUtils.isJsxElement(mockText())).toBe(false);
+  });
+
+  it('returns true for non-jsx elements', () => {
```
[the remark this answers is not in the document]
jazevedo620: Good catch 👍

packages/gatsby-transformer-mdx-introspection/__tests__/jsx-ast-utils.js
```diff
@@ -0,0 +1,282 @@
...
+  });
+
+  it('directly converts string literals without collapsing ever', () => {
+    const text = 'example string  ';
+    const jsx = `"${text}"`;
+    // String literal AST node taken from example AST
+    const node = {
+      type: 'StringLiteral',
+      start: 1,
+      end: text.length + 2,
```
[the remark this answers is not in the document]
jazevedo620: I think there may have been a few bugs in the code, I cleaned them up and added some explanations for the magic numbers

packages/gatsby-transformer-mdx-introspection/gatsby-node.js
```diff
@@ -1,89 +1,355 @@
...
+       "Array of component nodes that are children of the current node"
+       tree: [JSON!]!
+       "Original MDX file parent node"
+       mdx: Mdx! @link
+       "Text content of the current node. To configure text processing rules, use collapse/trim"
+       content(collapse: Boolean, trim: Boolean): String!
+       "A single child JSX/HTML node (excluding text children). If deep is set, then the field could take the value of any child of children of this node at any level"
+       child${nodeName}: ${nodeName}
+       "All child JSX/HTML nodes (excluding text children). If deep is set, then the field will return all children of children of this node at any level"
+       children${nodeName}: [${nodeName}]
```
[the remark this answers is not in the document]
jazevedo620: Yes, I believe so, but I think I tried it before and it interacted strangely with Gatsby's auto-generated children node. I'll look into it

packages/gatsby-transformer-mdx-introspection/package.json
```diff
@@ -16,8 +16,9 @@
   ],
   "scripts": {},
   "dependencies": {
-    "parse5": "5.1.1",
-    "remark": "11.0.2",
-    "remark-mdx": "1.5.1"
+    "@babel/parser": "^7.7.7",
```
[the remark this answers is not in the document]
jazevedo620: I replaced them in the `package.json`, but I'm not sure if I need to regenerate the lockfile.

packages/gatsby-transformer-mdx-introspection/gatsby-node.js
```diff
@@ -1,89 +1,355 @@
...
+    tagWhitelist,
+  } = withDefaults;
+  const appliedOptions = {
+    removeMdxCompilationArtifacts,
+    cleanWhitespace,
+    tagWhitelist,
+  };
+
+  const cacheKey = reducedCacheKey(node, appliedOptions);
+  const cachedForest = await cache.get(cacheKey);
```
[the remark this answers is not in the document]
jazevedo620: The "forest" name is an extension of the association between tree data structures and their natural counterparts-- a forest data structure is a collection of independent trees, generally represented by a collection of their heads. I'm not sure if this is common terminology; the only place I've heard it used is in my algorithms class at uni. If you think there's a better term, let me know.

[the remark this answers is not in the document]
jazevedo620: I'll add a comment clarifying this term in the code for now

packages/gatsby-transformer-mdx-introspection/src/jsx-ast-utils.js
jazevedo620: The reassignment is used to merge the two arrays (children and value, which is resolved to be an node array in this path) using concat. The reason it's a reassignment and not a `value.forEach(v => children.push(v))` or a `push(...value)` is just to save on the iteration/spread cost and directly merge, but if you think it shouldn't use reassignment I don't mind changing it.

packages/gatsby-transformer-mdx-introspection/src/reduce-jsx-ast.js
jazevedo620: I feel like the mutable array parameter isn't great as well-- it never felt like a good solution. I adapted it to use a callback to remove the mutable parameter.

packages/gatsby-transformer-mdx-introspection/gatsby-node.js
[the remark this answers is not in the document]
jazevedo620: I just tested it out and it seems like it does replace the non-null assertions with it's own return type. On the query info screen in GraphiQL, it shows this (also without the description):
example
I did add the assertions in, but they don't affect the final schema because of Gatsby's behavior. I remember wrestling with this for a while, but I'm not sure how to disable it and remove all warnings. Either way, I think the current state is sufficient, as not having the assertions/description are the only things that are wrong with the way it is currently.

jazevedo620:
Thanks for taking the time to review the changes @emmenko. I made all of the suggested changes, and left a few unresolved discussions open for you to take a look at when you get the chance.

Let me know if there's anything else I should change before the PR gets merged. Thanks

packages/gatsby-transformer-mdx-introspection/package.json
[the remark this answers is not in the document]
emmenko: Yes just run `yarn` again. Thanks

packages/gatsby-transformer-mdx-introspection/__tests__/integration.js
```diff
@@ -5,11 +5,15 @@ const transformMdx = require('../src/transform-mdx');
...
-const mockOptions = () => ({
+const createTestPluginOptions = () => ({
   cleanWhitespace: true,
   removeMdxCompilationArtifacts: true,
   shouldIndexNode: () => true,
-  tagWhitelist: [/^(?!(?:MDXLayout)|(?:p)|(?:span)|(?:tr)|(?:th)|(?:td)).*$/],
+  // This whitelist is an example of a blacklist that can be used for
+  // thorough parsing of a component tree. `p` and `span` are excluded
+  // due to being generally insignificant, and `MDXLayout` is excluded
+  // because it is always the root of the primary component tree
```
emmenko: Thanks for the explanation

packages/gatsby-transformer-mdx-introspection/__tests__/jsx-ast-utils.js
```diff
@@ -229,19 +234,24 @@ describe('reduceNode', () => {
   });
 
   it('directly converts template strings from JSX without delimiters', () => {
+    const tag = 'span';
     const text = 'template string';
-    const jsx = `<span>{\`${text}\`}</span>`;
+    const jsx = `<${tag}>{\`${text}\`}</${tag}>`;
     // Template literal AST node taken from example AST
+    // start of template literal (+3 for <, >, and {)
```
emmenko: 👍

packages/gatsby-transformer-mdx-introspection/gatsby-node.js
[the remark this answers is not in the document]
emmenko: Got it, thanks. I just never heard this term before.

packages/gatsby-transformer-mdx-introspection/src/reduce-jsx-ast.js
```diff
@@ -76,13 +76,14 @@ function transformSpreadAttribute(attribute, context) {
   // Keep track of any JSX elements seen outside of direct children to parse
   // and index (but not tag as children of the current element)
   const detachedHeads = [];
+  const handleDetachedHeads = heads => detachedHeads.push(...heads);
```
emmenko: 💯

jazevedo620:
Alright, I rebased the branch and fixed a couple linting errors. The latest version's functionality still works in my own dogfooding environment, so it should be ready to merge.

packages/gatsby-transformer-mdx-introspection/README.md
```diff
@@ -40,31 +45,108 @@ query MyQuery {
...
+**`tagWhitelist`** [array\<string | RegExp>] (optional, defaults to `['a', /^h[1-6]\$/]`)
+
+JSX components that generate Gatsby data nodes in the final output (other nodes still appear as children in `tree` and their children can generate nodes)
+
+##### Default value
+
+By default, the plugin only generates Gatsby data nodes for the following elements:
+
+- `a`
+- `h1`
```
nkuehn: I wonder whether the headings are a good default - the heading structure is already available from the MDX plugin GraphQL and if you can use that you should:

```graphql
query MdxHeadings {
  allMdx {
    nodes {
      headings {
        depth
        value
      }
    }
  }
}
````

If the idea behind these defaults is to point out that by using this plugin instead of the mdx plugin's built-in you can cover more or different cases we should explain that difference to point out advantages.

File: packages/gatsby-transformer-mdx-introspection/README.md [hunk 26]
```diff
@@ -1,37 +1,42 @@
 # gatsby-transformer-mdx-introspection
 
-This plugin statically analyzes the nodes generated by `gatsby-plugin-mdx`
-and generates GraphQL Nodes that allow to find out what MDX-sourced pages
-contain which manually placed React components.
+This plugin statically analyzes the nodes generated by `gatsby-plugin-mdx` and generates GraphQL Nodes that allow users to find out what MDX-sourced pages contain which manually placed React components or automatically placed HTML elements.
```
nkuehn:
```suggestion
This plugin statically analyzes the nodes generated by `gatsby-plugin-mdx` and generates GraphQL nodes that allow users to find out what MDX-sourced pages contain which manually placed React components or automatically placed HTML elements.
```

Plus: what's an "automatically placed HTML element"?  If you mean the HTML generated from markdown syntax we could word that more explicitly.

packages/gatsby-transformer-mdx-introspection/README.md
```diff
@@ -40,31 +45,108 @@ query MyQuery {
...
+
+Whether to collapse/trim whitespace in JSX snippets and string literals
+
+**`removeMdxCompilationArtifacts`** [boolean] (optional, defaults to true)
+
+Whether to remove attributes that are (usually) artifacts of MDX compilation (`mdxType` and `parentName`)
+
+**`shouldIndexNode`** [(node) ⇒ boolean] (optional, defaults to () ⇒ true)
+
+Predicate function used as a performance escape hatch to filter MDX files that get parsed/indexed. Use if not all MDX files need to be indexed and build times are prohibitive
```
nkuehn:
```suggestion
Predicate function used as a performance escape hatch to filter MDX files that get parsed/indexed. Use if not all MDX files need to be indexed.
```

packages/gatsby-transformer-mdx-introspection/README.md
jazevedo620: You're correct; I'll make that more clear in the README.

packages/gatsby-transformer-mdx-introspection/README.md
nkuehn: @jazevedo620  Do you have an opinion here?

jazevedo620: I was thinking about and I didn't have a strong attachment to the current defaults of `a` and `h1-6`, they just seemed like a useful starting point to introspect in documentation. I had forgot about the normal MDX plugin's header querying functionality, but the default value isn't designed to really be used that much.

In that case, it might be better to have the whitelist be a required option for the plugin to work, and it could warn to console/panic on build if it doesn't detect that it has been set.

emmenko:
@nkuehn anything else from your side?

emmenko:
FYI: I changed the target branch to `v2` instead of `master`, for a couple of reasons:

- this PR contains some breaking changes, so we can keep them in `v2` until we decide to release it (in case there might be additional breaking changes, so we can batch them). This way, `master` stays "clean" in case we have some urgent bug fixes.
- the VRT tests don't on a PR from a fork (because github secrets are not available there for security reasons), so when we merge it to `v2` they should run and we see if everything is fine.

packages/gatsby-transformer-mdx-introspection/README.md
jazevedo620: @nkuehn Do you have an opinion on whether the default should be changed to an empty whitelist or whether it should kept as is?

nkuehn: I suggest to document that it's a required parameter and leave the default empty.   We'll have happier users due to the performance gain. 

README should explain how to put a wildcard to first introspect everything and then put what you were looking for.

jazevedo620:
Sorry for the delay; I have been caught up in schoolwork.

@nkuehn I changed it so that `tagWhitelist` is a required option, and the plugin doesn't work without it. I pushed my changes, but they seem to be failing CI for (seemingly) unrelated reasons.

emmenko:
Alright, merging this into the `v2` branch. Further improvements can be done from that branch.

@jazevedo620 thanks again for your big contribution! 🙌

PR #206, opened 2020-01-13
Parameters design implementation

davifantasia:
This covers the following

- Table component for parameters and types
- Restructuring of Method component with `designSystem`, `Spacings`, etc. from ui-kit.


packages/gatsby-theme-api-docs/src/design-system.js
```diff
@@ -0,0 +1,39 @@
+import { designSystem } from '@commercetools-docs/ui-kit';
+
+export const tokens = {
```
emmenko: These tokens should be picked directly from the original uikit tokens, we don't need to re-implement them again.

packages/gatsby-theme-api-docs/src/design-system.js
```diff
@@ -0,0 +1,39 @@
...
+export const dimensions = {
+  widths: {
+    tableColumn: designSystem.pxToRem('200px'),
+    methodBorderLeft: designSystem.pxToRem('8px'),
+  },
+};
+
+export const colors = {
+  light: {
+    surfaceTableHead: '#f2f2f2',
```
emmenko: Are all those colors "new"? Or are they defined in the uikit design system?

PR #209, opened 2020-01-16
chore: try CI workflow with github actions

emmenko:
Read motivation here: https://github.com/commercetools/commercetools-docs-kit/pull/209#issuecomment-576024614

TL;DR: remove CircleCI and use GitHub Actions to run the CI workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -189,17 +189,17 @@ workflows:
   build_and_test:
     jobs:
       - build_lint_and_test
-      - vrt_test:
-          requires:
-            - build_lint_and_test
-      - e2e_test:
-          name: e2e_smoke_tests
-          requires:
-            - build_lint_and_test
-          start: yarn run serve -l 8000 -n public
-          wait-on: 'http://localhost:8000'
-          run-command: PERCY_TOKEN="$PERCY_TOKEN_SMOKE_TESTS_E2E" yarn run percy exec -- yarn test:e2e --env CI=true
-          store_artifacts: true
+      # - vrt_test:
+      #     requires:
+      #       - build_lint_and_test
+      # - e2e_test:
+      #     name: e2e_smoke_tests
+      #     requires:
+      #       - build_lint_and_test
+      #     start: yarn run serve -l 8000 -n public
+      #     wait-on: 'http://localhost:8000'
+      #     run-command: PERCY_TOKEN="$PERCY_TOKEN_SMOKE_TESTS_E2E" yarn run percy exec -- yarn test:e2e --env CI=true
+      #     store_artifacts: true
       - publish:
           requires:
             - vrt_test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,12 +200,12 @@ workflows:
       #     wait-on: 'http://localhost:8000'
       #     run-command: PERCY_TOKEN="$PERCY_TOKEN_SMOKE_TESTS_E2E" yarn run percy exec -- yarn test:e2e --env CI=true
       #     store_artifacts: true
-      - publish:
-          requires:
-            - vrt_test
-            - e2e_smoke_tests
-          filters:
-            branches:
-              only:
-                - master
+      # - publish:
+      #     requires:
+      #       - vrt_test
+      #       - e2e_smoke_tests
+      #     filters:
+      #       branches:
+      #         only:
+      #           - master
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -184,11 +184,17 @@ jobs:
       - setup_dependencies
       - publish_to_canary
 
+  noop:
+    executor: node_12
+    steps:
+      - run: echo 'Skipping build'
+
 workflows:
   version: 2
   build_and_test:
     jobs:
-      - build_lint_and_test
+      - noop
+      # - build_lint_and_test
       # - vrt_test:
       #     requires:
       #       - build_lint_and_test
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: actions/upload-artifact@v1
+        if: failure()
+        with:
+          name: cypress-videos
+          path: cypress/videos
+
+  publish_to_canary:
+    needs: build_lint_and_test
+
+    if: github.ref == 'refs/heads/master'
```
tdeekens: This stuff is so nice. Note, you may be aware of. We could add another workflow called prerelease. Then this workflow could take:

```yml
on:
  push:
    branches:
      - 'commercetools/master'
```

The downside being that we'd have to repeat the install steps. Cause I am not aware of cross workflow deps.

.github/workflows/main.yml
```diff
@@ -0,0 +1,68 @@
...
+jobs:
+  build_lint_and_test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
```
tdeekens: I used the native actions/checkout and actions/cache which also allows for pretty dope caching while actions/setup-node comes with yarn.

.github/workflows/main.yml
emmenko: That's what the `needs: build_lint_and_test` is for, you can add dependencies between jobs. Workflow deps do not exist yet.

.github/workflows/main.yml
emmenko: This action does the same thing as far as I know. Is there any particular difference that this action is missing?

tdeekens: Ah if it caches the install then awesome. Didn't know it did.

emmenko: Yeah https://github.com/bahmutov/npm-install/blob/master/index.js

.github/workflows/main.yml
```diff
@@ -0,0 +1,85 @@
...
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN_SMOKE_TESTS }}
+
+      - name: Running End-to-End tests
+        uses: cypress-io/github-action@v1.16.1
+        with:
+          headless: true
+          start: 'yarn run serve -l 8000 -n public'
+          wait-on: 'http://localhost:8000/docs-smoke-test'
+          env: 'CI=true'
+          command-prefix: 'yarn run percy exec --'
```
emmenko: FYI: this option does not exist yet. I opened a PR to add support for it: https://github.com/cypress-io/github-action/pull/61

Without that, we can't run Percy with Cypress.

On the other hand, the GitHub Action of Percy does not support waiting for a server to run, therefore we can't use that either. https://github.com/marketplace/actions/percy

emmenko: YAY! https://github.com/cypress-io/github-action/releases/tag/v1.20.0

nkuehn:
@emmenko can you shed some light on the motivation of switching (back?) to GH actions?

emmenko:
@nkuehn yes, I didn’t officially mentioned anything yet cause I wanted to try things out first.

The main reason is that we want to offload some of the jobs from circleci to reduce the amount of the resources that we use there, especially once we switch to the new plan.
Since we can now use github actions with our github org account, we wanted to use those for some of our OSS repositories. I wanted to start with the appkit, but since the docs kit has a very similar setup and the CI is simpler, I decided to try things out here.

Hope this clarifies things a bit, let me know if you have more questions. I’ll ping you again once this is ready for review.

.github/workflows/main.yml
```diff
@@ -0,0 +1,85 @@
+name: Main workflow
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
```
emmenko: @tdeekens I'm still unsure what to use exactly as trigger events.
I understand that `push` is triggered by each commit and `pull_requests` is triggered by commits pushed to an open PR.
So if a PR is open, the workflow is run twice.
If we only use `pull_request`, the workflow is run only when you open a PR, so there might be cases where the workflow is skipped.
The `push` event is still important on `master` though.

How would you set this up so that it "works as expected" and it does not run twice?

tdeekens: @emmenko yeah I am also still wrapping my head around those events. Generally, I would not want CI to run when I push to a branch is does not have a respectively open PR. Not necessarily as I might just want to "stash" my work upstream. Moreover, it cuts into the allowance or minutes on GitHub Actions often without having too I think.

That said, I would use `pull_request` on the general testing, building and linting workflow. As that then doesn't run on master I would add another workflow (or configure the same with two events) to have `push` but only run on `commercetools/commercetools-docs-kit/master`. At least that's how I got it closest to what I felt was good.

emmenko: Yeah ok.

The way it's configured now it's as you mentioned: triggered on `pull_request` events and `push` on `master` only.
So it should be good like this and does it meet your expectations?

tdeekens: Yeah, it matches what I'd expect. I wouldn't expect CI to run without a PR being opened. But when I open only on the last commit. That may vary from person to person but still feels like a decent default.

.circleci/config.yml
```diff
@@ -186,23 +186,23 @@ workflows:
   build_and_test:
```
emmenko: I'll clean up circleci once we merge this, to avoid failing builds on other PRs.

nkuehn:
Thanks for the context @emmenko !  I generally like the idea to keep the number of involved cloud services low in an open source repository - it makes the build and test output more transparent, too to less involved contributors.

emmenko:
Alright, given that we're all more or less ok with this approach, I'll merge this in within the next days and remove circleci.

PR #210, opened 2020-01-16
Method design implementation

davifantasia:
This the implementation of the [design](https://commercetools.invisionapp.com/d/main?origin=v7#/console/17784193/388243826/preview?scrollOffset=946) for a resource method.


packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
```diff
@@ -65,89 +65,84 @@ const Method = ({ apiKey, url, resourceUriParameters, method, methodType }) => {
...
+      <Description>{method.description}</Description>
+
+      <SpacingsStack scale="m">
+        <UrlScopesResponseContainer
+          css={css`
+            border-left-color: ${methodColor};
+          `}
+        >
+          <UrlScopesResponses
+            data={{
```
emmenko: Any particular reason for passing the data as a single prop instead of multiple props for each data property?

packages/gatsby-theme-api-docs/src/components/api/resource/method/url-scopes-responses/url-scopes-responses.js
```diff
@@ -1,21 +1,22 @@
...
 import styled from '@emotion/styled';
+import { customProperties } from '@commercetools-uikit/design-system';
 
 import { responseRepresentation } from '../../../../../utils/constants';
 import Scopes from './scopes';
 import Responses from './responses';
 
 const Container = styled.div`
-  padding: 1rem;
+  padding: ${customProperties.spacingM};
```
emmenko: We should use here the spacing from the docs uikit. 

```
designSystem.dimensions.spacings.m
```

packages/gatsby-theme-api-docs/src/components/api/resource/method/url-scopes-responses/scopes.js
```diff
@@ -1,22 +1,21 @@
...
-  color: #999;
+import Title from './title';
+
+const ListContainer = styled.p`
+  margin-top: ${customProperties.spacingXs};
 `;
 
 const Scopes = ({ scopes, title }) => {
   return (
     <div>
```
emmenko: You can use here SpacingsStack instead of defining a margin-top to the ListContainer

packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
davifantasia: There was a reason, I don't really remember exactly why at the moment.

packages/gatsby-theme-api-docs/src/components/api/resource/method/url-scopes-responses/url-scopes-responses.js
davifantasia: I see, thanks.

packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
emmenko: This question is still valid. I would spread the `data` into single props, and be explicit which props are passed to the component

```js
<UrlScopesResponses
  apiKey={apiKey}
  uris={uris}
  scopes={{ }}
  responses={method.responses}
/>
```

packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
```diff
@@ -65,95 +71,91 @@ const Method = ({ apiKey, url, resourceUriParameters, method, methodType }) => {
...
+          />
+        ) : null}
       </SpacingsStack>
-    </div>
+    </SpacingsStack>
   );
 };
 
-function computeMethodNameBackgroundColor(methodName) {
+function computeMethodColor(methodName) {
```
emmenko: Nitpick: rename to `getColorByMethod`

packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
```diff
@@ -65,95 +71,91 @@ const Method = ({ apiKey, url, resourceUriParameters, method, methodType }) => {
...
-        background-color: ${colors.light.methods.get};
-      `;
+      return colors.light.methods.get;
   }
 }
 
 Method.propTypes = {
   apiKey: PropTypes.string.isRequired,
-  url: PropTypes.string.isRequired,
+  uris: PropTypes.object.isRequired,
```
emmenko: Please always try to define the shape of the objects (`PropTypes.shape`). It also helps eslint to catch errors

packages/gatsby-theme-api-docs/src/components/api/resource/method/url-scopes-responses/responses.js
```diff
@@ -2,67 +2,53 @@ import React from 'react';
...
-            );
-          })}
-        </ResponseCodeType>
-      </Body>
+      {title ? <Title>{title}</Title> : null}
+
+      {responses.map(response => {
+        return (
+          <p key={response.code}>
+            <ResposeCode css={computeStatusCodeBackgroundColor(response.code)}>
```
emmenko: Nitpick: rename to `getStatusCodeStyles`

packages/gatsby-theme-api-docs/src/components/api/resource/method/url-scopes-responses/scopes.js
```diff
@@ -1,33 +1,24 @@
...
+      {scopes.map((scope, index) =>
+        index === scopes.length - 1 ? (
+          <Markdown.InlineCode key={scope}>{scope}</Markdown.InlineCode>
+        ) : (
+          <span key={scope}>
+            <Markdown.InlineCode>{scope}</Markdown.InlineCode>
+            {', '}
+          </span>
+        )
+      )}
```
emmenko: I think this can be simplified as:

```js
{scopes.map((scope, index) => (
  <>
    <Markdown.InlineCode>{scope}</Markdown.InlineCode>
    {index < scopes.length - 1 ? <span>{', '}</span> : null}
  </>
))}
```

packages/gatsby-theme-api-docs/src/components/api/resource/method/url-scopes-responses/title.js
```diff
@@ -0,0 +1,8 @@
+import styled from '@emotion/styled';
+import { customProperties } from '@commercetools-uikit/design-system';
+import { designSystem } from '@commercetools-docs/ui-kit';
+
+export default styled.div`
+  color: ${customProperties.colorNeutral60};
```
emmenko:
```suggestion
  color: ${designSystem.colors.light.textFaded};
```

packages/gatsby-theme-api-docs/src/design-system.js
```diff
@@ -35,5 +41,7 @@ export const typography = {
     th: designSystem.pxToRem('21px'),
```
emmenko: This looks like a magic number. Please use a "normal" value, like `20` or `24`

packages/gatsby-theme-api-docs/src/design-system.js
```diff
@@ -35,5 +41,7 @@ export const typography = {
...
     methodType: designSystem.pxToRem('26px'),
     methodTitle: designSystem.pxToRem('32px'),
+    responseCode: designSystem.pxToRem('16px'),
+    responseBodyType: designSystem.pxToRem('19px'),
```
emmenko: Same here, this looks like a magic number. Please change

websites/api-docs-smoke-test/src/generated-api-specs/test/api.raml
```diff
@@ -0,0 +1,111 @@
+#%RAML 1.0
```
emmenko: Are those changes still necessary? I thought you fixed that already...

packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
davifantasia: I guess the question from me is, what is the problem?

packages/gatsby-theme-api-docs/src/design-system.js
davifantasia: I got it from the design.

packages/gatsby-theme-api-docs/src/design-system.js
davifantasia: I got it from the design.

websites/api-docs-smoke-test/src/generated-api-specs/test/api.raml
davifantasia: It must be the merge with the master, as I did not add these files in this branch.

packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
davifantasia: Could you please leave out nitpicks when you review?

I don't think it's good investing time in a discussion about something that's subjective.

packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
emmenko: In general, it's always better to pass scalar values as props, which makes the component easier to optimize for "performance".

Why I also asked was that you are constructing the object to be passed as a single prop from values that are already separated, while then destructuring it again in the component itself. It does not make much sense to me.

packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
emmenko: Sorry but nitpicks are still part of a review, and they are not necessarily subjective.

In this case I'm providing a suggestion to have the name following conventions in other repos.

> I don't think it's good investing time in a discussion

Yes it's good to invest time to make the code more readable, and no it's not meant to be a discussion because it's a nitpick and it's up to you to take the suggestion or not.

packages/gatsby-theme-api-docs/src/design-system.js
emmenko: I know, but we as developers should not just take everything for granted and bring such things to the attention of the designer that gave the number to you. Same goes for colors, etc.
You need to keep an eye out for this things and when you see such "magic numbers" it usually means that something does not feel right.

Please check back with Sven about this, thanks.

websites/api-docs-smoke-test/src/generated-api-specs/test/api.raml
emmenko: Can you remove those things then? If they are here it means that something went wrong with updating your branch.

emmenko:
Also, don't forget to label the PR 😉

packages/gatsby-theme-api-docs/src/components/api/resource/method/method.js
davifantasia: I will change it but your reason doesn't really make sense to me.

emmenko: You don't need to change it because I said so, I wanted to understand what was the reason for doing that as it looked "weird" to me doing it that way.
I then explain how and why I would do it.

davifantasia:
Closed and implemented here:

https://github.com/commercetools/commercetools-docs-kit/pull/222

PR #218, opened 2020-01-20
Missing raml fix [WIP]

davifantasia:
This fixes missing api specs in the UI:

https://commercetools-docs-git-renovate-all.commercetools.now.sh/import-api/import-operation


emmenko:
Thanks, so we can then release this and things should work again?

emmenko:
Don't forget to assign the labels to PRs

davifantasia:
> Thanks, so we can then release this and things should work again?

It should work but still needs to be tested as I can't yet find a way to test locally until it's available on NPM.

websites/import-api/README.md
```diff
@@ -0,0 +1,16 @@
+# Import API
```
emmenko: Don't commit this here please!

PR #220, opened 2020-01-21
Missing raml fix

davifantasia:
Left out gatsby-transformer-raml version which also resulted is missing specs for import api, this should fix it.


packages/gatsby-theme-api-docs/package.json
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "@commercetools-docs/gatsby-theme-docs": "1.6.0",
     "@commercetools-docs/gatsby-transformer-mdx-introspection": "1.4.0",
-    "@commercetools-docs/gatsby-transformer-raml": "",
+    "@commercetools-docs/gatsby-transformer-raml": "1.6.1-canary.1",
```
emmenko: This is wrong, you just need to put the version defined in the `@commercetools-docs/gatsby-transformer-raml` package.json

emmenko: https://github.com/commercetools/commercetools-docs-kit/blob/65ae5ecb2bab1aeee353fdc0ebdee86bd2e1defa/packages/gatsby-transformer-raml/package.json#L2-L3

davifantasia: Even if I need the latest?

davifantasia: Because I want the version that has the changes I made yesterday.

davifantasia: Version 1.5.0 is 5 days old.

emmenko: It does not matter, what matters is that it needs to reference the version in the monorepository.

When lerna releases, it will bump both versions.

davifantasia: So when will Lerna release?

I ask so I can know when I can test in import-api.

emmenko: Usually, when you squash merge a PR, the merged commit will be released by lerna, depending on what changed.

So if you made changes to both packages, both will be released with the correct canary version.

However, in this case it didn't get released before because the version here was missing.

You can maybe change something in the raml package, like something in the readme. Then when you merge this PR, it will bump both packages (raml + api docs) with the canary release

packages/gatsby-transformer-raml/README.md
```diff
@@ -75,6 +75,8 @@ Example for reading all types:
       enumeration
       examples {
         name
+        displayName
+        description
```
davifantasia: @emmenko I guess a change like this is sufficient?

emmenko: Yeah it should work.

PR #223, opened 2020-01-22
feat(theme): implement top menu for global links

### LINKED ISSUE #123: Top Menu
- implemented in the kit (part of the theme)
 - pulls its content / links from a central URL   e.g.  https://docs.commercetools.com/menu.json
 - that URL is provided by  a new "home" website in the commercetools-docs repo. A netlify redirect places it into the root.

emmenko:
Closes #123

### DIFF
--- a/packages/gatsby-theme-docs/src/data/top-menu.yaml
+++ b/packages/gatsby-theme-docs/src/data/top-menu.yaml
@@ -1,0 +1,28 @@
+- menu-title: Developer Center
+  items:
+    - label: HTTP API
+      href: https://docs.commercetools.com/http-api
+    - label: GraphQL API
+      href: https://docs.commercetools.com/graphql-api
+      beta: true
+    - label: Platform Release Notes
+      href: https://docs.commercetools.com/release-notes
+    - label: SDKs and Tools
+      href: https://docs.commercetools.com/software-development-kits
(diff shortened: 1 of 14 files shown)


packages/gatsby-theme-docs/src/data/top-menu.yaml
```diff
@@ -0,0 +1,28 @@
...
+      href: https://docs.commercetools.com/http-api
+    - label: GraphQL API
+      href: https://docs.commercetools.com/graphql-api
+      beta: true
+    - label: Platform Release Notes
+      href: https://docs.commercetools.com/release-notes
+    - label: SDKs and Tools
+      href: https://docs.commercetools.com/software-development-kits
+    # - label: Import and Export Tools
+    #   href: https://docs.commercetools.com/import-api
```
emmenko: I kept this disabled for now, same for custom apps below

emmenko:
> I think we should make the meta-links (sign in / sign up / etc.) full https:// URLs until we're live and fully migrated to ease the link checking during migration.

All links defined there are absolute.

> I observed the menu cannot be opened until the statically rendered page is fully hydrated. I first thought it does not work at all. I can't really see how to fix that except not forgetting to keep the bundle size and load performance under closer observation.

I think it's because the menu is loaded only when you open it.

PR #227, opened 2020-01-23
refactor(theme): footer to contain global links as well

emmenko:
~~PS: I'll see if I can extract/reuse some common components from footer and top menu.~~


emmenko:
Verified and approved by @Shecki

PR #234, opened 2020-01-24
chore: ignore .vscode/settings.json, add recommended.code-workspace

emmenko:


davifantasia:
Whats the difference between `recommended.code-workspace` and `.editorconfig`?

emmenko:
They are 2 different things. The `recommended.code-workspace` is specifically for VSCode, the `.editorconfig` is more of a general config that many editors support.

emmenko:
The workspace config is also useful to have some of the editor setup "documented", since we now don't include the `settings.json` in git anymore.

PR #244, opened 2020-01-28
fix(writing-style): make custom apps capitalized

emmenko:
As suggested by @adinakleine 

> Custom Application is an established ct term due to this it has to be capitalized.


emmenko:
Please discuss between the two of you and reach a consensus of what form we should use 🙂

adinakleine:
Andrea also votes for **Custom Applications** as set ct term.

nkuehn:
@adinakleine  please provide your feedback as a review with comment, not just as a comment to allow this to be merged if you approve. Thanks !

PR #247, opened 2020-01-29
fix(writing-style): mapping of custom apps casing

emmenko:
Leftovers of #244


packages/writing-style/vale-styles/commercetools/Brand.yml
```diff
@@ -15,17 +15,3 @@ swap:
   SPHERE: the commercetools platform
   SPHERE.IO: the commercetools platform
   sphere.io: the commercetools platform
-  MC: Merchant Center
```
emmenko: All of these were already included in the `Product.yml` file.

Issue #221, opened 2020-01-21
Production build should fail if a Type or Resource is missing or not found


davifantasia:
- Look into error logging using Gatsby tools
- it's better if all errors are collected so build will only fail after that.
- consider e2e test

PR #250, opened 2020-01-29
feat: Production build should fail if a Type or Resource is missing or not found 

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/221

For missing type or resource
- Dev build renders friendly UI prompt
- Production build fails

### DIFF
--- a/packages/gatsby-theme-api-docs/src/components/type/type.js
+++ b/packages/gatsby-theme-api-docs/src/components/type/type.js
@@ -1,9 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { ContentNotifications } from '@commercetools-docs/ui-kit';
 import filterOutApiTypeSubtypes from '../../utils/filter-out-api-subtypes';
 import { generateTypeURN } from '../../utils/ctp-urn';
 import { useApiTypes } from '../../hooks/use-api-types';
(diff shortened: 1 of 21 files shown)


emmenko:
Ah we need to ignore the changelog from the writing style linter. Will open a separate PR.

packages/gatsby-theme-api-docs/src/components/type/type.js
```diff
@@ -47,6 +45,16 @@ const ApiType = props => {
   );
 };
 
+function doIfMissingType(type, apiKey) {
+  const errorMsg = `Type with name '${type}' not found in '${apiKey}' API`;
+
+  if (__DEVELOPMENT__) {
```
emmenko: You should just use `process.env.NODE_ENV !== 'production'`

packages/gatsby-theme-api-docs/src/utils/do-if-missing-in-api.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { ContentNotifications } from '@commercetools-docs/ui-kit';
+
+function doIfMissingInApi(errorMsg) {
```
emmenko: Maybe rename it simply `reportError`?

PR #253, opened 2020-01-29
refactor(theme): to make the link underline style configurable via a prop

emmenko:
This helps to avoid having to override styles directly, but instead changing the appearance via a prop.


packages/gatsby-theme-docs/src/components/beta-flag.js
```diff
@@ -39,7 +38,7 @@ const getStyles = props => {
 const BetaFlag = props => {
   if (props.href) {
     return (
-      <Link href={props.href} css={getStyles(props)}>
+      <Link href={props.href} noUnderline={true} css={getStyles(props)}>
```
emmenko: FYI: the prop is called `noUnderline` on purpose, because the default behavior should be with underline and you can decide to explicitly opt-out.

PR #256, opened 2020-01-30
Import API pre-beta launch

davifantasia:
This includes fixes necessary before the Beta launch.

Import API is copied here just for testing on an actual project. Will remove when changes are completed.

Related issue:
https://github.com/commercetools/commercetools-docs/issues/214


nkuehn:
@davifantasia  thanks!  I'll try to look over it later today.  

Can you please update the title and description - this is not launching the Import API as public beta, this is migrating it to the new RAML based API documentation implementation. 

what's being launched next week is Custom Applications,  Import API will separately move from private to public beta.    Doing that would mean to change the search visibility settings and adding it to the menu.   That's not supposed to happen yet here. 

Thanks !

PR #257, opened 2020-01-30
fix: import api bugs

davifantasia:
This PR addresses the following

- "PUT" methods was reported as missing, it was actually not parsed by gatsby-transformer-raml
- Type link not rendered appropriately: this affects property type. E.g., [ProcessingState](https://commercetools-docs.now.sh/import-api/import-operation#ctp:import:type:), in this case the url is not complete it should be https://commercetools-docs.now.sh/import-api/import-operation#ctp:import:type:ProcessingState
- Types defined on website not rendered at all as a link: types that are defined on a page on the website should be linked. This affect various property types and response representations on the Import API site. (EDIT - will create an [issue](https://github.com/commercetools/commercetools-docs-kit/issues/263) and will be fixed in a different PR)


packages/gatsby-theme-api-docs/src/components/resource/method/method.js
```diff
@@ -138,6 +138,7 @@ const Method = ({
 function computeMethodColor(methodName) {
   switch (methodName) {
     case 'post':
+    case 'put':
```
emmenko: Maybe we can map each case explicitly and throw an error in the `default` case.

packages/gatsby-theme-api-docs/src/hooks/use-api-resources.js
```diff
@@ -49,6 +49,38 @@ export default () => {
...
+                code
+                description
+                body {
+                  applicationjson {
+                    type
+                    builtinType
+                  }
+                }
+              }
+            }
```
emmenko: Can we use fragments to avoid repeating all the query fields?

packages/gatsby-theme-api-docs/src/components/resource/method/method.js
davifantasia: I don't get what you mean by mapping each case explicitly.

emmenko: Something like this:

```js
switch (methodName) {
  case 'post':
  // ...
  default:
    throw new Error(`Unknown method ${methodName}`)
}
```

davifantasia: Also, I'm not sure why an error needs to be thrown here, if the method does not exist this component would not render. Can you please explain your reasoning?

packages/gatsby-theme-api-docs/src/hooks/use-api-resources.js
```diff
@@ -3,7 +3,40 @@ import { useStaticQuery, graphql } from 'gatsby';
 export default () => {
   const queryResult = useStaticQuery(
     graphql`
-      {
+      fragment portOrPutMethod on RamlResourcePost {
```
emmenko: Thanks. Can we also try to include the other methods get,delete? Most of the fields are the same. We would probably need to split it up in multiple smaller fragments.

packages/gatsby-theme-api-docs/src/components/resource/method/method.js
davifantasia: Personally, I would rather not include a default case. Only reason I did that is because of eslint. What do you think about this approach?

emmenko: The intent is to help preventing "bugs", for example here by forgetting to map a method.

If all the methods are mapped, there is no problem. If a method is not mapped though, there will be an error. Does that make sense?

packages/gatsby-theme-api-docs/src/hooks/use-api-resources.js
nkuehn:
```suggestion
      fragment postOrPutMethod on RamlResourcePost {
```

I guess you mean "post", not "port"? 

although "get", "options" and "head" do not support a request body in HTTP  they are modelled uniformly in RAML and can (+ should) have the identical data model in our implementaiton.  What you have here is rather a "httpMethod" fragment and not a method-specific one.

davifantasia: I think it's best we do that in another PR, only way I see a proper implementation is

- redefine the method schemas to extend a common interface
- write fragments and queries

emmenko: Sure, this can be a follow up. Thanks

PR #260, opened 2020-01-31
fix(theme): sidebar link border color

emmenko:
image


emmenko:
I would have expected the VRTs to catch this, but it could be that the color different is so thin that it wasn't recognized as a visual difference.

Issue #264, opened 2020-01-31
Use fragments to compose queries for post, put, get, delete


davifantasia:
There are too many fields being repeated in the query for post, put, get, delete methods.  See the current query here:

https://github.com/commercetools/commercetools-docs-kit/blob/master/packages/gatsby-theme-api-docs/src/hooks/use-api-resources.js

Since [methods in our RAML implementation are modelled uniformly](https://github.com/commercetools/commercetools-docs-kit/pull/257#discussion_r373460172), then we only need to do 2 things

- redefine methods (uniform) schema in gatsby-transformer-raml
- use fragment for query in `use-api-resources` hook

PR #272, opened 2020-02-03
refactor: enhanced graphql for resources

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/264

### DIFF
--- a/packages/gatsby-theme-api-docs/src/hooks/use-api-resources.js
+++ b/packages/gatsby-theme-api-docs/src/hooks/use-api-resources.js
@@ -3,7 +3,7 @@ import { useStaticQuery, graphql } from 'gatsby';
 export default () => {
   const queryResult = useStaticQuery(
     graphql`
-      fragment methods on RamlResourceMethod {
+      fragment methods on Method {
         securedBy {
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-api-docs/src/hooks/use-api-resources.js
```diff
@@ -3,7 +3,7 @@ import { useStaticQuery, graphql } from 'gatsby';
 export default () => {
   const queryResult = useStaticQuery(
     graphql`
-      fragment postOrPutMethod on RamlResourcePost {
+      fragment methods on RamlResourceMethod {
```
emmenko: Some methods do not have the `body`, we should differentiate the fragments between the different methods. We can have a "base" method fragment, then for the methods that have the extra fields 

```js
fragment methodBody on ??? {
  applicationjson {
    type
    builtinType
  }
}
fragment methodFields on RamlResourceMethod {
  # all fields, without body
}

query GetAllRamlResources {
  allRamlResource {
    nodes {
      post {
        ...methodFields
        body {
          ...methodBody
        }
      }
    }
  }
}
```

PR #277, opened 2020-02-05
refactor: use more semantic HTML elements for the search crawler

nkuehn:
This
 * marks more navigation elements as `<nav>` to exclude them from the body content flow
 * marks the top right corner area (e.g. Github link and version) as `<aside>` to exclude it from the content flow (when reading the DOM linearly it's positioned between the Title H1 and the content)
 * marks the page content as `<article>` to address it semantically without having to adress the complete `<main>` which includes more.  It would have been nicer to put the <main> nearer to the body content but the grid-oriented DOM does not allow that (main should not contain things that are the same across many sites like in our case the top nav)
 * gives the actual markdown generated body content a `body-content` ID in the DOM (for transparency, not directly needed yet)
 * introduces a `site-title` ID for the microsite name that is then adressable in the crawler config. 

Once this and the matching changes in the legacy site codebase are live we can ask Algolia to update the search crawler config to this proposed structure:
https://github.com/nkuehn/docsearch-configs/blob/commercetools-changes-2020-02/configs/commercetools.json

This updated crawler config would push the structure in Algolia "down", making it match our content hierarchy of  site->page->section->subsection  (currently the site level is ignored and all pages are flat next to each other, which is starting to become a usability issue because of same-named pages in a list.

### DIFF
--- a/packages/gatsby-theme-docs/src/components/content-pagination.js
+++ b/packages/gatsby-theme-docs/src/components/content-pagination.js
@@ -15,7 +15,7 @@ const trimTrailingSlash = url => url.replace(/(\/?)$/, '');
 
 const isMatching = (a, b) => trimTrailingSlash(a) === trimTrailingSlash(b);
 
-const Container = styled.div`
+const Container = styled.nav`
   display: grid;
   grid-gap: ${designSystem.dimensions.spacings.m};
   grid-auto-columns: 1fr;
--- a/packages/gatsby-theme-docs/src/components/content-pagination.js
+++ b/packages/gatsby-theme-docs/src/components/content-pagination.js
@@ -124,7 +124,7 @@ export const PurePagination = props => {
   const nextPage = activeChapter.pages[currentPageLinkIndex + 1];
 
   return (
-    <Container>
+    <Container aria-label="Next / Previous in Chapter Navigation">
       {hasPagination && previousPage ? (
         <PaginationLink
           linkTo={previousPage.path}
--- a/packages/gatsby-theme-docs/src/layouts/internals/layout-main.js
+++ b/packages/gatsby-theme-docs/src/layouts/internals/layout-main.js
@@ -3,7 +3,7 @@ import styled from '@emotion/styled';
 import { css } from '@emotion/core';
 import { designSystem } from '@commercetools-docs/ui-kit';
 
-const Container = styled.div`
+const Container = styled.main`
   grid-area: main;
   min-width: 0;
   position: relative;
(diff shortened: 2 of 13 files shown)


packages/gatsby-theme-docs/src/layouts/internals/layout-main.js
```diff
@@ -23,8 +23,6 @@ const Container = styled.main`
     `}
 `;
 
-const LayoutMain = props => (
-  <Container role="main" aria-label="Page content" {...props} />
```
nkuehn: `<main>` elements don't need an explicit ARIA "main" role, the element is implying the role already.

emmenko: Yes, it was mostly for backwards compatibility.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main#Browser_compatibility

packages/gatsby-theme-docs/src/components/content-pagination.js
```diff
@@ -15,7 +15,7 @@ const trimTrailingSlash = url => url.replace(/(\/?)$/, '');
 
 const isMatching = (a, b) => trimTrailingSlash(a) === trimTrailingSlash(b);
 
-const Container = styled.div`
+const Container = styled.nav`
```
emmenko: Please assign a `aria-label` to the `<Container>` element to differentiate it from other navigation elements

> If a role is used more than once on a page, the `aria-label` attribute should also be used in order to distinguish between the two regions.

```js
<Container aria-label="Page content pagination">
```

packages/gatsby-theme-docs/src/layouts/internals/layout-main.js
nkuehn: If it's just for "classic" IE I would say we can leave it out.

packages/gatsby-theme-docs/src/components/content-pagination.js
nkuehn: Learnt something, thanks !

packages/gatsby-theme-docs/src/layouts/internals/layout-main.js
emmenko: Sure, you can remove the `const LayoutMain = props => <Container {...props} />;` and rename the `Container` to `LayoutMain`.

emmenko:
Also, don't forget to label the PRs please. This one would either be "bug" or "refactoring". Thanks

nkuehn:
@emmenko  when reading about Aria Labels I learnt that they are used like headings or buttons  / actions by screen readers, so formally they are Title Cased (the examples in the reference are written that way, too). 

Plus, in conversations in the last days I realized that "index nav" is not self-explanatory,  it's commonly referred to as the table of contents of the page.  So I changed the label.

emmenko:
@nkuehn I think you forgot to rename some labels

image

nkuehn:
adjusted only half of the tests in the last commit, sorry.  fixed now.

PR #286, opened 2020-02-12
fix(theme): rendering of formatted code block after re-hydration

emmenko:
I noticed a weird behavior regarding code blocks for syntax highlighting. As you can see from the first gif, during the initial server side rendering the code is properly formatted and highlighted, however as soon as the client side part kicks in (the "hydration phase"), the syntax highlighting gets lost.

**Before**

After some digging I was able to pin point the issue with the usage of `dangerouslySetInnerHTML`, which is the React way of using `innerHTML` to inject an HTML fragment into the DOM. However, for some unknown reasons, when the code is re-hydrated the syntax highlighting gets lost.

So I decided to write a little parser that converts the HTML string into a React component, so we can render the code block content as a normal component. You can see below that the code is highlighted after the application is re-hydrated.

**After**


nkuehn:
Wow that's an interesting problem with a creative solution.  Did you try to contact the gatsbyJS project about it? Or is the issue rather due to us not actually using the standard highlight components?

emmenko:
The only difference I can think of is that the gatsby-remark-prismjs plugin [works as an actual plugin](https://github.com/gatsbyjs/gatsby/blob/1ecc0ec84f473f6e18ccb175ec96cb4dc3b62064/packages/gatsby-remark-prismjs/src/index.js#L33). In our implementation we use the internals of the plugins but render it as a MDX React component.

emmenko:
This will be fixed by #288

Issue #287, opened 2020-02-13
PrismJS is bloating the JS bundles


emmenko:
I noticed that our JS bundles are quite big (> 1Mb), especially the `common` bundle. I took a deeper look using the webpack bundle analyzer and saw that most of the size comes from the `prismjs` library.

<img width="1630" alt="Screenshot 2020-02-13 at 09 33 36" src="https://user-images.githubusercontent.com/1110551/74416453-76bc8e00-4e45-11ea-9aad-11bc8a7c1f5b.png">

**Why is that?**

When we render the code block component, we try to parse the raw content into a formatted HTML that uses the prismjs tokens.
Most of the code has been taken from the `gatsby-remark-prismjs` plugin which at some point loads the PrismJS components:

https://github.com/commercetools/commercetools-docs-kit/blob/046140a445a20b1d61a15ff92857a2d6bf6a9d26/packages/ui-kit/src/utils/code-block-load-prism-language.js#L3

There is also a `require` import to the component specific to a language:

https://github.com/commercetools/commercetools-docs-kit/blob/046140a445a20b1d61a15ff92857a2d6bf6a9d26/packages/ui-kit/src/utils/code-block-load-prism-language.js#L49

Because of those imports, Webpack needs to bundle all the PrismJS components even though only a bunch of them are used.

**What can we do?**

This is obviously bad and we need to fix it somehow.

Usually what we would do in such cases is to use dynamic imports, which tells Webpack to create a split bundle. However, we need to check first if SSR still works as it should.

If that doesn't work, we need to find another solution, as we want the code blocks to have the correct syntax highlighting also on SSR. It can also be that we would need to look at another library.

**Why not using the `gatsby-remark-prismjs`?**

The gatsby plugin works on the remark plugin level (Markdown), which replaces the `code` element with the  generated HTML. This is obviously the ideal solution, as importing the PrismJS stuff happens during the build phase, so Webpack does not need to bundle any extra dependency.

However, we couldn't use the default plugin for a bunch of reasons:
- The code block needs to be customized for things like header title, tooltip, copy to clipboard, etc. The default plugin does not allow this level of customization, and therefore it's not possible to use that.
- The code block component needs to be used programmatically in certain views, for example in the API docs and using the plugin won't work because it operates on the Markdown (remark) as opposed to the HTML (rehype).

I'll do some more digging and experiment a bit what is possible to do in order to reduce the bundle size.
[...]

PR #288, opened 2020-02-16
refactor: rendering of code block for syntax highlighting

emmenko:
Fixes: #287 
Closes: #286 

This PR refactors how the syntax highlighting is rendered and as a consequence, it **fixes 2 big problems** we were having!

- The syntax highlighting got lost after re-hydration. An attempt to fix this was done in #286.
- The entire PrismJS library was bundled, which caused the common bundle to be quite big. Now it's half the size it was before.

Screenshot 2020-02-16 at 20 55 40

The library we are using now it's `prism-react-renderer`. It offers a better API to render and customize the highlighted syntax. It also ships with a more thinner and optimized version of PrismJs, which does not blow up the bundle.


packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -183,6 +183,12 @@ module.exports = (themeOptions = {}) => {
           exclude: [],
         },
       },
+      {
+        resolve: 'gatsby-plugin-webpack-bundle-analyser-v2',
```
emmenko: This is useful to analyze the bundle, when necessary

packages/ui-kit/package.json
```diff
@@ -17,7 +17,6 @@
   "files": [
     "dist",
     "package.json",
-    "code-block-themes.js",
```
emmenko: This is in theory a breaking change. I still need to decide if we want to mark it as a breaking change or not, since this was only used by the gatsby theme.

packages/ui-kit/src/components/code-block.js
```diff
@@ -2,13 +2,15 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
+import { css } from '@emotion/core';
 import Tooltip from '@commercetools-uikit/tooltip';
 import SpacingsInline from '@commercetools-uikit/spacings-inline';
 import { ClipboardIcon } from '@commercetools-uikit/icons';
+import theme from 'prism-react-renderer/themes/nightOwl';
```
emmenko: FYI: previously we were using the "dracula" theme but I find this to be a bit "nicer".

packages/ui-kit/package.json
nkuehn: No, we're still in alpha releases which are always considered breaking.

packages/ui-kit/src/components/code-block.js
nkuehn: https://github.com/sdras/night-owl-vscode-theme
vs. 
https://github.com/dracula/visual-studio-code

I'm not a fan of mixing design changes with refactors but agree that nightOwl matches our CI better than dracula (I always found that strong violet a mismatch)

emmenko:
I'll merge and release since it's an important fix/change.

PR #290, opened 2020-02-17
Scroll to chapter, keep scroll pos without flicker

nkuehn:
!! This is a PR to merge from nm-nav-scroll-position--nk into nm-nav-scroll-position, not into master.  re-route as you like if merging directly to master is better. 
---
I think I tamed the flickering by switching to the synchronous  `useLayoutEffect()` hook and removing the extra layer of `useCallback()` nesting. 

UX change: If no previous position is known, the nav now scrolls the current chapter heading to the top instead of the individual page.  

The smoke test site's fake nav entries confuse it but e.g. in the API smoke test it's fine:
https://commercetools-docs-kit-git-nm-nav-scroll-position-nk.commercetools.now.sh/api-docs-smoke-test/endpoints/simple-resource-get

One click flow is not ideal yet but I can live with it:  After a reload (e.g. navigating from another site) the first click on a nav entry does not keep the position but scrolls to the top of the chapter if you never scrolled before.  So on initial load the position seems to not be set yet or something like that.

### DIFF
--- a/packages/gatsby-theme-docs/src/hooks/use-active-section.js
+++ b/packages/gatsby-theme-docs/src/hooks/use-active-section.js
@@ -1,6 +1,6 @@
 import React from 'react';
-import throttle from 'lodash.throttle';
 import { designSystem } from '@commercetools-docs/ui-kit';
+import useScrollSpy from './use-scroll-spy';
 
 const calculateOffset = () => {
   const pxNumber = designSystem.dimensions.heights.header.replace(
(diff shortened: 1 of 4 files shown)


emmenko:
@nkuehn thanks for investigating this further. I think you found the problem, which was using `useEffect` instead of `useLayoutEffect`.

I rebased the base branch with master and pushed a commit: https://github.com/commercetools/commercetools-docs-kit/pull/175/commits/6958a309ca340487a6f88f00e88478af785c4c7c

Please have a look at that branch deploy preview, I think we can close this PR then.

PR #296, opened 2020-02-26
feat: add ramldoc-generator package

davifantasia:
This plugin generate RAML documents compatible with [RAML transformer ](https://www.npmjs.com/package/@commercetools-docs/gatsby-transformer-raml)plugin


packages/raml-generator/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@commercetools-docs/raml-generator",
+  "version": "1.0.0",
+  "description": "Generates RAML docs compatible with @commercetools-docs/gatsby-transformer-raml",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "src/index.js",
```
emmenko: This should be removed. The package does not export any API or something to be used as a JS package

packages/raml-generator/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Typical usage in a static generator based website project is to include it as a development dependency
+
+`yarn add --dev @commercetools-docs/raml-generator`
+
+and then run it inside the project directory
+
+`npx commercetools-docs-raml-generator <api-spec-name> <api-spec-path>`
+
+The first parameter is the name of the API spec - this will be the directory name of the generated RAML docs so we recommend camel case or snake case naming convention.
```
emmenko: It would be better to provide a concrete example on what the name could be.

packages/raml-generator/src/cli.js
```diff
@@ -0,0 +1,55 @@
+/** Based on this blog - https://www.twilio.com/blog/how-to-build-a-cli-with-node-js */
```
emmenko: I would like that we use the same setup for our CLI scripts. For example, see the broken-link-checker.

We usually use `mri` for parsing command arguments. For small CLI this is perfectly sufficient, as the library is also very small compared to e.g. `arg` and `inquirer`.
Also, `inquirer` is not really necessary for this use case.

packages/raml-generator/src/main.js
```diff
@@ -0,0 +1,45 @@
+const fs = require('fs');
+const path = require('path');
+const shell = require('shelljs');
+const chalk = require('chalk');
+
+function generateRaml(options) {
+  const resolvedSourcePath = path.resolve(options.apiSpecSourcePath);
+
+  // Check if the spec file exists.
+  fs.access(resolvedSourcePath, fs.constants.F_OK, err => {
```
emmenko: There is a simpler function to use: `fs.existsSync`

packages/raml-generator/src/main.js
```diff
@@ -0,0 +1,45 @@
...
+    )}`
+  );
+
+  const jarFile = path.resolve(
+    __dirname,
+    '../jar/cli-application-1.0.0-20200221085820-all.jar'
+  );
+
+  shell.exec(
+    `java -jar ${jarFile} generate ${options.apiSpecSourcePath} -o ${resolvedDestinationPath}/${options.apiSpecName} -t ramldoc`
```
emmenko: The script requires users to have `java`. This **needs to be mentioned** in the README! Also which java version etc.

packages/raml-generator/src/main.js
```diff
@@ -0,0 +1,45 @@
...
+
+  console.log(
+    `RAML docs will be generated in this directory: ${chalk.blue(
+      resolvedDestinationPath
+    )}`
+  );
+
+  const jarFile = path.resolve(
+    __dirname,
+    '../jar/cli-application-1.0.0-20200221085820-all.jar'
```
emmenko: Why is this hardcoded??

packages/raml-generator/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@commercetools-docs/raml-generator",
```
emmenko: The name "RAML generator" is a bit confusing for this package. I would suggest to find a more appropriate name, like `ramldoc-generator`.

packages/raml-generator/src/main.js
davifantasia: Because the location is consistent.

packages/raml-generator/src/main.js
davifantasia: I'm not sure what you mean - it's not "simpler" from my perspective.

packages/raml-generator/src/main.js
emmenko: What I mean is that the file name should be read programmatically and not hardcoded. This is error prone as if the jar file name changes, and you forget to update it here, the script won't work anymore.

```js
const fileName = fs.readdirSync(path.join(__dirname, '../jar'))
  .find(name => name.startsWith('cli-application'));
```

packages/raml-generator/src/cli.js
davifantasia: I don't understand what you mean by inquirer is not necessary for this use case.

packages/raml-generator/src/main.js
emmenko:
```js
if (!fs.existsSync(resolvedSourcePath)) {
  // do something
}
```

packages/raml-generator/src/cli.js
emmenko: It's not necessary to prompt the user to input the value, when you can simply pass them as command arguments.

Prompting usually is used for things like generating or initializing a project or something, where the user needs to provide multiple information.

davifantasia: > I would like that we use the same setup for our CLI scripts. For example, see the broken-link-checker.

I don't understand what you mean, it doesn't appear to me that the setup should be similar, at least one main reason is broken link checker does not take arguments.

emmenko: I'm referring to the part where you parse the arguments, as this is the only case where a library is useful that does that for you.

https://github.com/commercetools/commercetools-docs-kit/blob/14581f1d3d715de2cdf0d3e231b7f49777c8aa13/packages/broken-link-checker/bin/broken-link-checker.js#L11-L36

You can also check in the appkit, as we use it in all our scripts: https://github.com/commercetools/merchant-center-application-kit/search?q=mri&unscoped_q=mri

davifantasia: I use `mri` now.

packages/raml-generator/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@commercetools-docs/ramldoc-generator",
```
emmenko: The package folder also should be renamed

packages/raml-generator/package.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "keywords": [
+    "cli",
+    "ramldoc-generator"
+  ],
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "bin": {
+    "commercetools-docs-ramldoc-generator": "bin/ramldoc-generator.js"
```
emmenko: We can name the binary like this (we do the same with the broken link checker):

```suggestion
    "commercetools-ramldoc-generator": "bin/ramldoc-generator.js"
```

packages/raml-generator/src/cli.js
emmenko: Thanks but we should also remove `inquirer`, as I explained before

packages/raml-generator/src/main.js
davifantasia: Thanks for explaining, but your suggestion is still prone to the same error you mentioned from my perspective. But I'll make the changes as you want it.

emmenko: The code snippet was just an example to illustrate what I meant.
I don't know how the process to update the jar file is, so the final implementation should be based on that. I would assume that only 1 jar is included in the package.
The point is to make the updates to the jar file to not break the script.

davifantasia: Yes, there is only one jar file.

To update the jar file, manually downloading and copy-paste is necessary. I guess one benefit of manual download is that we can rename the file to exactly what we want and then keep that naming consistent.

emmenko: Yep, that is another option

packages/raml-generator/src/cli.js
davifantasia: From my perspective, `mri` does not perform the same function as `inquirer`, so I disagree when why you say `inquirer` is not necessary.

packages/raml-generator/package.json
davifantasia: I don't understand why this is necessary.

emmenko: Reasons:
- consistency with the other existing binaries
- shorter name
- unnecessary information in the name, the `commercetools` prefix is enough

emmenko: Reasons:
- consistency with the other existing binaries
- shorter name
- unnecessary information in the name, the `commercetools` prefix is enough

packages/ramldoc-generator/src/main.js
```diff
@@ -0,0 +1,40 @@
+const fs = require('fs');
+const path = require('path');
+const shell = require('shelljs');
+const chalk = require('chalk');
+
+function generateRaml(options) {
+  const resolvedSourcePath = path.resolve(options.apiSpecSourcePath);
```
emmenko: Can you explain the reason for using `path.resolve`?

packages/ramldoc-generator/src/main.js
```diff
@@ -0,0 +1,40 @@
...
+
+  // Check if the spec file exists.
+  if (!fs.existsSync(resolvedSourcePath)) {
+    console.error(
+      `"${resolvedSourcePath}" does not exist`,
+      chalk.red.bold('ERROR')
+    );
+    process.exit(1);
```
emmenko: NOTE: usually the library implementation should throw an `Error` and the file handling the CLI part should catch it and exit the process.

packages/ramldoc-generator/src/main.js
```diff
@@ -0,0 +1,40 @@
...
+function generateRamlIfSpecPath(options) {
+  const resolvedDestinationPath = path.resolve(options.apiSpecDestinationPath);
+
+  console.log(
+    `RAML docs will be generated in this directory: ${chalk.blue(
+      resolvedDestinationPath
+    )}`
+  );
+
+  const jarFile = path.resolve(__dirname, '../jar/rmf-codegen.jar');
```
emmenko:
```suggestion
  const jarFile = path.join(__dirname, '../jar/rmf-codegen.jar');
```

packages/ramldoc-generator/src/main.js
davifantasia: I prefer to work with absolute path, in case a user passes a relative path.

packages/ramldoc-generator/src/main.js
davifantasia: Why is this better?

davifantasia: Why is this better?

packages/raml-generator/src/cli.js
davifantasia: So how do you specify optional arguments?

packages/raml-generator/package.json
davifantasia: So why do we have "commercetools-docs" scope or namespace instead of "commercetools"?

packages/ramldoc-generator/src/main.js
emmenko: `__dirname` is already an absolute path, and you just want to append other partial paths.

That's what `path.join` is for. `path.resolve` does not work as you would expect in this case. See below:

```
> path.resolve(process.cwd(), '/foo')
'/foo'
> path.join(process.cwd(), '/foo')
'/Users/emmenko/dev/src/foo'

> path.resolve(process.cwd(), '..', '/foo')
'/foo'
> path.join(process.cwd(), '..', '/foo')
'/Users/emmenko/dev/foo'
```

packages/ramldoc-generator/src/main.js
emmenko: Ok thanks

packages/raml-generator/package.json
emmenko: The npm/package.json scope has nothing to do with the binary name

packages/raml-generator/src/cli.js
emmenko: Check here for an example: https://github.com/commercetools/merchant-center-application-kit/blob/d4273e55d3de7e99d7f4bba6aeaf786a2bbf63a8/packages/create-mc-app/bin/create-mc-app.js#L14-L33

Probably it makes sense to specify the src/dest paths as options:

```
commercetools-ramldoc-generator [spec_name] [options]

--src   <path> The path to the source file
--dest <path> (optional) The path to the destination file
```

The `dest` option has a default value, if none is provided.

packages/ramldoc-generator/src/main.js
davifantasia: Why use `process.cwd()`?

Anyway, in my case it works exactly as expected.

emmenko: process.cwd() was just for the example, as I run the code in the node repl and __dirname is not available there.

Using path.resolve is fine if you know what you’re doing, but it might lead to unexpected results.
That’s why usually for building paths you would use path.join.

packages/ramldoc-generator/src/main.js
```diff
@@ -0,0 +1,36 @@
...
+function generateRamlIfSpecPath(options) {
+  const resolvedDestinationPath = path.resolve(options.dest);
+
+  console.log(
+    `RAML docs will be generated in this directory: ${chalk.blue(
+      resolvedDestinationPath
+    )}`
+  );
+
+  const jarFile = path.resolve(__dirname, '../jar/rmf-codegen.jar');
```
emmenko: @davifantasia can you explain the process of getting the JAR file? We should also document this...

cc @nkuehn

davifantasia: I don't know how to go about documenting that process:

- is it a part of the ReadMe?
- what kind of title would it have if documented?
- is the documentation targeted at users or someone who is just interested in how the package works?
- do you have an existing structure for documenting something like this?
- other things you want to add.

emmenko: You can add a file `CONTRIBUTING.md` and describe there where to get the JAR file. This is usually targeted to people that wants to contribute to the development of the package.
Or just add a section to the `README.md`.

packages/ramldoc-generator/src/cli.js
```diff
@@ -0,0 +1,46 @@
...
+    process.exit(1);
+  }
+}
+
+function parseArgumentsIntoOptions(rawArgs) {
+  const argv = rawArgs.slice(2);
+
+  const args = mri(argv, {
+    alias: { help: ['h'] },
+    default: { help: false, name: '', src: '', dest: './src/api-specs/' },
```
emmenko: Only default values should be defined here

```suggestion
    default: { help: false, dest: './src/api-specs/' },
```

packages/ramldoc-generator/src/cli.js
```diff
@@ -0,0 +1,46 @@
...
+
+async function promptForMissingOptions(options) {
+  if (options.help || !(options.name && options.src)) {
+    console.log(`
+    Usage: commercetools-ramldoc-generator --name <api-spec-name> --src <api-spec-source-path>
+
+    Displays help information.
+
+    Options:
+      --dest <api-spec-destination-path>               (optional) the path for the generated RAML documents."
```
emmenko:
```suggestion
      --name <api-spec-name>                      The name of the generated spec file.
      --src <api-spec-source-path>               The path to the RAML file spec.
      --dest <api-spec-destination-path>      (optional) The path to the folder where to write the generated spec file (default: "./src/api-specs/").
```

packages/ramldoc-generator/src/main.js
```diff
@@ -0,0 +1,36 @@
...
+const path = require('path');
+const shell = require('shelljs');
+const chalk = require('chalk');
+
+function generateRaml(options) {
+  const resolvedSourcePath = path.resolve(options.src);
+
+  // Check if the spec file exists.
+  if (!fs.existsSync(resolvedSourcePath)) {
+    throw new Error(chalk.red.bold(`"${resolvedSourcePath}" does not exist`));
```
emmenko: Nitpick: you should use `chalk` in the `cli.js` file, not in the error message

packages/ramldoc-generator/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+For globally installation, there is an optional third parameter that should be passed to specify the location for the generated RAML like so:
+
+`npx commercetools-ramldoc-generator --name <api-spec-name> --src <api-spec-source-path> --dest <api-spec-destination-path>`
+
+## RMF-Codegen Jar File
+
+The [rmf-codegen](https://github.com/vrapio/rmf-codegen) jar file is deployed on [Bintray](https://bintray.com/vrapio/vrapio/rmf-codegen#files/io%2Fvrap%2Frmf%2Fcodegen%2Fcli-application%2F1.0.0-20200221085820). A direct download link of the cli application, that includes necessary dependencies, is found here:
+
+https://dl.bintray.com/vrapio/vrapio/io/vrap/rmf/codegen/cli-application/1.0.0-20200221085820/
```
emmenko: This is the link to a specific version, we should just use:

```suggestion
https://dl.bintray.com/vrapio/vrapio/io/vrap/rmf/codegen/cli-application/
```

packages/ramldoc-generator/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+If the module is locally installed in the project, the generated raml documents will be located in `./src/api-specs`.
+
+For globally installation, there is an optional third parameter that should be passed to specify the location for the generated RAML like so:
+
+`npx commercetools-ramldoc-generator --name <api-spec-name> --src <api-spec-source-path> --dest <api-spec-destination-path>`
+
+## RMF-Codegen Jar File
+
+The [rmf-codegen](https://github.com/vrapio/rmf-codegen) jar file is deployed on [Bintray](https://bintray.com/vrapio/vrapio/rmf-codegen#files/io%2Fvrap%2Frmf%2Fcodegen%2Fcli-application%2F1.0.0-20200221085820). A direct download link of the cli application, that includes necessary dependencies, is found here:
```
emmenko:
```suggestion
The [rmf-codegen](https://github.com/vrapio/rmf-codegen) jar file is deployed on [Bintray](https://bintray.com/vrapio/vrapio/rmf-codegen). A direct download link of the cli application, that includes necessary dependencies, is found here:
```

websites/api-docs-smoke-test/package.json
```diff
@@ -10,11 +10,16 @@
...
+    "postbuild": "rm -rf ../../public/api-docs-smoke-test && mkdir -p ../../public && mv public ../../public/api-docs-smoke-test",
+    "generate-ramldoc": "npx commercetools-ramldoc-generator --name test --src ./source-raml/test/api.raml"
   },
   "dependencies": {
     "@commercetools-docs/gatsby-theme-api-docs": "1.8.1",
     "gatsby": "2.19.17",
     "gatsby-cli": "2.8.29"
+  },
+
+  "devDependencies": {
```
nkuehn: If that is possible (i.e. resolves the binary) I suggest to move that dependency into the @commercetools-docs/gatsby-theme-api-docs theme package so we don't have to maintain it in every website package.json.

packages/ramldoc-generator/scripts/download-rmf-codegen.js
```diff
@@ -0,0 +1,58 @@
...
+      reject(error);
+    });
+    fileStream.on('finish', resolve);
+  });
+
+  // Assign proper write permissions to the file
+  abortIfError(shelljs.chmod(755, jarPath));
+};
+
+if (!hasbin.sync('java')) {
```
emmenko: It's sufficient to do something like this:

```js
const isJavaInstalled = () => {
  try {
    const result = shelljs.exec('java -version', { silent: true });
    return result.code === 0;
  } catch (error) {
    return false;
  }
}
```

Let's avoid using unnecessary dependencies.

packages/ramldoc-generator/src/cli.js
```diff
@@ -0,0 +1,49 @@
+const mri = require('mri');
+const chalk = require('chalk');
+const { generateRaml } = require('./main');
+
+async function cli(args) {
+  const options = parseArgumentsIntoOptions(args);
+  promptForMissingOptions(options);
+
+  try {
+    await generateRaml(options);
```
emmenko: The `generateRaml` function does not seem to be async.

packages/ramldoc-generator/src/cli.js
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  return {
+    name: args.name,
+    src: args.src,
+    dest: args.dest,
+  };
+}
+
+async function promptForMissingOptions(options) {
```
emmenko: Also not async

Issue #64, opened 2019-11-14
Component: Single Language code example from file


nkuehn:
> update: here's a branch with experiments in an actual MDX file: https://github.com/commercetools/commercetools-docs-kit/blob/extended-code-syntax-brainstorming/websites/docs-smoke-test/src/content/code-samples/multi-language-code.mdx

Alternatively to writing example code in markdown as fenced code blocks an author must be able to store the example code in a separate standalone file and pull it into any MDX page.  This provides the advantage that the code examples can be own fully valid software projects that are automatically tested. 

This issue covers the case of a single-language example like fenced code blocks. 

We decided to leverage MDX import statements and custom webpack loaders, which is how gatsbyJS is generally built internally, too.   Webpack can be configured cleanly by a plugin like the theme (see reference link below)

Any import statement loading from a path containing `/code-examples/` (plus maybe further restrictions like whitelisting packages if possible) is managed by a custom loader that returns a function component that is wrapping our standard Code component.  It's detecting the language by file name and injects the content. 

The following code is pseudo-MDX
```md
import CartCreateExampleJs from '@commercetools-docs/code-examples/javascript/cart/create.js'
import CartCreateExampleJava from '../../code-examples/java/cart/create.java'

<CartCreateExampleJs title="Creating a Cart in Javascript" />
<CartCreateExampleJava highlightLines="5-10" title="Creating a Cart in Java" />

```

## Not necessary

We don't need to support cases that cannot be derived from the file name extension (e.g.  individual terminal cli commands that are not part of a shell script)

## References:
 * https://www.gatsbyjs.org/docs/add-custom-webpack-config/
 * how it's currently written in markdown (Liquid templating using a parameterized template import):  https://github.com/commercetools/docs.commercetools.com/blob/master/sdk/jvm-sdk.md  (private repository)
 * https://prismjs.com/plugins/file-highlight/   (not nice because it does an HTTP call at runtime from the browser)

nkuehn:
update from a conversation with @emmenko  about how to approach the topic:
 * we will go for an import and webpack-based approach. Users learn the import notation and have to lean it anyways for MDX transclusion.  
 * code examples from outside the repository will be provided via a special package in the respective content repository which pulls the examples from whereever they live (e.g. the repository of an SDK) and then provides the files as a javascript package. 

I'll update the issue description here

FYI @davifantasia

davifantasia:
I just want to say I started working on this.

[...]

PR #299, opened 2020-03-02
Code sample from file

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/64


packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -209,6 +210,18 @@ exports.onCreateWebpackConfig = ({ actions, getConfig }) => {
         },
       ],
     },
+    {
+      test: /\.(js|java)$/,
```
emmenko: This shouldn't be hardcoded

packages/gatsby-theme-docs/src/components/code-sample-from-file.js
```diff
@@ -0,0 +1,24 @@
...
+import { CodeBlock } from '@commercetools-docs/ui-kit';
+
+// import Java from '../code-samples/test.java';
+
+// CodeBlockFromImport
+function CodeSampleFromFile(props) {
+  const [sample, setSample] = useState('');
+
+  // props.import - path to code sample
+  import(`../code-samples/${props.name}`).then(result => {
```
emmenko: How does this work with SSR?

packages/gatsby-theme-docs/src/components/code-sample-from-file.js
```diff
@@ -0,0 +1,24 @@
...
+    setSample(result.default);
+  });
+
+  return <CodeBlock content={sample} language={props.language} />;
+}
+
+CodeSampleFromFile.propTypes = {
+  name: PropTypes.string,
+  language: PropTypes.string,
```
emmenko: The language should be derived by the file name extension, we don't need to pass it explicitly.

packages/gatsby-theme-docs/src/loaders/code-sample.js
```diff
@@ -0,0 +1,8 @@
+function codeSample(content) {
+  this.cacheable();
+
+  this.value = content;
+  return `module.exports = ${JSON.stringify(content)}`;
```
emmenko: So basically you're wrapping the file content into an export.

What type is the `content`? A string or what?

packages/gatsby-theme-docs/src/components/code-sample-from-file.js
emmenko: Why not using a `require`?

davifantasia:
Closing this in favour of https://github.com/commercetools/commercetools-docs-kit/pull/300

PR #300, opened 2020-03-05
Code Sample Import - Different Approach

davifantasia:
### `CodeExample` Component

This exposes `CodeExample` component from the base theme that can take props available for use in `CodeBlock` component like

- title
- highlightLines
- noPromptLines

The main prop for `CodeExample` is `import`, which is the path to the code sample file.

Other `CodeBlock` props (language, content) are generated within the component.

Note that code example files should be put in `src/code-examples` directory.

#### Example usage:

`<CodeExample import="/src/code-examples/test.js" title="JavaScript Code Sample" />`

### Code Examples Generator Node Module

This tool is built to be used in conjunction with the `CodeExample` component. It generates files in `src/code-examples` from any node module, but it is expected that said node module should be dedicated to code samples.

#### Example usage:

Include as dependency

`yarn add --dev @commercetools-docs/code-examples-generator`

and then run it inside the project directory

`npx commercetools-code-examples-generator --packagename <code-examples-node-module-package-name> --sourcedir <optional-name-of-specific-directory-from-node-module>`

### Use Cases Addressed

- Code samples imported from external file from within the same directory.
- Code samples from a package, this could be managed by any team.

### Others

Closes https://github.com/commercetools/commercetools-docs-kit/issues/64

Replaces https://github.com/commercetools/commercetools-docs-kit/pull/299


nkuehn:
Hi @davifantasia and @jenschude , can we schedule a time where you explain the background and rationale of this package?  Superficially it seems to be a workaround to issues with the babel / import behavior of the CodeExample  Component.   

But since copying files into the local content directories (version controlled) does not look to me like a workflow that would work reliably in the tech writer teams we need to speak about this before you invest any more  time. 

I could be that it's even a favorable workflow but I want to be sure about this before you invest any more time.

nkuehn:
update from today's conversation:  This serves a purpose and we should keep the idea open but we have no current need for it since it would cause confusing inside a repository where both the source package with the examples and the copied targets live in one place. 

So "On Hold" for the time being.

PR #304, opened 2020-03-09
feat: code examples provided on graphQL by a transformer plugin 

davifantasia:
We chose to implement and provide a practically working approach that uses the same architecture as the RAML type rendering to be able to move on with the migration of the site independenly of whether we get the babel / webpack loader <-> MDX  related issues sorted out. 

### Gatsby Transformer Code Examples

`@commercetools-docs/gatsby-transformer-code-examples` transforms file contents into CodeExample types in available in graphql. Supported file formats are:

- cs
- graphql
- java
- js
- json
- php
- sh
- ts
- yaml

### `CodeExample` Component

`CodeExample` component uses the hook query `use-code-examples.js` to query CodeExample types and then return content wrapped with `CodeBlock` component.

Technically, any file in the project, with supported format can be rendered as an example but it is recommended to stick the the `code-examples` directory for this use case.

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -88,6 +88,14 @@ module.exports = (themeOptions = {}) => {
           path: path.resolve(`./src/content`),
         },
       },
+      // Code samples
+      {
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'code-examples',
+          path: path.resolve(`./src/code-examples`),
+        },
+      },
 
       /**
        * Transformers for making content available in graphql queries
@@ -98,6 +106,9 @@ module.exports = (themeOptions = {}) => {
 
       // For querying configuration data
       'gatsby-transformer-yaml',
+
+      // For querying code samples
+      `@commercetools-docs/gatsby-transformer-code-examples`,
 
       // For querying MDX
       {
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -107,7 +107,7 @@ module.exports = (themeOptions = {}) => {
       // For querying configuration data
       'gatsby-transformer-yaml',
 
-      // For querying code samples
+      // For querying code examples
       `@commercetools-docs/gatsby-transformer-code-examples`,
 
       // For querying MDX
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -92,7 +92,7 @@ module.exports = (themeOptions = {}) => {
       {
         resolve: 'gatsby-source-filesystem',
         options: {
-          name: 'code-examples',
+          name: 'codeExamples',
           path: path.resolve(`./src/code-examples`),
         },
       },
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -88,7 +88,7 @@ module.exports = (themeOptions = {}) => {
           path: path.resolve(`./src/content`),
         },
       },
-      // Code samples
+      // Code examples
       {
         resolve: 'gatsby-source-filesystem',
         options: {
--- a/packages/gatsby-theme-docs/src/components/code-example.js
+++ b/packages/gatsby-theme-docs/src/components/code-example.js
@@ -7,7 +7,7 @@ function CodeExample(props) {
   const codeExamples = useCodeExamples();
 
   const codeExample = codeExamples.find(example => {
-    return example.absolutePath.includes(props.file);
+    return example.absolutePath.endsWith(props.file);
   });
 
   if (!codeExample) {
--- a/packages/gatsby-theme-docs/src/components/code-example.js
+++ b/packages/gatsby-theme-docs/src/components/code-example.js
@@ -11,7 +11,7 @@ function CodeExample(props) {
   });
 
   if (!codeExample) {
-    return reportError('Code example does not exist');
+    return reportError(`Unable to find file "${props.file}" within the "code-examples" folder. Make sure that the file exists and matches the value passed to the "file" property.`);
   }
 
   const language = props.file.split('.').pop();
--- a/packages/gatsby-theme-docs/src/components/code-example.js
+++ b/packages/gatsby-theme-docs/src/components/code-example.js
@@ -1,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CodeBlock, ContentNotifications } from '@commercetools-docs/ui-kit';
+import useCodeExamples from '../hooks/use-code-examples';
+
+function CodeExample(props) {
+  const codeExamples = useCodeExamples();
+
+  const codeExample = codeExamples.find(example => {
+    return example.absolutePath.includes(props.file);
+  });
+
+  if (!codeExample) {
+    return reportError('Code example does not exist');
+  }
+
+  const language = props.file.split('.').pop();
+
+  return (
+    <CodeBlock
+      content={codeExample.content}
+      language={language}
+      title={props.title}
+      highlightLines={props.highlightLines}
+      noPromptLines={props.noPromptLines}
+    />
+  );
+}
+
+function reportError(errorMsg) {
+  if (process.env.NODE_ENV !== 'production') {
+    return <ContentNotifications.Error>{errorMsg}</ContentNotifications.Error>;
+  }
+
+  throw new Error(errorMsg);
+}
+
+CodeExample.propTypes = {
+  file: PropTypes.string.isRequired,
+  title: PropTypes.string,
+  highlightLines: PropTypes.arrayOf(PropTypes.number),
+  noPromptLines: PropTypes.arrayOf(PropTypes.number),
+};
+
+export default CodeExample;
--- a/packages/gatsby-theme-docs/src/components/code-example.js
+++ b/packages/gatsby-theme-docs/src/components/code-example.js
@@ -11,15 +11,15 @@ function CodeExample(props) {
   });
 
   if (!codeExample) {
-    return reportError(`Unable to find file "${props.file}" within the "code-examples" folder. Make sure that the file exists and matches the value passed to the "file" property.`);
+    return reportError(
+      `Unable to find file "${props.file}" within the "code-examples" folder. Make sure that the file exists and matches the value passed to the "file" property.`
+    );
   }
-
-  const language = props.file.split('.').pop();
 
   return (
     <CodeBlock
       content={codeExample.content}
-      language={language}
+      language={codeExample.extension}
       title={props.title}
       highlightLines={props.highlightLines}
       noPromptLines={props.noPromptLines}
--- a/packages/gatsby-theme-docs/src/components/code-example.js
+++ b/packages/gatsby-theme-docs/src/components/code-example.js
@@ -7,7 +7,7 @@ function CodeExample(props) {
   const codeExamples = useCodeExamples();
 
   const codeExample = codeExamples.find(example => {
-    return example.path.endsWith(props.path);
+    return example.path === props.path;
   });
 
   if (!codeExample) {
--- a/packages/gatsby-theme-docs/src/components/code-example.js
+++ b/packages/gatsby-theme-docs/src/components/code-example.js
@@ -7,19 +7,19 @@ function CodeExample(props) {
   const codeExamples = useCodeExamples();
 
   const codeExample = codeExamples.find(example => {
-    return example.absolutePath.endsWith(props.file);
+    return example.path.endsWith(props.path);
   });
 
   if (!codeExample) {
     return reportError(
-      `Unable to find file "${props.file}" within the "code-examples" folder. Make sure that the file exists and matches the value passed to the "file" property.`
+      `Unable to find file "${props.path}" within the "code-examples" folder. Make sure that the file exists and matches the value passed to the "file" property.`
     );
   }
 
   return (
     <CodeBlock
       content={codeExample.content}
-      language={codeExample.extension}
+      language={codeExample.language}
       title={props.title}
       highlightLines={props.highlightLines}
       noPromptLines={props.noPromptLines}
@@ -36,7 +36,7 @@ function reportError(errorMsg) {
 }
 
 CodeExample.propTypes = {
-  file: PropTypes.string.isRequired,
+  path: PropTypes.string.isRequired,
   title: PropTypes.string,
   highlightLines: PropTypes.arrayOf(PropTypes.number),
   noPromptLines: PropTypes.arrayOf(PropTypes.number),
--- a/packages/gatsby-theme-docs/src/hooks/use-code-examples.js
+++ b/packages/gatsby-theme-docs/src/hooks/use-code-examples.js
@@ -1,0 +1,20 @@
+import { useStaticQuery, graphql } from 'gatsby';
+
+export default () => {
+  const queryResult = useStaticQuery(
+    graphql`
+      query GetCodeExamplesQuery {
+        allCodeExample {
+          nodes {
+            name
+            extension
+            absolutePath
+            content
+          }
+        }
+      }
+    `
+  );
+
+  return queryResult.allCodeExample.nodes;
+};
--- a/packages/gatsby-theme-docs/src/hooks/use-code-examples.js
+++ b/packages/gatsby-theme-docs/src/hooks/use-code-examples.js
@@ -7,8 +7,8 @@ export default () => {
         allCodeExample {
           nodes {
             name
-            extension
-            absolutePath
+            language
+            path
             content
           }
         }
--- a/packages/gatsby-theme-docs/src/hooks/use-code-examples.js
+++ b/packages/gatsby-theme-docs/src/hooks/use-code-examples.js
@@ -3,7 +3,7 @@ import { useStaticQuery, graphql } from 'gatsby';
 export default () => {
   const queryResult = useStaticQuery(
     graphql`
-      query GetCodeExamplesQuery {
+      query GetAllCodeExamplesQuery {
         allCodeExample {
           nodes {
             name
--- a/packages/gatsby-transformer-code-examples/README.md
+++ b/packages/gatsby-transformer-code-examples/README.md
@@ -8,7 +8,7 @@ This plugin transforms file contents into CodeExample types in available in grap
 
 ## Usage
 
-As a prerequisite configure a source plugin, e.g. `gatsby-source-filesystem` to point the directory of the code examples, for example `src/code-examples`.  Other source plugins can be used, too. 
+As a prerequisite configure a source plugin, e.g. `gatsby-source-filesystem` to point the directory of the code examples, for example `src/code-examples`. Other source plugins can be used, too.
 
 Example `gatsby-config.js` content:
 
@@ -19,7 +19,7 @@ module.exports = {
     {
       resolve: `gatsby-source-filesystem`,
       options: {
-        name: `code-examples`,
+        name: `codeExamples`,
         path: `${__dirname}/src/code-examples`,
       },
     },
@@ -31,7 +31,7 @@ module.exports = {
 ## GraphQl Query Example
 
 ```graphql
-query GetCodeExamplesQuery {
+query GetAllCodeExamplesQuery {
   allCodeExample {
     nodes {
       name
@@ -43,14 +43,13 @@ query GetCodeExamplesQuery {
 }
 ```
 
-## Supported File Formats
+## Supported MIME Types
 
-- cs
-- graphql
-- java
-- js
-- json
-- php
-- sh
-- ts
-- yaml
+- application/javascript
+- text/x-java-source
+- application/octet-stream
+- application/json
+- application/x-httpd-php
+- application/x-sh
+- video/mp2t
+- text/yaml
--- a/packages/gatsby-transformer-code-examples/README.md
+++ b/packages/gatsby-transformer-code-examples/README.md
@@ -47,12 +47,16 @@ query GetAllCodeExamplesQuery {
 
 ## Supported MIME Types
 
-- application/javascript
-- text/x-java-source
-- application/octet-stream
-- application/json
-- application/x-httpd-php
-- application/x-sh
-- video/mp2t
-- text/yaml
-- text/vnd.curl
+- `application/javascript` (JavaScript)
+- `text/x-java-source` (Java)
+- `application/json` (JSON)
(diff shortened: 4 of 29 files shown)


davifantasia:
@nkuehn per our discussion, I extracted this part to only implement the support of rendering code example files from an mdx document.

emmenko:
This is a nice approach! I'll take a look later.

PS: @davifantasia the label to the PR is missing 😜

packages/gatsby-theme-docs/src/components/code-example.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const codeExample = codeExamples.find(example => {
+    return example.absolutePath.includes(props.file);
+  });
+
+  if (!codeExample) {
+    return reportError('Code example does not exist');
+  }
+
+  const language = props.file.split('.').pop();
```
nkuehn: You have put the extension property into your GraphQL surface so I think you should use it here, too instead of re-parsing it from the file path.

packages/gatsby-transformer-code-examples/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+This plugin transforms file contents into CodeExample types in available in graphql.
+
+## Installation
+
+`npm install --save @commercetools-docs/gatsby-transformer-code-examples`
+
+## Usage
+
+As a prerequisite configure `gatsby-source-filesystem` plugin to point the directory of the code examples, for example `src/code-examples`.
```
nkuehn:
```suggestion
As a prerequisite configure a source plugin, e.g. `gatsby-source-filesystem` to point the directory of the code examples, for example `src/code-examples`.  Other source plugins can be used, too. 
```

packages/gatsby-theme-docs/src/components/code-example.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CodeBlock, ContentNotifications } from '@commercetools-docs/ui-kit';
+import useCodeExamples from '../hooks/use-code-examples';
+
+function CodeExample(props) {
+  const codeExamples = useCodeExamples();
+
+  const codeExample = codeExamples.find(example => {
+    return example.absolutePath.includes(props.file);
```
nkuehn: I find filtering using `includes()` relatively adventurous because many can match (even if you would only look at the end of the string. 

Why not require an exact match?  (see below in the plugin for my proposal concerning filtering the graphQL by path prefix)

packages/gatsby-transformer-code-examples/README.md
```diff
@@ -0,0 +1,56 @@
...
+module.exports = {
+  plugins: [
+    {
+      resolve: `gatsby-source-filesystem`,
+      options: {
+        name: `code-examples`,
+        path: `${__dirname}/src/code-examples`,
+      },
+    },
+    `@commercetools-docs/gatsby-transformer-code-examples`,
```
nkuehn: Since the plugin is putting every node from the complete Graphql that has the matching media type and all the file contents are loaded into the client side bundle in our theme I would feel better having a path prefix configuration in the code examples transformer that filters out only the content of a specific folder. 

default can be `/` but we would put  `/src/code-examples/` into our theme's config.  That would also have the nice side effect that users would not have to explicitly put  the  `code-examples` part into the MDX although it's conventional anyways.

packages/gatsby-transformer-code-examples/gatsby-node.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+async function onCreateNode({
+  node,
+  actions,
+  loadNodeContent,
+  createNodeId,
+  createContentDigest,
+}) {
+  const validExtensions = [
```
nkuehn: I suggest to rename it to `validMediaTypes`  because that's what's actually happening.   Some media types are linked to multiple extensions.

packages/gatsby-transformer-code-examples/gatsby-node.js
```diff
@@ -0,0 +1,74 @@
...
+    'application/javascript',
+    'text/x-java-source',
+    'application/octet-stream',
+    'application/json',
+    'application/x-httpd-php',
+    'application/x-sh',
+    'video/mp2t',
```
nkuehn: Maybe add an explanatory comment that this is typescript, which does not (yet) have an own media type.

packages/gatsby-transformer-code-examples/gatsby-node.js
```diff
@@ -0,0 +1,74 @@
...
+    createParentChildLink({ parent: node, child: codeSampleNode });
+  }
+
+  const content = await loadNodeContent(node);
+
+  transformObject(
+    {
+      name: node.name,
+      extension: node.extension,
+      absolutePath: node.absolutePath,
```
nkuehn: If you choose to allow a prefix, then the path should rather be the relative path.  I would prefer to not work with absolute paths at all because they can be missing or totally different strings in source plugins that are not filesystem (but still provide the content, e.g. the git plugin.

packages/gatsby-transformer-code-examples/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@commercetools-docs/gatsby-transformer-code-examples",
+  "version": "1.0.0",
+  "license": "MIT",
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "description": "Provides file contents as CodeExample GraphQl types.",
```
nkuehn:
```suggestion
  "description": "Provides contents of nodes with source code mime types as CodeExample GraphQl types.",
```

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -94,6 +94,14 @@ module.exports = (themeOptions = {}) => {
           path: path.resolve(`./src/content`),
         },
       },
+      // Code samples
+      {
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'code-examples',
```
emmenko: We use camelCase notation in the other configs

```suggestion
          name: 'codeExamples',
```

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -94,6 +94,14 @@ module.exports = (themeOptions = {}) => {
           path: path.resolve(`./src/content`),
         },
       },
+      // Code samples
```
emmenko: Nit: for naming consistencies

```suggestion
      // Code examples
```

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -105,6 +113,9 @@ module.exports = (themeOptions = {}) => {
       // For querying configuration data
       'gatsby-transformer-yaml',
 
+      // For querying code samples
```
emmenko: Nit: for naming consistencies

```suggestion
      // For querying code examples
```

packages/gatsby-theme-docs/src/components/code-example.js
emmenko: We can use `endsWith`

```suggestion
    return example.absolutePath.endsWith(props.file);
```

packages/gatsby-theme-docs/src/components/code-example.js
```diff
@@ -0,0 +1,45 @@
...
+      title={props.title}
+      highlightLines={props.highlightLines}
+      noPromptLines={props.noPromptLines}
+    />
+  );
+}
+
+function reportError(errorMsg) {
+  if (process.env.NODE_ENV !== 'production') {
+    return <ContentNotifications.Error>{errorMsg}</ContentNotifications.Error>;
```
emmenko: I don't understand what the purpose of these notification errors is. Can you clarify that?

packages/gatsby-theme-docs/src/components/code-example.js
emmenko: Agree

packages/gatsby-theme-docs/src/components/code-example.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  throw new Error(errorMsg);
+}
+
+CodeExample.propTypes = {
+  file: PropTypes.string.isRequired,
```
emmenko: Isn't `path` a more appropriate name to use? I find `file` a bit confusing, as it's not clear it's a string path.

```
<CodeExample file="foo.js" />

<CodeExample path="foo.js" />
```

packages/gatsby-theme-docs/src/hooks/use-code-examples.js
```diff
@@ -0,0 +1,20 @@
+import { useStaticQuery, graphql } from 'gatsby';
+
+export default () => {
+  const queryResult = useStaticQuery(
+    graphql`
+      query GetCodeExamplesQuery {
```
emmenko:
```suggestion
      query GetAllCodeExamples {
```

packages/gatsby-transformer-code-examples/README.md
```diff
@@ -0,0 +1,56 @@
+# Gatsby Transformer Code Examples
+
+This plugin transforms file contents into CodeExample types in available in graphql.
```
emmenko:
```suggestion
This plugin transforms file contents into `CodeExample` types and makes them available in Gatsby GraphQL schema.
```

packages/gatsby-transformer-code-examples/README.md
```diff
@@ -0,0 +1,56 @@
+# Gatsby Transformer Code Examples
```
emmenko: The package is missing the `LICENSE` file

packages/gatsby-transformer-code-examples/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```js
+// In your gatsby-config.js
+module.exports = {
+  plugins: [
+    {
+      resolve: `gatsby-source-filesystem`,
+      options: {
+        name: `code-examples`,
+        path: `${__dirname}/src/code-examples`,
```
emmenko: The path should resolve the directory of the website, not from the theme.

```suggestion
        path: path.resolve('./src/code-examples'),
```

packages/gatsby-transformer-code-examples/README.md
```diff
@@ -0,0 +1,56 @@
...
+Example `gatsby-config.js` content:
+
+```js
+// In your gatsby-config.js
+module.exports = {
+  plugins: [
+    {
+      resolve: `gatsby-source-filesystem`,
+      options: {
+        name: `code-examples`,
```
emmenko:
```suggestion
        name: `codeExamples`,
```

packages/gatsby-transformer-code-examples/README.md
emmenko: You mean a plugin option? Yeah that makes sense. I guess we can have it as a default `src/code-examples`, then you just use the component as

```
<CodeExample file="foo.js" />
```

which resolves to `src/code-examples/foo.js`.

Is that what you had in mind?

packages/gatsby-transformer-code-examples/README.md
```diff
@@ -0,0 +1,56 @@
...
+    },
+    `@commercetools-docs/gatsby-transformer-code-examples`,
+  ],
+};
+```
+
+## GraphQl Query Example
+
+```graphql
+query GetCodeExamplesQuery {
```
emmenko:
```suggestion
query GetAllCodeExamples {
```

packages/gatsby-transformer-code-examples/README.md
```diff
@@ -0,0 +1,56 @@
...
+      name
+      extension
+      absolutePath
+      content
+    }
+  }
+}
+```
+
+## Supported File Formats
```
emmenko: Shouldn't we list the supported media types instead?

packages/gatsby-transformer-code-examples/gatsby-node.js
```diff
@@ -0,0 +1,74 @@
...
+    'application/x-httpd-php',
+    'application/x-sh',
+    'video/mp2t',
+    'text/yaml',
+  ];
+  if (!validExtensions.includes(node.internal.mediaType)) return;
+
+  const { createNode, createParentChildLink } = actions;
+
+  function transformObject(obj, id, type) {
```
emmenko: What's the purpose of having this function here?

davifantasia: The location does not matter.

packages/gatsby-transformer-code-examples/README.md
davifantasia: I don't see the difference, but if that makes more sense to you, let me know.

packages/gatsby-theme-docs/src/components/code-example.js
davifantasia: It's a similar setup we have in the api-theme - render useful message to author if in development, break in production. This was a feedback from @nkuehn .

emmenko: Alright, thanks

packages/gatsby-theme-docs/src/components/code-example.js
```diff
@@ -0,0 +1,45 @@
...
+
+function CodeExample(props) {
+  const codeExamples = useCodeExamples();
+
+  const codeExample = codeExamples.find(example => {
+    return example.absolutePath.includes(props.file);
+  });
+
+  if (!codeExample) {
+    return reportError('Code example does not exist');
```
emmenko: Let's provide a more useful error message

```suggestion
    return reportError(`Unable to find file "${props.file}" within the "code-examples" folder. Make sure that the file exists and matches the value passed to the "file" property.`);
```

packages/gatsby-transformer-code-examples/gatsby-node.js
```diff
@@ -0,0 +1,74 @@
...
+    {
+      name: node.name,
+      extension: node.extension,
+      absolutePath: node.absolutePath,
+      content,
+    },
+    createNodeId(`${node.id} >>> CodeExample`),
+    'CodeExample'
+  );
+}
```
emmenko:
```suggestion
  const content = await loadNodeContent(node);
  const nodeId = createNodeId(`${node.id} >>> CodeExample`);
  const codeExampleProps = {
    name: node.name,
    extension: node.extension,
    absolutePath: node.absolutePath,
    content,
  };
  const codeExampleNode = {
    ...codeExampleProps,
    id: nodeId,
    children: [],
    parent: node.id,
    internal: {
      contentDigest: createContentDigest(codeExampleProps),
      type: 'CodeExample',
    },
  };
  createNode(codeExampleNode);
  createParentChildLink({ parent: node, child: codeExampleNode });
}
```

packages/gatsby-transformer-code-examples/gatsby-node.js
emmenko: I mean, the function is not necessary in the first place (see suggestion below)

packages/gatsby-transformer-code-examples/gatsby-node.js
davifantasia: I see, for me, a function should ideally not exceed 4 lines, the reason is because in my experience, those kinds of functions are very readable, and they tend to encourage clean code, which I try to practice.

So I would rather do something like

```
async function onCreateNode() {
  if (!validMediaTypes()) return;
  ...
  transformObject();
}
```

emmenko: > for me, a function should ideally not exceed 4 lines

Not sure what you mean here. As long as the function logic is readable, a function can be short or long.
Extracting logic into multiple functions is not always a good thing if not necessary, as it just introduces unnecessary indirection.

Declaring a function closure within the main function, to just use it within the main function, does not help much, it's just unnecessary cognitive complexity.

packages/gatsby-transformer-code-examples/README.md
emmenko: Listing the file extensions gives the impression that only those exact values are supported, which is not the case here as the supported values are based on the mime types.

packages/gatsby-transformer-code-examples/gatsby-node.js
```diff
@@ -0,0 +1,80 @@
...
+
+function isValidMediaType(mediaType) {
+  // some files have no unique media type, see comments below
+  const validMediaTypes = [
```
emmenko: This variable does not need to be declared inside the function, as it's a constant value. Otherwise the variable is initalized every time the function is called.
Instead, just declared it within the file.

davifantasia: What do you mean "within the file"?

emmenko: Declare it as a variable in the `gatsby-node.js`, not within a function.

websites/docs-smoke-test/src/code-examples/example.cs
```diff
@@ -0,0 +1,5 @@
+public class MyClass {
+  public static void main(String[] args) {
+    System.out.println("Hello World");
+  }
+}
```
emmenko: Are you sure this is correct C# code? 🧐

websites/docs-smoke-test/src/code-examples/example.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
emmenko: We should also include an example for `curl`

packages/gatsby-transformer-code-examples/gatsby-node.js
davifantasia: I'm sorry but I disagree.

packages/gatsby-transformer-code-examples/gatsby-node.js
```diff
@@ -0,0 +1,93 @@
...
+};
+
+async function onCreateNode({
+  node,
+  actions,
+  loadNodeContent,
+  createNodeId,
+  createContentDigest,
+}) {
+  if (!isValidNode(node)) return;
```
emmenko: Should we maybe log a warning or something if the node is not valid?

websites/docs-smoke-test/src/code-examples/example.cs
davifantasia: I'm not sure, I found it somewhere online. I don't remember where.

davifantasia: O I see, it's Java.

packages/gatsby-transformer-code-examples/gatsby-node.js
davifantasia: There are many node types used by Gatsby internally or by other plugins that are not valid for this plugin - I don't think we should log those.

emmenko: True, make sense then

websites/docs-smoke-test/src/code-examples/example.php
```diff
@@ -0,0 +1,16 @@
+public class MyClass {
+  public static void main(String[] args) {
+    System.out.println("Hello World");
+  }
+}
```
emmenko: Bad copy-paste?

davifantasia: https://gph.is/2pbhdJ3

davifantasia: (face-palm)

davifantasia: This is strange, this doesn't exist for me locally.

websites/docs-smoke-test/src/code-examples/example.sh
davifantasia: I'm not very familiar with curl. is the file extension ".curl"?

emmenko: I guess... https://filext.com/file-extension/CURL

We will have examples in curl (especially for the API requests), so we need to support this as well.

packages/gatsby-theme-docs/src/components/code-example.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CodeBlock, ContentNotifications } from '@commercetools-docs/ui-kit';
+import useCodeExamples from '../hooks/use-code-examples';
+
+function CodeExample(props) {
+  const codeExamples = useCodeExamples();
```
emmenko: Wondering if it makes more sense to have the hook `useCodeExample` (singular), which returns a single code example:

```js
const codeExample = useCodeExample(props.path)
```

The advantage with this is that the logic to find the code example is encapsulated in the hook (less repetition in case it's used somewhere else).
Just something to consider...

packages/gatsby-transformer-code-examples/README.md
emmenko: Should we look into this as well? @davifantasia @nkuehn

websites/docs-smoke-test/src/code-examples/example.yml
```diff
@@ -0,0 +1,10 @@
+- title: Menu
```
emmenko: The file extension here should be `yaml`. Or you add a mapping `yml -> yaml`

https://github.com/commercetools/commercetools-docs-kit/blob/718ea08ac05a8db7c973fef8a3330c5cf15aea97/packages/ui-kit/src/components/code-block.js#L197-L203

websites/docs-smoke-test/src/code-examples/example.sh
nkuehn: curl is not a common file extension (filext.com is an autogenerated SEO site to pull traffic) but for practical reasons we can agree to allow `*.curl` files, map it to the bash highlighter and call it "cURL" in the UI  (--> multiline PR is related).   That would allow us to selectively choose whether the user sees "cURL" instead of Terminal / Shell / Console which is what the code actually is.

websites/docs-smoke-test/src/code-examples/example.yml
nkuehn: both are common, I vote for a mapping.

packages/gatsby-theme-docs/src/components/code-example.js
nkuehn: I first thought that's good too but then remembered that during the implementation of the API types hook which is the same code design I found that I need use case specific handling e.g. when multiple entries matched or when none matched.  So It's not as obvious and for now I would leave the current state.

packages/gatsby-transformer-code-examples/README.md
nkuehn: it's implemented already now, jsut not reflected yet in the 

for now it's a hardwired convention, if need comes up we can add an option that defaults to the that convention later.

nkuehn:
@emmenko  for me this is good to go, the remaining topic from your feedback is the function structure which I don't see as a blocker.

PR #312, opened 2020-03-18
Samples for the authoring structure of release notes

davifantasia:
This PR proposes samples of release notes structures that will be needed to implement [release notes design](https://commercetools.invisionapp.com/share/ZFWC4K04S7V#/screens).

Approval and feedback needed from key authors like POs and documentation


nkuehn:
@andrea-ct  I optionally included you here for review because you said you intend to write more of the future release notes.  

My format proposal ended up being not extremely different to the current one, but more precisely defined in its metadata and usage.   

This is purely how you write them technically in files and folders and what goes into the file.  It's not (yet) about how you choose the language and what vs. how long you write.

websites/docs-smoke-test/src/release-notes/relase-format-definition.mdx
```diff
@@ -0,0 +1,35 @@
...
+#  - Products
+#  - Product List
+#  - Products
+#  - Product List
+description: |
+  Write Release Notes in the new structured format and see them immediately published in RSS feeds,
+  the Merchant Center and on twitter.com.
+---
+
+The File name can be chosen arbitrarily and subfolders. Only the content of the file has a meaning.
```
andrea-ct:
```suggestion
Choose any file name you want. Only the content of the file has a meaning.
```

... not sure what you meant with the subfolders bit. That it can be in any folder?

websites/docs-smoke-test/src/content/release-notes.mdx
```diff
@@ -0,0 +1,12 @@
+---
+title: Docs Kit Release Notes Test
+# TODO frontmatter to instruct that this is a release note layout? or MDX imports?
```
emmenko: Are release notes supposed to be "integrated" within an existing docs website, like MC?
Or should they "live" in their own website / gatsby project?

In the first case, it would probably make more sense to have a dedicated **folder**, so that the appropriate layout is automatically selected.

In both cases, I'm not sure I would make it configurable via frontmatter (very much error prone).

websites/docs-smoke-test/src/content/release-notes.mdx
```diff
@@ -0,0 +1,12 @@
...
+---
+
+<Subtitle>Custom content above the Release notes is needed for per site customization</Subtitle>
```
emmenko:
```suggestion
<Subtitle>

Custom content above the Release notes is needed for per site customization

</Subtitle>
```

websites/docs-smoke-test/src/content/release-notes.mdx
```diff
@@ -0,0 +1,12 @@
...
+
+To stay informed about the latest releases subscribe to the RSS feed (TODO RSS would be better e.g. in the top right corner?) or sign up for the (test product newsletter)[https://example.com].
+
+<Warn>TODO how to put it here? Can this automatically trigger the filter in the sidebar or would we have to explicitly instruct the layout via frontmatter to not render the index nav??</Warn>
```
emmenko:
```suggestion
<Warn>

TODO how to put it here? Can this automatically trigger the filter in the sidebar or would we have to explicitly instruct the layout via frontmatter to not render the index nav??

</Warn>
```

websites/docs-smoke-test/src/content/release-notes.mdx
```diff
@@ -0,0 +1,12 @@
...
+
+`<ReleaseNotes />`
```
emmenko: Can you explain what this does?

websites/docs-smoke-test/src/release-notes/bugfix.mdx
```diff
@@ -0,0 +1,12 @@
...
+date: 2020-03-16
+type: Resolved Issue
+topics:
+  - Settings
+description: |
+  No more old release notes popping up in external feeds, we found the bug.
+---
+
+ * Authors can change old releases now, for example to fix spelling errors without the release note popping up again in feed consumers
+ * Release Notes have an autogenerated unique ID now.
```
emmenko: Some questions regarding this:
- what are the requirements for this ID?
- where/what is the ID supposed to be used for?

websites/docs-smoke-test/src/release-notes/bugfix.mdx
```diff
@@ -0,0 +1,12 @@
+---
+title: Release notes not reappearing when changed
+date: 2020-03-16
+type: Resolved Issue
```
emmenko: Wondering if we should provide here only "identifiers". The actual "labels" are then mapped based on the identifier.
The reason here is to avoid possible typos and wrong letter casing.

For example: `new`, `resolved`, etc.

websites/docs-smoke-test/src/release-notes/bugfix.mdx
```diff
@@ -0,0 +1,12 @@
...
+topics:
+  - Settings
```
emmenko: Same for the topics. Provide only identifiers that are mapped to pre-defined labels.

websites/docs-smoke-test/src/release-notes/bugfix.mdx
```diff
@@ -0,0 +1,12 @@
+---
```
emmenko: Wondering if it makes sense to have a "draft" flag or something. Probably not, just double checking that.

websites/docs-smoke-test/src/release-notes/relase-format-definition.mdx
```diff
@@ -0,0 +1,35 @@
+---
```
emmenko: The file name has a typo: `relase` -> `release`

websites/docs-smoke-test/src/release-notes/relase-format-definition.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+The File name can be chosen arbitrarily and subfolders. Only the content of the file has a meaning.
+
+The File can use any allowed MDX expressiveness allowed in documentation but it's strongly advised to only use basic markdown and no headings before the "read more" separator.
+
+Metadata is defined in YAML frontmatter.
+
+<!--more-->
+
+The `<!--more-->` syntax for separating out the long form from the part shown in the list is to be taken over from the Jekyll convention. Unprepared new users have understood it intuitively because it's explicit. And the company knows it already.
```
emmenko: I kind of like the concept of the "excerpt" macro from confluence.

If we were to adopt it here, we could have an `<Excerpt>` block which would contain the part of the content that is always visible, while the rest (if any) would be toggled.

```
<Excerpt>

In this release we bla bla...

</Excerpt>

This part is now automatically hidden by default and only visible when clicking on the `read more` button.
```

---

Anyway, maybe I misunderstood the intention here...

websites/docs-smoke-test/src/content/release-notes.mdx
nkuehn: This was intended to trigger a conversation with you how we want authors to tell the implementation that this page is a release notes page. 

So it worked :-) - let's talk about it directly.

websites/docs-smoke-test/src/release-notes/bugfix.mdx
nkuehn: WIP, this is basically lorem ipsum.   Can't revert a PR back to draft status,  @davifantasia  created it as an open PR directly.

websites/docs-smoke-test/src/release-notes/bugfix.mdx
nkuehn: good idea for the types since they are intended to be finite and predefined, thanks !

websites/docs-smoke-test/src/release-notes/bugfix.mdx
nkuehn: We came to the conclusion to not have predefined topics but make it dynamic, i.e. the UI is populated from the topics that are actually used.   Basically like most filter engines work - most painless for the authors to add more. 

We will only need separate configuration once the need comes up to hide old topics in the filter nav.  That would need to be a per-website data file defining the filterable topics, but even that would not be a hard whitelist (you never want to touch old release notes from five years ago).

websites/docs-smoke-test/src/release-notes/bugfix.mdx
nkuehn: Jekyll has  both time-based and explicit via "draft" frontmatter logic and we went lengths to deactivate both in the current site because it interferes in ugly ways with a git branch/merge based publishing workflow.  
So for the `commercetools-docs` It's a feature to not have a draft flag. Unpublished drafts are a git branch together with the matching documentation changes. 

This may be different in repos like the application kit - there branching is decoupled from documentation changes at least to some degree and I can imagine you want to avoid long-open branches there?

websites/docs-smoke-test/src/release-notes/relase-format-definition.mdx
nkuehn: Yes, I think it's similar to confluence, you tell the renderer which part of the document is to be shown in the list view of entries as a shorter version and which is only shown if the entry is rendered as an own whole page. 

I'm not a fan though of making the marker a Block wrapper because 
1.  that doesn't guarantee that the excerpt is always the beginning of the document.  
2. the majority of our release notes does not have a long form behind "Read More" so that would mean that all those would have to be seemingly pointlessly wrapped into an `<Excerpt>` block just to render at all in the list view.  Sounds error prone to me. 

PS:  It also overlaps a bit with the intention of a `description` frontmatter, but the latter has harsher restrictions in length and therefore needs to be worded differently.  

PPS:  GatsbyJS has built-in  "excerpt" logic on its GraphQL surface but the author can't control where the excerpt ends. It's just making it a plain text and cutting off after a certain number of words (or characters, not sure).

websites/docs-smoke-test/src/release-notes/relase-format-definition.mdx
nkuehn: Yes, I suggest to allow any folder structure inside `release-notes` so you can e.g. later archive stuff into `release-notes/2019-Q1` or whatever structure the website maintainer likes without affecting how the site renders.   

The current file list is pretty long and starts to get annoying.

websites/docs-smoke-test/src/content/release-notes.mdx
nkuehn: **Separate website or not?**
I checked with stakeholders like POs (and I think even you) and the conclusion was to stick with separate release notes lists per part of our product (i.e. currently MC, API, App Kit, SDKs, ...) instead of a central aggregated one.  We can nevertheless choose to have a special website that contains nothing but certain release notes if we want. 

**Why the content folder?**
We chose to make the URL namespace inside a website 1:1 the folder structure in `src/content` so there is no conflict free namespace or "folder" outside that physical folder.   That's why sticking to that and having at least a placeholder file for the release notes URL felt better to me. 

The text, heading etc. above the release notes list is custom per e.g. MC vs. API vs. others and I like the Idea that it can be easily modified and authored by docs and PO staff like any other page.  That's how bigger component based CMS systems work, too.  Everything is a page, but some consist of just one big component.  

**Frontmatter // how to control the page layout? *** 
I don't really like it, too - just mentioned it because the that's how Jekyll is doing it right now - you put a "layout" frontmatter to choose the page template.  

How do you imagine to author home pages like https://docs.commercetools.com/custom-applications/ in the future?  It's a similar case of a certain page URL having a differing layout.  Apropos, It would be great if that can be MDX, too :-)

websites/docs-smoke-test/src/release-notes/bugfix.mdx
nkuehn: how about `feature` , `fix` and `enhancement`??

websites/docs-smoke-test/src/release-notes/bugfix.mdx
emmenko: Sure, my point is that in the front matter you only provide the identifiers, to avoid typos etc.
Each website can then define its own list of topics as a config map.

websites/docs-smoke-test/src/release-notes/bugfix.mdx
emmenko: Yeah I thought so. Thanks

websites/docs-smoke-test/src/release-notes/relase-format-definition.mdx
emmenko: Looks like we can simply use Gatsby for this: https://using-remark.gatsbyjs.org/excerpts/

Probably that’s what you had in mind in the first place 🙂

nkuehn: Good!  I was obviously too lazy to read the whole page.   I thing we can combine both the separator and a (high) `pruneLength`  to get a robust tool.

websites/docs-smoke-test/src/release-notes/bugfix.mdx
nkuehn: I'll ad an optional "published" key to the spec, but not the template now which can be used for both preventing publication if you have to merge to master for some reason and also to be able to hide old release notes that have to be removed for some reason.   "draft" would not be intuitive for the latter case.

websites/docs-smoke-test/src/release-notes/relase-format-definition.mdx
```diff
@@ -0,0 +1,34 @@
...
+  - FooBar
+#  - Baz
+#  - Foo
+---
+
+The file can have any name as long as it resides in `src/release-notes`. Only the content of the file controls the website content. 
+
+The file can live in folders to better organize large numbers of past release notes. 
+
+The File can use any allowed MDX allowed in documentation but it's strongly advised to only use basic markdown and no headings before the "read more" separator.
```
adinakleine:
```suggestion
The file can use any MDX allowed in documentation, but it's strongly advised to only use basic markdown and no headings before the "read more" separator.
```

websites/docs-smoke-test/src/release-notes/relase-format-definition.mdx
```diff
@@ -0,0 +1,34 @@
+---
+date: 2000-03-03
+title: Release Note Format Definition
+description: |
+  The description must be a stand-alone text conveying the key change and its advantages for external feed consumers, 
+  redistribution, notifications an others.
```
adinakleine:
```suggestion
  redistribution, notifications and others. 
```

PR #323, opened 2020-03-25
feat(writing-style): upgrade to vale 2.1.0, don't let vale fail when called from the VSCode plugin

nkuehn:
the VSCode vale plugin passes an empty `--config` flag in some situations (not sure which but I realized that in our old docs repo).  That leads to two config params being passed and the last one being empty.  

This change detects this specific pattern, it's not intending to be a general solution for all sorts of misconfigurations in the params passed - only supposed to fix this special case because we are setting and own --config parameter.  

In addition, this PR upgrades vale to the new 2.x version which requires some refactoring of the downloader  and the vale configuration:
 * the 7za based extraction on unixes fails on the new artifacts (I have no idea why), so this uses the operating system tar on unices
 * the new vale has some new built-in styles to be configured and requires the "write-good" style to be held locally now, so it's also persisted here. 
 * to make sure the vale binary is actually upgraded, it's now saved with the version in the file name


packages/writing-style/bin/commercetools-vale.js
```diff
@@ -8,7 +8,16 @@ const binaryFileName = os.platform() === 'win32' ? 'vale.exe' : 'vale';
 
 const valeBinary = path.join(__dirname, `./${binaryFileName}`);
 const valeConfig = path.join(__dirname, `../.vale.ini`);
-const result = shelljs.exec(
-  `${valeBinary} --config ${valeConfig} ${process.argv.slice(2).join(' ')}`
```
emmenko: Out of curiosity: what happens if we pass the `--config` as last parameter? Will that be used even if there are duplicated params before that?

```
`${valeBinary} ${process.argv.slice(2).join(' ')} --config ${valeConfig}`
```

nkuehn: It doesn't catch that case but as said the implementation is specifically targeting the situation with the VSCode plugin which passes the --format parameter after an empty --config.    

passing --config as last or second last does not work in any case (also directly against the vale binary).  You need to pass a config file after it and then the directory to lint at the end.

packages/writing-style/bin/commercetools-vale.js
```diff
@@ -8,7 +8,16 @@ const binaryFileName = os.platform() === 'win32' ? 'vale.exe' : 'vale';
...
-);
+let params = process.argv.slice(2);
+// simplistic check whether the --config param was passed empty by the VSCode plugin (empty = followed by another flag)
+const configParamPosition = params.findIndex((p) => p === '--config');
+if (
+  params[configParamPosition + 1] &&
+  params[configParamPosition + 1].startsWith('-')
```
emmenko: I don't fully understand what this check is for, if you then just do a filter.

nkuehn: this is checking whether the parameter _after_ the `--config` is not a value but another config flag like `-v` or `--format` etc. .    The config should be filtered out only if it has no value (you seem to be able to pass it twice the program takes the last one, so it's OK to pass a config to our wrapper to override our config.

emmenko: Ah ok, got it now

PR #330, opened 2020-03-30
fix: query parameters types that are objects now render as links

### LINKED ISSUE #322: Types of properties of query parameters should be rendered as link
Current situation on category import when import api is run locally:

Screenshot 2020-03-25 at 16 44 56

`ProcessingState` should be rendered as a link.

davifantasia:
Fixes #322

### DIFF
--- a/packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
+++ b/packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
@@ -3,8 +3,7 @@ import PropTypes from 'prop-types';
 import SpacingsStack from '@commercetools-uikit/spacings-stack';
 import { Markdown } from '@commercetools-docs/ui-kit';
 import { useTypeLocations } from '../../../../hooks/use-type-locations';
-import capitalizeFirst from '../../../../utils/capitalize-first';
-import renderTypeAsLink from '../../../../utils/render-type-as-link';
+import generateTypeToRender from '../../../../utils/generate-type-to-render';
 import Required from '../../../required';
 
 const NameType = ({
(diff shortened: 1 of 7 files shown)


packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -13,17 +12,11 @@ const NameType = ({
...
 }) => {
-  const typeLocations = useTypeLocations();
-  const dataForTypeToLinkTo =
-    property.type === 'array' && property.items
-      ? { type: property.items.type, displayPrefix: 'Array of ' }
-      : { type: property.type, displayPrefix: '' };
-
-  const type = renderTypeAsLink(
+  const typeToRender = generateTypeToRender({
+    typeLocations: useTypeLocations(),
```
emmenko: Please keep usage of react hooks within the body function, otherwise it's confusing and might lead to unnecessary issues (for example render a hook within a condition).

```suggestion
  const typeLocations = useTypeLocations();
  const typeToRender = generateTypeToRender({
    typeLocations,
```

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -39,8 +32,15 @@ const NameType = ({
...
-        <span className="name">{dataForTypeToLinkTo.displayPrefix}</span>
-        {typeof type === 'string' ? <span className="name">{type}</span> : type}
+        {typeToRender.displayPrefix && (
+          <span className="name">{typeToRender.displayPrefix}</span>
+        )}
+
+        {typeof typeToRender.type === 'string' ? (
+          <span className="name">{typeToRender.type}</span>
+        ) : (
+          typeToRender.type
```
emmenko: I'm confused. If `typeToRender.type` is not a string, what can it be?

davifantasia: A `<Link/>`

emmenko: I see...kind of confusing by whatever

PR #332, opened 2020-03-31
feat(theme): set up release notes template

davifantasia:
This PR includes

- site header
- rendering list of release notes

This PR does not include

- release notes [design](https://commercetools.invisionapp.com/share/ZFWC4K04S7V#/screens) 
- filtering capability

UPDATE:

After discussing with @emmenko , this PR will include creating each release note details page.


packages/gatsby-theme-docs/src/pages/release-notes.js
```diff
@@ -0,0 +1,116 @@
...
+  SEO,
+  Link,
+  ThemeProvider,
+  CodeExample,
+  MultiCodeExample,
+} from '../components';
+import PlaceholderMarkdownComponents from '../overrides/markdown-components';
+
+// See https://mdxjs.com/getting-started#table-of-components
+const components = {
```
emmenko: This shouldn't be duplicated. We should extract this and use it in both templates

packages/gatsby-theme-docs/src/pages/release-notes.js
```diff
@@ -0,0 +1,116 @@
+import React from 'react';
```
emmenko: image

This should be a `template`, not a page

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -105,6 +105,16 @@ exports.onCreateNode = ({ node, getNode, actions }, pluginOptions) => {
...
     });
+    // Release note flag, used for filter in release notes query
+    actions.createNodeField({
+      node,
+      name: 'isReleaseNote',
+      value: node.fileAbsolutePath.startsWith(
+        path.resolve(`./src/release-notes`)
+      )
+        ? Boolean(true)
+        : Boolean(false),
```
emmenko: What? 🙃

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -105,6 +105,16 @@ exports.onCreateNode = ({ node, getNode, actions }, pluginOptions) => {
         Boolean(node.frontmatter.excludeFromSearchIndex) ||
         Boolean(pluginOptions.excludeFromSearchIndex),
     });
+    // Release note flag, used for filter in release notes query
+    actions.createNodeField({
+      node,
+      name: 'isReleaseNote',
```
emmenko: Why do we need this flag?

packages/gatsby-theme-docs/src/pages/release-notes.js
davifantasia: Thanks, but please note that the PR is still a draft.

packages/gatsby-theme-docs/gatsby-node.js
davifantasia: I wrote why it is needed in the comment but to clarify:

- the release notes overview needs list of release notes, so this field enables filtering out only release notes when the query is executed.

packages/gatsby-theme-docs/src/pages/release-notes.js
davifantasia: I thought about that, and this appears to be a better/simpler approach than using a template.

packages/gatsby-theme-docs/gatsby-node.js
emmenko: But why. You can use the graphql filters and for example filter by the file path or the name you gave to the source-filesystem.

I don't think we need to add a new field just for filtering.

packages/gatsby-theme-docs/gatsby-node.js
davifantasia: I could also write

```
Boolean(node.fileAbsolutePath.startsWith(path.resolve(`./src/release-notes`) )
```

Do you prefer this?

packages/gatsby-theme-docs/src/pages/release-notes.js
emmenko: Why, we already discussed this. It should work the same was as for the content pages. Why does it need to be different?

packages/gatsby-theme-docs/gatsby-node.js
emmenko: Just to be clear: `startsWith` returns a boolean!

packages/gatsby-theme-docs/gatsby-node.js
davifantasia: I thought about hose options but I could not find an implementation that works.

packages/gatsby-theme-docs/src/pages/release-notes.js
davifantasia: Using template now.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -111,18 +111,32 @@ exports.onCreateNode = ({ node, getNode, actions }, pluginOptions) => {
...
   const allMdxPagesResult = await graphql(`
+    fragment fieldsFragment on Mdx {
+      id
+      fields {
+        slug
+        title
+        beta
+        isGlobalBeta
+        excludeFromSearchIndex
+      }
```
emmenko, line 122: PS: I don't think the content pages and the release note pages have the same frontmatter fields (for example the beta flags).
Therefore, I would not use a fragment and explicitly list the fields on each query.

Maybe @nkuehn can give some more input?

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -111,18 +111,32 @@ exports.onCreateNode = ({ node, getNode, actions }, pluginOptions) => {
...
-            id
-            fields {
-              slug
-              title
-              beta
-              isGlobalBeta
-              excludeFromSearchIndex
-            }
+      releaseNotes: allFile(
+        filter: { sourceInstanceName: { eq: "releaseNotes" } }
```
emmenko, line 127: Thanks 💯

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -147,31 +161,58 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+        },
+      });
+    }
+  });
+
+  // create release notes overview page
+  actions.createPage({
+    path: '/release-notes',
+    component: require.resolve('./src/templates/release-notes.js'),
+  });
```
emmenko, line 198: This is not necessary, we can use an `index.mdx` in the `release-notes` folder.
Then in the `forEach` below you can use the "list overview" template for the `index.mdx` page, and the "detail page" template for all other pages.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -147,31 +161,58 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
-      // our page layout component
-      context: {
-        ...node.fields,
-        shortTitle: matchingNavigationPage
-          ? matchingNavigationPage.title
-          : undefined,
-      },
-    });
+  pages.forEach(({ childMdx }) => {
+    if (childMdx) {
```
emmenko, line 170: Why do we need this?

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -147,31 +161,58 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+    component: require.resolve('./src/templates/release-notes.js'),
+  });
+
+  // create details pages for each release note
+  allMdxPagesResult.data.releaseNotes.nodes.forEach(({ childMdx }) => {
+    if (childMdx) {
+      actions.createPage({
+        // This is the slug you created before
+        // (or `node.frontmatter.slug`)
+        path: childMdx.fields.slug,
```
emmenko, line 206: @nkuehn was mentioning something about the URLs of the release notes to be different than the content pages...not sure what the expected outcome here is.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -147,31 +161,58 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
-      // You can use the values in this context in
-      // our page layout component
-      context: {
-        ...node.fields,
-        shortTitle: matchingNavigationPage
-          ? matchingNavigationPage.title
-          : undefined,
-      },
-    });
+  pages.forEach(({ childMdx }) => {
```
emmenko, line 169: PS: we should throw an error if one of the content pages is named `release-notes.mdx`

packages/gatsby-theme-docs/src/templates/release-note.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { graphql } from 'gatsby';
+
+const ReleaseNote = (props) => {
```
emmenko, line 5: All template pages should use the same "setup", meaning the different providers, etc.

packages/gatsby-theme-docs/gatsby-node.js
emmenko: At least, from what I understood, in the `index.mdx` file you should be able to still write some intro content and then get the list of the release notes **below** that.

So the template for the release notes should query both the single page and all other pages.
The single page content should be rendered first, then the list of release notes.

packages/gatsby-theme-docs/gatsby-node.js
davifantasia: There was a `null` case.

packages/gatsby-theme-docs/gatsby-node.js
nkuehn: 1. yes, the meaning and choice of allowed frontmatter are completely different and I would be happy if the query prevents using e.g. a release notes frontmatter (like the "slug" which is there only for migration) in regular pages. 
2. most of the frontmatter defined in the release note spec example is not queried here yet.

packages/gatsby-theme-docs/src/templates/releases.js
```diff
@@ -0,0 +1,150 @@
...
+  em: Markdown.Em,
+  strong: Markdown.Strong,
+  delete: Markdown.Delete,
+  hr: Markdown.Hr,
+  pre: Markdown.CodeBlock,
+  // Custom component specific to the Gatsby theme
+  a: Link,
+
+  // Official react components to be used in MDX files
+  Subtitle,
```
nkuehn, line 57: Subtitle should not be allowed in release notes  (Assuming this template is for the release note details page - on the top of the overview page we do need it though). 

If we omit it here using it in a release note will either break or just render a div (not sure, but both would be good)

packages/gatsby-theme-docs/src/templates/releases.js
```diff
@@ -0,0 +1,150 @@
...
+  p: Markdown.Paragraph,
+  // NOTE: we want to ensure that only one h1 exists on each page.
+  // Therefore, we map the markdown header elements starting from h2.
+  // The h1 header will be automatically rendered based on the page title.
+  h1: Markdown.withAnchorLink(Markdown.H2),
+  h2: Markdown.withAnchorLink(Markdown.H3),
+  h3: Markdown.withAnchorLink(Markdown.H4),
+  h4: Markdown.withAnchorLink(Markdown.H5),
+  h5: Markdown.withAnchorLink(Markdown.H6),
+  h6: Markdown.withAnchorLink(Markdown.H6),
```
nkuehn, line 33: Same here, in my opinion a release note should not have a heading structure in itself (Assuming this template is for the release note details page)

We can consider allowing one level to give a longer RN some structure (that would be H1 in MDX and sven would have to decide how to render it).  The rest should IMHO just render as paragraphs.

packages/gatsby-theme-docs/src/templates/releases.js
```diff
@@ -0,0 +1,150 @@
...
+const ReleaseNote = (props) => (
+  <ThemeProvider>
+    <LayoutContent
+      pageContext={props.pageContext}
+      pageData={props.data.releasesIndex}
+    >
+      <MDXProvider components={components}>
+        <Markdown.TypographyPage>
+          <SEO
+            title={props.pageContext.shortTitle || props.pageContext.title}
```
nkuehn, line 78: If this is for the release notes details page: There is no shortTitle frontmatter or metadata possible for a release note.  

On the overview page it's possible though.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -147,34 +175,64 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
     });
   });
+
+  createAllReleaseNotesPages(
+    actions,
+    allMdxPagesResult.data.releaseNotes.nodes
+  );
 };
 
+function createAllReleaseNotesPages(actions, nodes) {
```
emmenko, line 219: Sorry but this function is really unnecessary, and so are the arguments passed, since it’s defined as a closure.
Please keep it simple and render the `forEach` only.

packages/gatsby-theme-docs/src/templates/releases.js
```diff
@@ -0,0 +1,150 @@
...
+  MultiCodeExample,
+
+  // Custom React components that can be injected from each website
+  // See ../overrides/README.md
+  ...PlaceholderMarkdownComponents,
+};
+
+const ReleaseNote = (props) => (
+  <ThemeProvider>
+    <LayoutContent
```
emmenko, line 71: This should be `LayoutReleaseNotes` which renders the appropriate layout for the release notes.

packages/gatsby-theme-docs/src/templates/releases.js
```diff
@@ -0,0 +1,150 @@
...
+
+export default ReleaseNote;
+
+export const query = graphql`
+  query QueryReleasesPage($slug: String!) {
+    releasesIndex: mdx(fields: { slug: { eq: $slug } }) {
+      body
+    }
+
+    releaseNotes: allFile(filter: { sourceInstanceName: { eq: "releases" } }) {
```
emmenko, line 139: You need to filter out the index page and non mdx files.

However, as suggested in the call, we could just inject this list as a context field since we already have all the data we need in gatsby-node.

packages/gatsby-theme-docs/src/templates/releases.js
```diff
@@ -0,0 +1,150 @@
...
+            })}
+          </div>
+        </Markdown.TypographyPage>
+      </MDXProvider>
+    </LayoutContent>
+  </ThemeProvider>
+);
+
+ReleaseNote.propTypes = {
+  pageContext: PropTypes.shape({
```
emmenko, line 104: As @nkuehn mentioned, release note files have different frontmatter fields

packages/gatsby-theme-docs/src/templates/releases.js
```diff
@@ -0,0 +1,150 @@
...
+          <SEO
+            title={props.pageContext.shortTitle || props.pageContext.title}
+            excludeFromSearchIndex={props.pageContext.excludeFromSearchIndex}
+          />
+          <div>
+            <MDXRenderer>{props.data.releasesIndex.body}</MDXRenderer>
+          </div>
+          <div>
+            {props.data.releaseNotes.nodes.map((node, index) => {
+              return (
```
emmenko, line 86: We should move this part into a component like `<ReleaseNote>`

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -147,34 +175,64 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+  nodes.forEach(({ childMdx, name }) => {
+    const template =
+      name === 'index'
+        ? './src/templates/releases.js'
+        : './src/templates/page-content.js';
```
emmenko, line 224: Why is this using the page content layout?

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -97,6 +97,14 @@ module.exports = (themeOptions = {}) => {
           path: path.resolve(`./src/code-examples`),
         },
       },
+      // Release notes
+      {
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'releases',
+          path: path.resolve(`./src/releases`),
```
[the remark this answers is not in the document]
nkuehn: I changed it because I realized that it's confusing if the URL and the local paths use different conventions.

nkuehn:
@davifantasia  @emmenko  thanks for the hint and sorry for commenting on layout / design specific topics.  Just ignore these comments or mark them resolved offtopic.

packages/gatsby-theme-docs/gatsby-node.js
emmenko: I got tricked by the diff in the phone sorry, I see now the function is moved outside.
Still, I don't see a reason for extracting this part into a function as we have all the other logic of the `createPages` function in one place.

emmenko:
@davifantasia I opened a PR (based on this branch) to keep the minimum setup that we need as the first step. I thought it will be easier to do that instead of keep commenting back and forth, so that you can see what I meant with my comments.

I left some TODOs that should/can be done in follow up PRs, for example regarding the layout, the markdown elements, etc (what @nkuehn was also referring to). Those need some clarification and are not necessary in this PR.

Let me know if you have further questions, thanks.

emmenko:
FYI: once we merge the other PR here, this would be good to go from my side (as the first step).

PR #334, opened 2020-03-31
chore: update cookie consent banner script

nkuehn:
We received a new cookie consent banner ID for the docs kit, here's my try to implement it correctly. 

The raw script copy/paste we received form Marketing is:
```html
<!-- OneTrust Cookies Consent Notice start -->
<script type="text/javascript" src="https://cdn.cookielaw.org/consent/b104027d-4d10-4b75-9675-9ffef11562a8/OtAutoBlock.js"></script>
<script type="text/javascript" src="https://cdn.cookielaw.org/scripttemplates/otSDKStub.js" type="text/javascript" charset="UTF-8" data-domain-script="b104027d-4d10-4b75-9675-9ffef11562a8"></script>
<script type="text/javascript">
  function OptanonWrapper() {}
</script>
<!-- OneTrust Cookies Consent Notice end -->


nkuehn:
@emmenko  thanks for spotting that you can pass a list (which is obvious by the name of the function, but I just did not take enough car). 

Concerning deferring the execution I am not very inclined to restart the debug cycle with that horrible script.  I am pretty sure that I already tried deferring the execution and it did not work or caused weird things like two complete cookie consent instances laid over each other, rendering the page overlay as black. 

Another reason why I'm hesitant is that we can't really test it as easily without actually running it on a commercetools.com domain in production.

emmenko:
There should be no issue with `defer` as far as I can tell. The MDN doc says:

> This Boolean attribute is set to indicate to a browser that the script is meant to be executed after the document has been parsed, but before firing DOMContentLoaded.
> Scripts with the defer attribute will prevent the DOMContentLoaded event from firing until the script has loaded and finished evaluating.

The purpose of the attribute is just to eliminate the blocking of the parse phase of the browser. However, the script will still be loaded and executed before the page is officially loaded.

Anyway, if you still prefer to keep it out I can change that.

nkuehn:
@emmenko  sorry, clicked "update from master" which is not a rebase.  

Let's try defer, by spec it should work - we just need to be very fast in rolling a complete kit release cycle when we find out it's blocking the complete custom apps docs after go-live on a commercetools.com domain.

emmenko:
I can try this in the custom apps docs first

PR #335, opened 2020-04-02
refactor(theme): to keep the minimum setup for release notes

emmenko:
Base branch #332

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-node.js
+++ b/packages/gatsby-theme-docs/gatsby-node.js
@@ -211,7 +211,11 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
     });
   });
 
+  const isOverviewPage = (name) => name === 'index';
   const releaseNotePages = allMdxPagesResult.data.releaseNotes.nodes;
+  const releaseNotePagesWithoutIndex = allMdxPagesResult.data.releaseNotes.nodes.filter(
+    (node) => !isOverviewPage(node.name)
+  );
   releaseNotePages.forEach(({ childMdx, name }) => {
     actions.createPage({
       // TODO: how should the path be named exactly?
@@ -219,8 +223,9 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
       component: require.resolve('./src/templates/releases.js'),
       context: {
         ...childMdx.fields,
-        isOverviewPage: name === 'index',
-        releaseNotePages,
+        ...(isOverviewPage(name)
+          ? { releaseNotes: releaseNotePagesWithoutIndex }
+          : {}),
       },
     });
   });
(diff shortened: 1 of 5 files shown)


emmenko:
This is how the overview page looks like (note there is not layout implemented yet, just the data that is rendered).

image

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -68,30 +69,43 @@ exports.sourceNodes = ({ actions }) => {
...
+    name: 'excludeFromSearchIndex',
+    value:
+      Boolean(node.frontmatter.excludeFromSearchIndex) ||
+      Boolean(pluginOptions.excludeFromSearchIndex),
+  });
+
+  const isContentPage = node.fileAbsolutePath.startsWith(
+    path.resolve('src/content')
+  );
+  if (isContentPage) {
```
emmenko, line 101: Here we create nodes that are specific to content pages

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -102,54 +116,60 @@ exports.onCreateNode = ({ node, getNode, actions }, pluginOptions) => {
       name: 'isGlobalBeta',
       value: Boolean(pluginOptions.beta),
     });
+  }
+
+  const isReleaseNotesPage = node.fileAbsolutePath.startsWith(
+    path.resolve('src/releases')
+  );
+  if (isReleaseNotesPage) {
```
emmenko, line 124: Here we create nodes that are specific to release notes pages

packages/gatsby-theme-docs/src/templates/releases.js
```diff
@@ -3,147 +3,93 @@ import PropTypes from 'prop-types';
...
     }
-
-    releaseNotes: allFile(filter: { sourceInstanceName: { eq: "releases" } }) {
+    allReleaseNotes: allFile(
+      filter: {
+        sourceInstanceName: { eq: "releases" }
+        internal: { mediaType: { eq: "text/mdx" } }
+        name: { ne: "index" }
+      }
+    ) @include(if: $isOverviewPage) {
```
emmenko, line 81: I added this so that this query is executed only when the page is the overview page.

packages/gatsby-theme-docs/src/templates/releases.js
```diff
@@ -3,147 +3,93 @@ import PropTypes from 'prop-types';
...
+    mdx(fields: { slug: { eq: $slug } }) {
       body
     }
-
-    releaseNotes: allFile(filter: { sourceInstanceName: { eq: "releases" } }) {
+    allReleaseNotes: allFile(
+      filter: {
+        sourceInstanceName: { eq: "releases" }
+        internal: { mediaType: { eq: "text/mdx" } }
+        name: { ne: "index" }
```
emmenko, line 79: We also need to filter for all pages except for the index page.

PR #336, opened 2020-04-02
feat(theme): implement release notes page layout

emmenko:
This PR implements the page layout structure for the release note pages. The layout is the same as for the content pages, except for the absence of the site navigation and the page filters (to be done separately).

**This does not implement the designs of the release notes, only the layout**.


emmenko:
I removed the footer now.

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -19,6 +19,7 @@ const defaultOptions = {
   beta: false,
   gaTrackingId: undefined,
   excludeFromSearchIndex: true,
+  hasReleaseNotes: false,
```
emmenko, line 22: The link to the release notes is toggled by this config flag.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -233,10 +238,11 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
     actions.createPage({
       // TODO: how should the path be named exactly?
       path: childMdx.fields.slug,
-      component: require.resolve('./src/templates/releases.js'),
+      component: isOverviewPage
+        ? require.resolve('./src/templates/release-notes-list.js')
+        : require.resolve('./src/templates/release-notes-detail.js'),
```
emmenko, line 243: After some back and forth I think it's better to just have two separate templates, it makes things easier to implement and separates a bit the concerns, as the pages are indeed different.

packages/gatsby-theme-docs/src/components/release-notes-subscribe-links.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const ReleaseNotesSubscribeLinks = () => <div>{'TODO subscribe links'}</div>;
```
emmenko, line 3: Here we will add the "subscribe" links (rss feed, etc.). To be done in a separate PR.

packages/gatsby-theme-docs/src/hooks/use-layout-state.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+
+const useLayoutState = () => {
```
emmenko, line 3: I moved all this state into a hook, to make is easier to reuse across the different layout components.

packages/gatsby-theme-docs/src/layouts/internals/layout-header-logo.js
```diff
@@ -0,0 +1,28 @@
...
+  @media screen and (${designSystem.dimensions.viewports.desktop}) {
+    width: ${designSystem.dimensions.widths.pageNavigation};
+  }
+`;
+
+const LayoutHeaderLogo = () => (
+  <Container>
+    <LogoButton />
+  </Container>
+);
```
emmenko, line 26: Just extracted for better reusability

packages/gatsby-theme-docs/src/layouts/internals/layout-page-release-notes-filters.js
```diff
@@ -0,0 +1,34 @@
...
+`;
+const StickyContainer = styled.div`
+  position: sticky;
+  top: ${designSystem.dimensions.spacings.xxl};
+  margin: 0 0 ${designSystem.dimensions.spacings.s};
+`;
+
+const LayoutPageReleaseNotesFilters = () => (
+  <GridContainer>
+    <StickyContainer>{'TODO: filters'}</StickyContainer>
```
emmenko, line 29: Here is where the filters will be implemented. To be done in a separate PR.

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -66,8 +60,77 @@ const LinkItem = styled.div`
...
+  }
+
+  > * + * {
+    margin: 0 0 0 ${designSystem.dimensions.spacings.s};
+  }
+`;
+const activeLinkStyles = `
+  border-left: ${designSystem.dimensions.spacings.xs} solid ${designSystem.colors.light.linkNavigation} !important;
+  color: ${designSystem.colors.light.linkNavigation} !important;
+`;
```
emmenko, line 83: Extracted into variables for better readability

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -66,8 +60,77 @@ const LinkItem = styled.div`
...
 
 const SidebarLink = (props) => {
+  const {
+    locationPath,
+    customStyles,
+    customActiveStyles,
```
emmenko, line 89: These two props are new, to allow to override some of the link styles.

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -66,8 +60,77 @@ const LinkItem = styled.div`
...
+  );
+};
+SidebarLink.propTypes = {
+  to: PropTypes.string.isRequired,
+  locationPath: PropTypes.string,
+  customStyles: PropTypes.string,
+  customActiveStyles: PropTypes.string,
+};
+
+const SidebarLinkWrapper = (props) => {
```
emmenko, line 133: I split the `SidebarLink` component into two, with the wrapper containing the logic for the scrolling position.

The base link component is use for the release notes link.

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -226,9 +246,16 @@ SidebarChapter.propTypes = {
...
+  location: PropTypes.shape({
+    state: PropTypes.shape({
+      sidebarScrollPosition: PropTypes.number,
+    }),
+    pathname: PropTypes.string.isRequired,
+  }).isRequired,
 };
 
-const Sidebar = (props) => {
+const SidebarNavigationLinks = (props) => {
```
emmenko, line 258: This component is also extracted and contains the normal navigation links.

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -273,18 +332,56 @@ const Sidebar = (props) => {
...
-            isGlobalBeta={props.isGlobalBeta}
-            onLinkClick={props.onLinkClick}
-            nextScrollPosition={nextScrollPosition}
-          />
-        ))}
-      </ScrollContainer>
+      {props.hasReleaseNotes && (
+        <ReleaseNotesTitle>
+          <SidebarLink
+            to="/releases"
```
emmenko, line 338: This is the link to the release notes.

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -273,18 +332,56 @@ const Sidebar = (props) => {
...
+              <LinkSubtitle>{'Relase notes'}</LinkSubtitle>
+              <ReleaseNotesIcon
+                size="medium"
+                color={isReleasePage ? 'linkNavigation' : 'link'}
+              />
+            </SpacingsInline>
+          </SidebarLink>
+        </ReleaseNotesTitle>
+      )}
+      {props.hasReleaseNotes && isReleasePage ? (
```
emmenko, line 367: If this condition is true, we render the "back" link

packages/gatsby-theme-docs/src/layouts/release-notes-detail.js
```diff
@@ -0,0 +1,86 @@
...
+              <Link href="/releases" noUnderline={true}>
+                <span
+                  css={css`
+                    font-size: ${designSystem.typography.fontSizes.small};
+                  `}
+                >
+                  {'Back to all releases'}
+                </span>
+              </Link>
+            </SpacingsInline>
```
emmenko, line 55: The detail page has this "back" link at the top

packages/gatsby-theme-docs/src/templates/release-notes-detail.js
```diff
@@ -0,0 +1,62 @@
...
+import { SEO, ThemeProvider } from '../components';
+import markdownComponents from '../markdown-components';
+
+const releaseNoteMarkdownComponents = {
+  ...markdownComponents,
+  // NOTE: release notes content can only have headings starting from h4.
+  h1: Markdown.withAnchorLink(Markdown.H4),
+  h2: Markdown.withAnchorLink(Markdown.H5),
+  h3: Markdown.withAnchorLink(Markdown.H6),
+  h4: Markdown.withAnchorLink(Markdown.H6),
```
emmenko, line 17: This is the mapped list of headings for the release notes content.

cc @nkuehn

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -0,0 +1,111 @@
...
+      <Markdown.TypographyPage>
+        <SEO
+          title={props.pageContext.title}
+          excludeFromSearchIndex={props.pageContext.excludeFromSearchIndex}
+        />
+        <MDXProvider components={markdownComponents}>
+          <div>
+            <MDXRenderer>{props.data.mdx.body}</MDXRenderer>
+          </div>
+        </MDXProvider>
```
emmenko, line 34: Note that here we use the normal page content markdown elements, which is used by the `releases/index.mdx` file content.

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -0,0 +1,111 @@
...
+        <div>
+          <MDXProvider components={releaseNoteMarkdownComponents}>
```
emmenko, line 36: And for the list of release notes content we use the modified list of markdown elements.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -14,15 +14,15 @@ const trimTrailingSlash = (url) => url.replace(/(\/?)$/, '');
...
+exports.onPreBootstrap = ({ reporter }, pluginOptions) => {
   const requiredDirectories = [
     'src/data',
     'src/images',
     'src/content',
     'src/content/files',
     'src/code-examples',
-    'src/releases',
-  ];
+    pluginOptions.hasReleaseNotes && 'src/releases',
```
emmenko, line 24: FYI: @davifantasia you forgot this in the last PR

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -174,32 +177,61 @@ function generateReleaseNoteSlug(node) {
...
-              title
-              beta
-              isGlobalBeta
-              excludeFromSearchIndex
+        releaseNotes: allFile(
+          filter: {
+            sourceInstanceName: { eq: "releases" }
+            internal: { mediaType: { eq: "text/mdx" } }
+          }
+        ) @include(if: $hasReleaseNotes) {
```
emmenko, line 204: @davifantasia this is what I meant with the conditional directive. This way we still send 1 graphql query and we can use fragments.

emmenko: Damn, it does not work. I guess gatsby parses the schema first and it fails at that point.

packages/gatsby-theme-docs/README.md
```diff
@@ -70,6 +70,8 @@ The project structure should contain at least the following files and folders:
 
   - `beta` (_optional_): indicates that the website should be marked as **beta**. Each page gets a beta flag, no matter if the page frontmatter has it defined or not. Furthermore, in the main navigation, the beta flag is shown near the website title and not next to each link.
 
+  - `hasReleaseNotes` (_optional_): indicates that the website should include links to the Release Notes pages, in the main navigation.
```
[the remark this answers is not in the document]
emmenko: I see the point but when you start developing you can just use `process.env.NODE_ENV` to determine if the release notes should be generated on production or not.

For the noindex we already have the flag.

As for the gatsby node setup, we are leveraging gatsby schema inference for e.g. the frontmatter fields, which works one you have "real" pages. Not having pages would mean that we need to patch the schema types to include those "optional" page fields.
So I would prefer to only enable the release notes setup based on the option flag.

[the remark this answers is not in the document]
nkuehn: Let's speak about that, Im not sure we are speaking about the same thing? 

We can choose to start without an explicit schema now but my experience in the API Types and API Endpoints implementation is that you start to handle so many special cases and user errors that you end up with a fixed schema in any case.  
Here we would have to define the Schema for the `MdxFields` type and I can imaging that that would make a lot of the code simpler - maybe don't even have to have a config flag at all?  

Another advantage of explicitly defined schemata is that the error message that a user gets when they put schema-incompatible data in a document's frontmatter. With a defined schema the error is describing the actual problem.  With schema inference it's very hard to understand the error message because it just says that you have incompatible data on one document - and it's even random depending on which document is read first, it can happen that the correct document is complained about because the wrong one was read first. 

TL;DR:  Inferred schema can get really annoying to users.  If it's possible to define the `MdxFields` schema I vote for getting rid of the config flag altogether until we know more about the use cases.   Otherwise let's keep it as and do a refactor when the case comes up.  

PS: noindex is for the whole site but the situation I am speaking about is a site that is already indexed with content but gets release notes later.  Let's address that in a separate change when somebody really needs it. 

PPS: "Just" modifying code to  look into NODE_ENV isn't really an option for authors, but also here - we can address that if a case really comes up .

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -26,27 +33,29 @@ const ScrollContainer = styled.div`
...
+  content: '';
+  width: 100%;
+  height: ${scrollFaderHeight}px;
+  background: linear-gradient(
+    180deg,
+    ${designSystem.colors.light.surfaceSecondary2},
+    transparent
+  );
+  position: absolute;
+  bottom: -calc(${scrollFaderHeight}px + 2px);
```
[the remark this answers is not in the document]
emmenko: Weird, this worked on Firefox (I didn’t check Chrome). Thanks for pointing it out

packages/gatsby-theme-docs/README.md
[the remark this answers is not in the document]
emmenko: For the content pages we are also not using a self-defined schema. If that's the direction you want to go, it's fine by me.

Let me have another look then if we can remove the `hasReleaseNotes` flags altogether and infer that from the query results.

> noindex is for the whole site but the situation I am speaking about is a site that is already indexed with content but gets release notes later. Let's address that in a separate change when somebody really needs it.

The noindex option is available both globally but also on a per-page level.

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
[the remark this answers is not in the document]
emmenko: https://github.com/commercetools/commercetools-docs-kit/pull/336/commits/190ab93da31bd67a6d01c63cacd51e744e03e3fd

packages/gatsby-theme-docs/README.md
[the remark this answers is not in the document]
emmenko: Alright, I changed things a bit and created our own types and nodes. This simplifies several things actually, so it was a good thing to look into it. @nkuehn Thanks for pointing that out!

See https://github.com/commercetools/commercetools-docs-kit/pull/336/commits/5ba7b737a182236babf575b9dd1ab9397cbaac44

PS: I followed the example described in this article: https://www.christopherbiscardi.com/post/constructing-query-types-in-themes

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -173,33 +280,33 @@ function generateReleaseNoteSlug(node) {
...
+) {
+  const result = await graphql(`
     query QueryAllContentPages {
-      contents: allFile(
-        filter: {
-          sourceInstanceName: { eq: "content" }
-          internal: { mediaType: { eq: "text/mdx" } }
-        }
-      ) {
+      allContentPage {
```
emmenko, line 294: We now have our own custom query fields.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -173,33 +280,33 @@ function generateReleaseNoteSlug(node) {
...
-              excludeFromSearchIndex
-            }
-          }
-          name
+          slug
         }
       }
+      allReleaseNotePage {
+        totalCount
+      }
```
emmenko, line 301: We can use this to determine if there are any release notes in the website.

We can then be more specific and define some filters to which release notes we want to include. For example, we can exclude all the ones that are not marked as published.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -217,84 +324,67 @@ exports.createPages = async ({ graphql, actions, reporter }, pluginOptions) => {
...
-      // This is the slug you created before
-      // (or `node.frontmatter.slug`)
-      path: childMdx.fields.slug,
+      path: slug,
       // This component will wrap our MDX content
-      component: require.resolve('./src/templates/page-content.js'),
-      // You can use the values in this context in
-      // our page layout component
+      component: isOverviewPage
+        ? require.resolve('./src/templates/release-notes-list.js')
```
emmenko, line 341: The release overview page needs to be rendered here now. It still has its own template but from a graphql type perspective is considered a "content page" and therefore is queried with all other content pages.

The file is now `src/content/releases.mdx` instead of `src/releases/index.mdx`, for the reasons mentioned above.

packages/gatsby-theme-docs/src/templates/page-content.js
```diff
@@ -106,7 +54,11 @@ export default PageContentTemplate;
 
 export const query = graphql`
   query QueryMarkdownPage($slug: String!) {
-    mdx(fields: { slug: { eq: $slug } }) {
+    contentPage(slug: { eq: $slug }) {
+      title
+      beta
+      isGlobalBeta
+      excludeFromSearchIndex
```
emmenko, line 61: Those fields should be queried here, we don't need to pass them with the page context.
Also note the we are using the `contentPage` query field.

packages/gatsby-theme-docs/src/templates/release-notes-detail.js
```diff
@@ -0,0 +1,69 @@
...
+
+export default ReleaseNotesDetailTemplate;
+
+export const query = graphql`
+  query QueryReleaseDetailPage($slug: String!) {
+    releaseNotePage(slug: { eq: $slug }) {
+      title
+      isGlobalBeta
+      excludeFromSearchIndex
+      date(formatString: "YYYY-MM-DD")
```
emmenko, line 62: We can format the date here.

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -0,0 +1,111 @@
...
+      nodes {
+        slug
+        title
+        date(formatString: "YYYY-MM-DD")
+        description
+        type
+        topics
+        body
+      }
+    }
```
emmenko, line 109: Here we query all release notes and sort them by date descending.

packages/gatsby-theme-api-docs/src/hooks/use-type-locations.js
```diff
@@ -40,8 +43,12 @@ export const useTypeLocations = () => {
...
             }
             page: mdx {
-              fields {
-                slug
+              parent {
+                id
+                ... on File {
+                  relativePath
+                  ext
+                }
```
emmenko, line 51: This is now necessary because the `slug` is not created as a MDX field anymore.

emmenko:
@nkuehn the FOUC bug is fixed now, I'll merge the PR once CI passes.

PR #339, opened 2020-04-03
feat(release-notes): Release notes fields and paths

davifantasia:
Changes here include:

## Release notes custom fields
* date
* description
* type
* topics

## Paths
* Release notes overview page - `/releases`
* Release note details pages - `/releases/<date>-<title>`

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-node.js
+++ b/packages/gatsby-theme-docs/gatsby-node.js
@@ -29,43 +29,7 @@ exports.onPreBootstrap = ({ reporter }) => {
       fs.mkdirSync(dir);
     }
   });
-
-  createReleaseNoteIndex(reporter);
 };
-
-function createReleaseNoteIndex(reporter) {
-  const releaseNoteIndex = 'src/releases/index.mdx';
-  if (!fs.existsSync(releaseNoteIndex)) {
-    fs.writeFile(
-      releaseNoteIndex,
-      `---
-date: 1970-01-01 # necessary in this format
-title: (title goes here)
-description: |
-  (description for the template)
-type: enhancement # feature | fix | enhancement
-topics:
-  - foo
-  - barr
-  # Topics is an array, here are some examples:
-  #  - Products
-  #  - Product List
-  #  - Categories
-  #  - Carts
-  #  - API Clients
-  #  - Settings
----
-      `,
-      (error) => {
-        if (error) {
-          throw error;
-        }
-
-        reporter.info(`Successfully created "${releaseNoteIndex}"`);
-      }
-    );
-  }
-}
 
 exports.sourceNodes = ({ actions }) => {
   actions.createTypes(`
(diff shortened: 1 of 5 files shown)


packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -122,15 +122,54 @@ exports.onCreateNode = ({ node, getNode, actions }, pluginOptions) => {
     path.resolve('src/releases')
   );
   if (isReleaseNotesPage) {
-    const releaseNoteSlug = trimTrailingSlash(slug) || '/';
+    let releaseNoteSlug = generateReleaseNoteSlug(
+      node.frontmatter.date,
+      node.frontmatter.title,
+      node.fileAbsolutePath
+    );
+    releaseNoteSlug = trimTrailingSlash(releaseNoteSlug) || '/';
```
emmenko, line 130: Let's just move the `trimTrailingSlash(releaseNoteSlug) || '/'` part into the `generateReleaseNoteSlug` function

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -122,15 +122,54 @@ exports.onCreateNode = ({ node, getNode, actions }, pluginOptions) => {
     path.resolve('src/releases')
   );
   if (isReleaseNotesPage) {
-    const releaseNoteSlug = trimTrailingSlash(slug) || '/';
+    let releaseNoteSlug = generateReleaseNoteSlug(
+      node.frontmatter.date,
+      node.frontmatter.title,
+      node.fileAbsolutePath
+    );
```
emmenko, line 129: Nit: I think it's easier to just pass the `node` as single argument

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -122,15 +122,54 @@ exports.onCreateNode = ({ node, getNode, actions }, pluginOptions) => {
...
+
+  if (fileAbsolutePath.endsWith('index.mdx')) {
+    return basePath;
+  }
+
+  const slug = `${date.split('T')[0]} ${title}`
+    .toLowerCase()
+    .replace(/[^a-z0-9]+/g, '-') // replace non alphanumeric with hyphen
+    .replace(/(^-|-$)+/g, ''); // remove pre or post string hyphens
+  return `${basePath}/${slug}`.replace(/\/\/+/g, '/'); // replace two or more slashes with single slash
```
emmenko, line 170: Can you explain a bit how the slug is supposed to be built? And I mean not just to me but add some comments in the code (for future references). Thanks

davifantasia: After speaking with @nkuehn , it's better to use a library to be able to deal with things like transforming special characters and not miss information in the url. So I'm currently using [slugify](https://www.npmjs.com/package/slugify).

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -28,8 +29,44 @@ exports.onPreBootstrap = ({ reporter }) => {
...
+        if (error) {
+          throw error;
+        }
+
+        reporter.info(`Successfully created "${releaseNoteIndex}"`);
+      }
+    );
+  }
+}
+
```
emmenko, line 69: What is this? 🧐

davifantasia: Build fails for other micro-sites because release notes fields are not available to be queried, this just creates a file that provides the fields in graphql and as a release note template.

emmenko: Sorry but this is not the right approach.

First of all, release notes are optional for a website. Therefore we need to deal with that and do not assume that there will always be release notes.
For instance, in the my other PR I added a configuration option `hasReleaseNotes` that can be configured for the theme: https://github.com/commercetools/commercetools-docs-kit/pull/336/files#diff-afcde1e6d3112cafc056b077adde712fR73

Based on this flag, we should then setup and render the release notes related stuff or not.

Furthermore, if we need to ensure that fields/nodes are defined (in case they are optional) we need to tell gatsby about it. For example like we do with the navigation yaml: https://github.com/commercetools/commercetools-docs-kit/blob/e18623745a28e85c23581836a0d073f50cbbcf1a/packages/gatsby-theme-docs/gatsby-node.js#L33-L69

davifantasia: Why not define contents field in the same way you do with navigation yaml? Why does it have to be unique to release notes?

davifantasia:
> Nice, is it working as expected now?

Yes

PR #341, opened 2020-04-03
Tune writing style checker for plugin compatibility and custom apps

nkuehn:
* provides a local "raw"  vale command that the config can be directly passed to.  This proved more robust when using the VSCode plugin. 
 * adjust recommended VSCode workspace to use the config directly from the package to make transparent that it's not the same as an NPM version.  Also allows for quicker local feedback in the editor when tuning the rules. 
 * add a few acronyms that are common in software development 
 * allow the "ZEIT" brand in acronymys (false negative, it's not an aconym).

### DIFF
--- a/packages/writing-style/package.json
+++ b/packages/writing-style/package.json
@@ -15,7 +15,7 @@
   },
   "bin": {
     "commercetools-vale": "./bin/commercetools-vale.js",
-    "vale": "./bin/vale-2.1.0"
+    "commercetools-vale-bin": "./bin/vale-2.1.0"
   },
   "scripts": {
     "postinstall": "node scripts/download-vale.js",
--- a/packages/writing-style/package.json
+++ b/packages/writing-style/package.json
@@ -14,7 +14,8 @@
     "url": "https://github.com/commercetools/commercetools-docs-kit.git"
   },
   "bin": {
-    "commercetools-vale": "./bin/commercetools-vale.js"
+    "commercetools-vale": "./bin/commercetools-vale.js",
+    "vale": "./bin/vale-2.1.0"
   },
   "scripts": {
     "postinstall": "node scripts/download-vale.js",
--- a/packages/writing-style/README.md
+++ b/packages/writing-style/README.md
@@ -3,7 +3,8 @@
 This package provides
 
 - commercetools specific writing style and terminology rules for the configurable [vale](https://errata-ai.github.io/vale/) prose linter.
-- the `commercetools-vale` command line tool which wraps a bundled vale build and calls it using the included writing style and configuration
+- the `commercetools-vale` command-line tool which wraps a bundled vale build and calls it using the included writing style and configuration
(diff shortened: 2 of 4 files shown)


packages/writing-style/package.json
```diff
@@ -14,7 +14,8 @@
     "url": "https://github.com/commercetools/commercetools-docs-kit.git"
   },
   "bin": {
-    "commercetools-vale": "./bin/commercetools-vale.js"
+    "commercetools-vale": "./bin/commercetools-vale.js",
+    "vale": "./bin/vale-2.1.0"
```
emmenko, line 18: We should be careful with this, as the symlinked binary name might conflict with other packages exposing such name.
We should try to use a more specific name.

emmenko: Or we don't expose the npm binary and just have a "proxy" bash script or something that points to the specific binary version.
I suppose this is mostly the case when you want to use it in e.g. VSCode.
With that also, we can read the vale version from the package.json and use it in the proxy script, instead of having the version hardcoded multiple times.

nkuehn: I'm renaming the symlink and adding it to the Readme.    Forgot that you can install it globally and that would collide with a default vale installation.

emmenko: It's not just the global installation, but also the `node_modules/.bin` symlinks

nkuehn:
PS: the plugin now works for me, too :-)

packages/writing-style/package.json
emmenko: Have you considered my alternative proposal, of having a "proxy" script?

nkuehn: Yes, but the technical intent here is to get rid of any wrappers (for other reasons, too like keeping the color output). 
is the new name ok?

emmenko: My point was that if we just need access to the vale binary for the vscode setup, I wouldn't put it in the npm binary list because we already have the `commercetools-vale` binary.

In vscode you just point the binary to

```
"vscode-vale.path": "${workspaceFolder}/node_modules/@commercetools-docs/writing-styleles/vale-bin",
"vscode-vale.configPath": "${workspaceFolder}/node_modules/@commercetools-docs/writing-styleles/.vale.ini",
```

The `vale-bin` then is just a proxy script to the downloaded binary version.

nkuehn: Yes, I got your point but I (as one of the actual users) _do_  want to call it without going through a proxy script to get colored console output.   Can you live with that fact?  I wrote it in the above comment for a reason.

emmenko: I think you're still missing my point. The `commercetools-vale` stays as-is, no changes there. I assume that users will primarily use that binary, if they want to.

Now, regarding the second binary: my question (which you still haven't answered) is whether this binary is only supposed to be used for the VSCode plugin or are there any other use cases for which you (as a user) would need to use it directly, instead of using the `commercetools-vale` binary.

No hard feelings, I'm just trying to understand the requirements as it's still not clear to me.

emmenko:
Don't forget to assign the label, thanks

PR #347, opened 2020-04-07
fix(theme): to use height auto for small screens

### LINKED ISSUE #345: Bug: prints only first page
We have use cases where a page must be printeable in its full length.   These include not publicly listed pages that are documented by customers as a PDF (e.g. a list of APIs, in the future price lists for training and PS services etc. ).   Currently only the first page is printed.

The most minimalist way to do it to let the div that has `.. role="application" ..` have `height: auto; width: auto;`  instead of 100vh/100vw  in a print media query.  

Printing does not have to be perfect or specially designed but at least work at all.   It's actually pretty okay,  we might opt to hide the main footer part and just leave the one with "copyright 2020 commercetools".  But that's all optional.

emmenko:
Fixes #345 

It works for Chrome and Safari. On Firefox I wasn't able to make it work, but it seems that Firefox has problems with printing pages in general.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -39,5 +39,9 @@
     "no-missing-end-of-source-newline": null,
     "function-name-case": null,
     "value-keyword-case": null,
+    "media-feature-name-no-unknown": [
+      true,
+      { "ignoreMediaFeatureNames": [/percy/] }
+    ]
   }
 }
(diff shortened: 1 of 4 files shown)


emmenko:
I don't understand why the percy snapshot is failing here...🧐

Issue #346, opened 2020-04-07
Cookie consent overlay does not fade away after accepting all cookies


davifantasia:
Not only does the translucent overlay not fade away, but it appears clicking on accept does not work either because after refresh, cookies are asked to be accepted again. So basically the site is unclickable and unscrollable.

### Reproduce
- Visit https://commercetools-docs-kit.now.sh/docs-smoke-test/
- Click on "Accept All Cookies"

Browser: Chrome

Image of site with overlay _before_ clicking "Accept All Cookies":
![Screenshot 2020-04-07 at 10 51 10](https://user-images.githubusercontent.com/8238495/78649514-d3747c00-78bd-11ea-857f-b4a836a5f51c.png)

Image of site with overlay _after_ clicking "Accept All Cookies":
![Screenshot 2020-04-07 at 10 49 31](https://user-images.githubusercontent.com/8238495/78649291-87293c00-78bd-11ea-9e4d-a498028db517.png)

emmenko:
The cookie consent is supposed to "work" only on commercetools.com domains, that's why it does not get away. On the other hand, I expect it to be visible only on commercetools.com domains.

It's just a monumental shitty script.

Maybe we should render it dynamically by checking first what `window.origin` is. So if you are indeed in commercetools.com, we attach the related scripts to the dom, otherwise we don't.

I can look into it if you want.

nkuehn:
@emmenko  let me that that one, I was able to get it done on docs.commercetools.com and we already ran that circle that the script does not work at all when dynamically attached (that's why it's in `gatsby-ssr.js` now).

emmenko:
So what do you plan to do?

nkuehn:
See PR #348   (do it as we successfully do on the jekyll site for over a year now to avoid throwing more worktime into this sinkhole)

emmenko:
This is fixed now.

PR #348, opened 2020-04-07
fix(theme): dynamically render the cookie consent scripts only on *.commercetools.com domain

nkuehn:
addresses #346 

This is the technique that is used on the current docs.commercetools.com  site - it's a client side approach that activates a CSS rule that removes the complete thing from the visible DOM, letting it do it's buggy work however it wants as long as it doesn't show up. 

The check is on the .commercetools.com  domain level because that's the cookie scope the script uses. 
 The root cause it that the cookie consent script fails on other domains and does not close.

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-browser.js
+++ b/packages/gatsby-theme-docs/gatsby-browser.js
@@ -31,6 +31,8 @@ export const onClientEntry = (
         whitelistUrls: ['docs.commercetools.com', 'now.sh'],
       });
     });
+    if (window && !window.location.host.includes('.commercetools.com'))
+      document.body.classList.add('not-on-cookie-domain');
   }
 
   // Require additional Prism languages.
--- a/packages/gatsby-theme-docs/gatsby-browser.js
+++ b/packages/gatsby-theme-docs/gatsby-browser.js
@@ -17,6 +17,17 @@ const environment =
     ? 'production'
     : 'preview';
 
+const injectScript = (url, attributes = {}, onLoad) => {
+  const script = document.createElement('script');
+  script.type = 'text/javascript';
+  script.src = url;
+  Object.keys(attributes).forEach((key) => {
+    script.setAttribute(key, attributes[key]);
+  });
+  if (onLoad) script.onload = onLoad;
+  document.body.appendChild(script);
+};
+
 export const onClientEntry = (
   _, // eslint-disable-line
   pluginOptions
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-docs/gatsby-browser.js
```diff
@@ -31,6 +31,8 @@ export const onClientEntry = (
         whitelistUrls: ['docs.commercetools.com', 'now.sh'],
       });
     });
+    if (window && !window.location.host.includes('.commercetools.com'))
+      document.body.classList.add('not-on-cookie-domain');
```
nkuehn, line 35: Trying to be explicit in the name - that's what it really does, it's not actually "Production" or "staging". 

"window" boolean check intends to make sure that it does not run server side. 

using the not completely backwards compatible classList()  call  is okay since it's only called when not called on our production domain, which is ourselves looking at previews.

packages/gatsby-theme-docs/gatsby-ssr.js
```diff
@@ -94,5 +95,16 @@ export const replaceRenderer = ({
         __html: css,
       }}
     />,
+    <style
```
nkuehn, line 98: I could have put that into the global emotion style too but I think it's better to handle the topic in the same place as other such purely technical hacks.

packages/gatsby-theme-docs/gatsby-browser.js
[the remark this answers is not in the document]
emmenko: Ah I guess this won't work because the banner is rendered after... 🤔

Let me see if I can find a better solution. I just would like to avoid using class names.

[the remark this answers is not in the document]
nkuehn: Yes, that stupid script is not completed on load, it first triggers fetching a jquery somewhere and when that is there etc. pp.  
JS is imperative and hence fragile to what happens when on the time scale.  CSS is declarative and you can decide to just not care.   

We  _could_   do an engineering beauty contest and  hook a MutationObsever to the body element, watching for "childlist" changes and then manually killing / hiding them if they are the optanon stuff etc. pp.  

This beast is challenging the stubborn engineer in me, too that there _must_ be a clean solution but the manager in me says we're already 600% over what the topic should be worth in dev time budget :-)

packages/gatsby-theme-docs/gatsby-browser.js
```diff
@@ -32,6 +43,17 @@ export const onClientEntry = (
...
+  if (window && window.location.host.includes('.commercetools.com')) {
+    injectScript(
+      'https://cdn.cookielaw.org/consent/b104027d-4d10-4b75-9675-9ffef11562a8/OtAutoBlock.js'
+    );
+    injectScript(
+      'https://cdn.cookielaw.org/scripttemplates/otSDKStub.js',
+      { 'data-domain-script': 'b104027d-4d10-4b75-9675-9ffef11562a8' },
+      'function OptanonWrapper() {};'
+    );
+  }
```
emmenko, line 56: This seems to do the trick.



I'd like to try this approach first.

emmenko:
@nkuehn please take a look, I would like to try this approach first.

nkuehn:
OK, just merge and approve yourself - we'll have to wait all through production anyways to test.

emmenko:
I'll test the canary release on appkit.

PR #351, opened 2020-04-08
fix(theme): remove check for zeit github branch

### LINKED ISSUE #349: Cannot reliably depend on NOW_GITHUB_* environment variables.
Ref #342 

TL;DR: the Github ZEIT Now integration seems to alias the domain when a PR is merged to master. This makes sense on one side, because there are technically no changes, but as soon as we rely on `NOW_GITHUB_*` environment variables, we have a problem.

This is the case for things like rendering the cookie consent scripts.

https://github.com/commercetools/commercetools-docs-kit/blob/40be3d66ada8f38671f7844d284d6afca2c2c0b9/packages/gatsby-theme-docs/gatsby-ssr.js#L16-L18

See:
- https://spectrum.chat/zeit/now/now-for-github-can-we-force-deployment-building-upon-merges-to-master~71a27bb7-1b6a-4fdd-bcc6-8d9f1c39f0c0
- https://github.com/zeit/now/issues/2407

We need to find a different solution for conditionally render elements or not, as we can't rely on those environment variables anymore.

emmenko:
Fixes #349

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-browser.js
+++ b/packages/gatsby-theme-docs/gatsby-browser.js
@@ -12,10 +12,6 @@ import { docsCache } from './create-emotion-cache';
 
 const isProduction = process.env.GATSBY_NODE_ENV === 'production';
 const commitSha = process.env.GATSBY_NOW_GITHUB_COMMIT_SHA;
-const environment =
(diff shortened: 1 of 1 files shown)


emmenko:
I'll merge the other PR first #348

PR #353, opened 2020-04-08
chore(ci): add workflow to verify PR labels are assigned

emmenko:
~~See https://github.com/jesusvasquez333/verify-pr-label-action~~

So that I don't have to remind you to assign the labels on each PR 😅


emmenko:
Hmm the action is a bit too verbose.

.github/workflows/verify-pr-labels.yml
```diff
@@ -0,0 +1,20 @@
+name: Verify Pull Request Labels
+
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
```
nkuehn, line 5: I agree, that's too verbose in this form. 

To avoid hassling us on draft PR you can add a `if: github.event.pull_request.draft == false`  condition and add the `ready_for_review` event type to trigger it  

Removing `synchronize` would help too, the need for a label does not change when you push to the branch.

nkuehn:
This is a friendly bot now :-)

PR #354, opened 2020-04-09
fix(theme): to use height auto for small screens

emmenko:
Duplicate of #347 

Just checking if the percy snapshots fail here as well.


emmenko:
Wow, so this is actually related to the height changes...what the hell?? 🤔

Issue #361, opened 2020-04-14
Release Notes: generate RSS feed


nkuehn:
* The feed needs to be backwards compatible with the existing feed, esp. concerning the "guid" of the posts, to prevent all entries to re-appear in all consumers when migrating. See 
 https://docs.commercetools.com/release-notes.xml  
  * !! an alternative option to full back compatibility is to configure the RSS to have a cutoff date so we include only new releases after migrating
  * !! another idea:  If there is a "slug" field present (shows that it's migrated from the old site), don't include it  ->  but that would become a legacy soon. 
 * There's a pretty powerful RSS plugin (official) for gatsby, need to tune the custom GraphQL et.c 
 * Feed URL  https://docs.commercetools.com/microsite-name/release-notes.rss

PR #364, opened 2020-04-14
feat: generate rss feed

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/361


packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -36,11 +36,14 @@ module.exports = (themeOptions = {}) => {
...
 
+  const productionHostname = 'docs.commercetools.com';
+
   return {
     siteMetadata: {
       author: 'commercetools',
-      productionHostname: 'docs.commercetools.com',
+      productionHostname,
       betaLink: null,
+      siteUrl: `https://${productionHostname}/${pluginOptions.websiteKey}`,
```
emmenko, line 46: Note: the `websiteKey` is not a guarantee that it's the same as the URL path.

What you're looking for is the `pathPrefix` I think.

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -210,6 +213,54 @@ module.exports = (themeOptions = {}) => {
       'gatsby-plugin-remove-trailing-slashes',
       'gatsby-plugin-meta-redirect',
       'gatsby-plugin-netlify-cache',
+      {
+        resolve: 'gatsby-plugin-feed',
```
emmenko, line 217: Wondering if we should make this configurable 🤔

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -210,6 +213,54 @@ module.exports = (themeOptions = {}) => {
...
+                });
+              },
+              query: `
+              {
+                allReleaseNotePage(
+                  sort: { order: DESC, fields: date },
+                ) {
+                    nodes {
+                      description
+                      body
```
nkuehn, line 250: If we don't find a solution for the encoding (see above) this field does not need to be queried any more.

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -210,6 +213,54 @@ module.exports = (themeOptions = {}) => {
...
+        `,
+          feeds: [
+            {
+              serialize: ({ query: { site, allReleaseNotePage } }) => {
+                return allReleaseNotePage.nodes.map((node) => {
+                  return {
+                    ...node,
+                    url: site.siteMetadata.siteUrl + node.slug,
+                    guid: site.siteMetadata.siteUrl + node.slug,
+                    custom_elements: [{ 'content:encoded': node.body }],
```
nkuehn, line 239: MDX is not HTML, you can't just trow MDX into an RSS as HTML encoded content.   

I don't know if it's realistic to get a usable plaintext or "minimal HTML" representation directly from the GraphQL for our (sometimes custom) MDX, so I wold be fine omitting it as we currently do, too.  

But we can try - one option that publishers take is to put the excerpt until "read more" here so people move on to your actual website instead of staying in some feed consumer UI.  But even then it has to be either plaintext or plain semantic HTML

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -210,6 +213,54 @@ module.exports = (themeOptions = {}) => {
...
+                      body
+                      slug
+                      title
+                      date
+                    }
+                }
+              }
+            `,
+              output: '/release-notes.xml',
+              title: `${pluginOptions.websiteKey}'s rss feed`,
```
nkuehn, line 259: Let's try a better name  :-)     

pseudocode proposal:  `commercetools ${siteMetadata.title} Release Notes`

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -210,6 +213,54 @@ module.exports = (themeOptions = {}) => {
       'gatsby-plugin-remove-trailing-slashes',
       'gatsby-plugin-meta-redirect',
       'gatsby-plugin-netlify-cache',
+      {
```
nkuehn, line 216: The feed generator needs to go above the plugins that explicitly have a comment "the following plugins need to be last" above them.

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -210,6 +213,54 @@ module.exports = (themeOptions = {}) => {
...
+              {
+                allReleaseNotePage(
+                  limit: 5, sort: { order: DESC, fields: date }, filter: {date: {gt: "1999-01-01"}},
+                ) {
+                    nodes {
+                      description
+                      slug
+                      title
+                      date
+                      categories: topics
```
nkuehn, line 253: I added this as a quick win feature

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -267,7 +267,7 @@ function generateReleaseNoteSlug(node) {
   const date = node.frontmatter.date ? node.frontmatter.date.split('T')[0] : '';
   const title = node.frontmatter.title ? node.frontmatter.title : '';
 
-  const slug = slugify(`${date} ${title}`, { lower: true });
+  const slug = slugify(`${date} ${title}`, { lower: true, strict: true });
```
nkuehn, line 270: I did add "strict" to the slugify to make sure e.g. not parantheses and other characters are in the URL to have them consistent with all existing URLs.

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -210,6 +213,54 @@ module.exports = (themeOptions = {}) => {
...
+                    }
+                }
+              }
+            `,
+              output: '/release-notes.xml',
```
emmenko, line 258: @nkuehn just to clarify, will we have the following URL routes?

- `/releases`
- `/release-notes.xml`

emmenko:
> only load and run the plugin when the theme option "hasReleaseNotes" is true.

@nkuehn there is no such option anymore.

packages/gatsby-theme-docs/gatsby-config.js
nkuehn: Yes.  If they are not horribly bad I keep the existing URL patterns.  But if you find them ugly and aren't scared of ending up with a big massive redirects setup for migration we can "clean it up", too and make it  `/releases.xml`. 
No real emotions, just push it.

nkuehn:
> @nkuehn there is no such option anymore.

You're right.  I assumed it's still there because it's still set in our test site's config: 
 https://github.com/commercetools/commercetools-docs-kit/blob/master/websites/docs-smoke-test/gatsby-config.js#L17 

I'm fine with an empty RSS feed if that means we can avoid a new config option.

packages/gatsby-theme-docs/gatsby-config.js
emmenko: Ok, I was thinking more of something like `/releases/feed.xml`

nkuehn: Fine too.  Including the XML extension prevents collisions and  `/feed`  was a conventional URL.

websites/docs-smoke-test/gatsby-config.js
```diff
@@ -12,9 +12,9 @@ module.exports = {
     {
       resolve: '@commercetools-docs/gatsby-theme-docs',
       options: {
+        title: 'Docs Smoke Test',
```
emmenko, line 15: This is also duplicated

davifantasia: Yea, not quite sure what the alternative would look like yet.

emmenko: We could "move" the `siteMetadata` as a theme plugin option, then spread the value in the theme `gatsby-config.js` as we have some defaults for `siteMetadata`.

This would be a breaking change though.

davifantasia: I'm looking to override the title in the `setup` field of the plugin, where we have access to the siteMetadata.

https://www.gatsbyjs.org/docs/adding-an-rss-feed/#syntax-for-itunes-rss-blocks

But I see that the feed's global title is the same as the description so I'm trying to fix that now.

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -203,6 +203,67 @@ module.exports = (themeOptions = {}) => {
...
+              siteMetadata {
+                title
+                description
+                siteUrl
+                site_url: siteUrl
+              }
+            }
+          }
+        `,
+          setup: (options) => {
```
davifantasia, line 221: We can override the site title here to customize with title from `siteMetadata`.

But the link and description of the channel are lost once `setup` setup is implemented. Even when I tested with just returning the option parameter.

emmenko: You need to spread the options like this: https://github.com/gatsbyjs/gatsby/blob/055844312b8e681673fd453a8a556693eabb7ac1/packages/gatsby-plugin-feed/src/internals.js#L29-L39

websites/docs-smoke-test/gatsby-config.js
nkuehn: @emmenko  we just refactored the way the options are read, no duplicates any more

emmenko: Thanks, I like the `setup` approach. The configuration has still an error though (see CI)

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -203,6 +203,68 @@ module.exports = (themeOptions = {}) => {
...
+              },
+            } = options;
+            return {
+              ...options,
+              title: `commercetools ${title} Release Notes`,
+              description,
+              site_url: siteUrl,
+              language: 'en',
+              categories: ['commercetools', 'e-commerce'],
+            };
```
emmenko, line 235:
```suggestion
            const {
              query: {
                site: { siteMetadata },
              },
              ...rest,
            } = options;
            return {
              ...siteMetadata,
              ...rest,
              title: `commercetools ${title} Release Notes`,
              site_url: siteUrl,
              language: 'en',
              categories: ['commercetools', 'e-commerce'],
            };
```

davifantasia: I'm working on it.

PR #366, opened 2020-04-15
Writing Style Linter: Update vale and prevent binary from being published

nkuehn:
This fixes the notorious issue that the wrong vale binary was executed on linux hosts, e.g. CI.   I updated `.gitignore` when changing the binary naming convention but oversaw `.npmignore`. 

Since this requires a new package version anyways, I am updating the vale version to a recent bugfix in the same step.


emmenko:
That's also what I meant here: https://github.com/commercetools/commercetools-docs-kit/pull/341#discussion_r403715523

> With that also, we can read the vale version from the package.json and use it in the proxy script, instead of having the version hardcoded multiple times.

To avoid this situations where we forget to update both versions.

nkuehn:
> That's also what I meant here: [#341 (comment)](https://github.com/commercetools/commercetools-docs-kit/pull/341#discussion_r403715523)
> 
> > With that also, we can read the vale version from the package.json and use it in the proxy script, instead of having the version hardcoded multiple times.
> 
> To avoid this situations where we forget to update both versions.

OK, you beat me to try it now.   I've just pushed a wrapper that is actually preserving all console interaction and coloring etc by not using the fat `shelljs` wrapper but calling a process spawn that directly connects the stdin / stdout / stderr pipes without node.js being able to read them.  

That did the job, so here we go.   

We could remove shelljs in the other script, too  - I didn't do it here because I don't remember why it was introduced from the beginning.

emmenko:
@nkuehn I pushed a commit to reuse the common logic between the bin scripts. Please have a look

packages/writing-style/bin/commercetools-vale.js
```diff
@@ -1,25 +1,11 @@
...
 const valeConfig = path.join(__dirname, `../.vale.ini`);
-let params = process.argv.slice(2);
-// simplistic check whether the --config param was passed empty by the VSCode plugin (empty = followed by another flag)
-const configParamPosition = params.findIndex((p) => p === '--config');
-if (
-  params[configParamPosition + 1] &&
-  params[configParamPosition + 1].startsWith('-')
-) {
-  params = params.filter((p) => p !== '--config');
-}
```
emmenko, line 21: As far as I understand, this is not necessary anymore, as for VSCode we can use the "config" option of the plugin and use the `vale-bin` binary.

nkuehn:
@emmenko  since you moved the binary to the root folder the ignore rule was a bit dangerous to also cover the `vale-styles` directory so I renamed it to just `styles` which is the convention anyways in vale. 

I overhauled the README to explain the VSCode usage and some other things.

emmenko:
Perfect, thanks!

PR #370, opened 2020-04-17
fix(theme): to not restrict content page nodes for specific filesystem

emmenko:
The appkit ui components pages are currently broken because their are not built anymore.
This was because we assumed that the pages were built only from the `sourceInstanceName: content` (my mistake).

Additionally, I moved the sourcefile for the code examples into its related transformer, to keep things tight together.

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -89,14 +89,6 @@ module.exports = (themeOptions = {}) => {
           path: path.resolve(`./src/content`),
         },
       },
-      // Code examples
-      {
-        resolve: 'gatsby-source-filesystem',
-        options: {
-          name: 'codeExamples',
-          path: path.resolve(`./src/code-examples`),
-        },
-      },
       // Release notes
       {
         resolve: 'gatsby-source-filesystem',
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -89,14 +89,6 @@ module.exports = (themeOptions = {}) => {
           path: path.resolve(`./src/content`),
         },
       },
-      // Code examples
-      {
```
emmenko, line 93: Moved to the transformer

packages/gatsby-transformer-code-examples/gatsby-node.js
```diff
@@ -73,14 +86,11 @@ async function onCreateNode({
...
     'CodeExample'
   );
-}
+};
 
 function isValidNode(node) {
   return (
     validMediaTypes.includes(node.internal.mediaType) &&
-    node.dir.startsWith(examplesDirectory)
+    node.sourceInstanceName === 'codeExamples'
```
emmenko, line 94: Here we can simply use this check, which is more robust.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -267,6 +231,39 @@ exports.onCreateNode = (
       child: node,
     });
   }
+
+  // If not explicitly handled, always fall back to build the page as a "content" page.
+  // This is useful in case the website requires additional MDX pages located in
+  // other file system directories, and thus with different `sourceInstanceName` names.
+  const contentPageFieldData = {
```
emmenko, line 238: The fix is to fall back to building the node as a "content" page, if not explicitly handled (e.g. for release notes)

Issue #360, opened 2020-04-14
Release Notes: Details Pages (designed)


nkuehn:
* [x] Design and generate the Details pages.  
 * [x] Find a solution to use an explicit excerpt separator instead of chopping off after a given number of characters.  The remark plugin supports this on GraphQL but the MDX plugin doesn't.  Maybe we can do something in the layer that creates our custom grapql Schema. 
 * [x] change the GraphQL query of the overview page to not load more than the excerpt
 * [x] Change the overview list page to show the "Read More"  link if there is such a separator in place
 * [x] NEW: Make the title of the post in the overview a link to the details page, also for entries that do not have a read more.   Styling decided with Sven (design):  Like in the top navigation, i.e. black but on hover colored in that color.   

Design - https://commercetools.invisionapp.com/d/main?origin=v7#/console/19438556/412794930/preview?scrollOffset=299

nkuehn:
@davifantasia  I've updated the description of this issue to address the problem that MDX does not have excerpt separators and added the UX requirement that the title in the overview is a link.

davifantasia:
This is done now.

Issue #321, opened 2020-03-25
Descriptions of properties of a type should formatted using design guidelines


davifantasia:
## Problem
Descriptions of properties of a type currently use basic html formatting, see the following examples:

HTML code element on this page - https://commercetools-docs.now.sh/import-api/import-operation#importoperation
![Screenshot 2020-03-25 at 12 55 02](https://user-images.githubusercontent.com/8238495/77533843-e63b8980-6e97-11ea-9226-0b45eb4183b1.png)

Anchor link on this page - https://commercetools-docs.now.sh/import-api/product-draft-import#pricedraftimport
![Screenshot 2020-03-25 at 12 52 07](https://user-images.githubusercontent.com/8238495/77533628-81802f00-6e97-11ea-908e-e4b04235a087.png)

## Design
https://commercetools.invisionapp.com/d/main?origin=v7#/console/17784193/390033493/preview?scrollOffset=425

EDIT:

The task to parse only the allowed markdown syntaxes and render custom components should be a separate PR.
- disallowed markdowns will be rendered as text

nkuehn:
I think this is is the option in the `remark-reactjs` lib you need to fill with a selection of our components (from the docs-ui-kit package):

https://github.com/remarkjs/remark-react#optionsremarkreactcomponents

We only need to provide the components that are actually allowed (see confluence)

davifantasia:
Fixed in https://github.com/commercetools/commercetools-docs-kit/pull/375

PR #375, opened 2020-04-17
Release Notes: Render excerpt in list with reduced formatting, show read more. 

nkuehn:
This partially supports #360 but not completely. 

Because it introduces the infrastructure for the same use case this also addresses #321  (proper rendering of markdown in API RAML description texts)

Rendering excerpts turned out to not be possible using gatsby's original `excerpt_separator` configuration because this setting is only applicable to the remark plugin, but not to MDX.  The MDX plugin does only support plaintext excerpts that are chopped off at a specific number of characters. 

Also, I originally intended to only allow rendering a subset of markdown in the release notes list view.  We left the requirement behind to save complexity, but now it turned out it's actually easier to do it this way because we have to build a "safe" limited markdown renderer anyways for other places where small fragments of MD are placed into other layouts. 

Solution here is to "chop" the raw MDX into halves ourselves and providing the first part as a new property on GraphQL.  Plus a boolean convenience flag that shows that there is not more in the full body (it's hard to compare because the full body is javascript source code already). 

The Component that renders a release note now behaves differently depending on whether the full body was passed or just an excerpt.  

If only the excerpt is passed it's rendering this using a plain markdown renderer that has all HTML / MD constructs "disabled" that could lead to problems e.g. if the content would be placed into a table cell or inside other blocks.


packages/gatsby-theme-docs/src/layouts/internals/layout-release-note.js
```diff
@@ -59,7 +66,9 @@ ReleaseNote.propTypes = {
   description: PropTypes.string.isRequired,
   type: PropTypes.oneOf(['feature', 'enhancement', 'fix']).isRequired,
   topics: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
-  body: PropTypes.string.isRequired,
+  body: PropTypes.string,
+  rawExcerpt: PropTypes.string,
```
nkuehn, line 70: This is not ideal, but since the GraphQL data source is already strongly typed I think it's acceptable.  We could write a custom validation function to make sure either one is passed if we think this component will be re-used a lot.

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -73,7 +62,8 @@ ReleaseNotesListTemplate.propTypes = {
           description: PropTypes.string.isRequired,
           type: PropTypes.string.isRequired,
           topics: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
-          body: PropTypes.string.isRequired,
+          rawExcerpt: PropTypes.string.isRequired,
```
nkuehn, line 65: Side effect: not loading the full content is good for the page weight too.

nkuehn:
Side remark: We shouldn't forget to introduce visual regression tests for the release notes and consider expanding to some of the API documentation cases again. 
I expected this to break the VRTs for the API docs but it doesn't  - the cases it changes aren't tested.

websites/docs-smoke-test/src/releases/excerpt-rendering-test.mdx
```diff
@@ -0,0 +1,82 @@
...
+  <dd>Definition of the Term</dd>
+</dl>
+
+separator line below this
+
+----
+
+<!--more-->
+
+----
```
emmenko, line 49: So is this format how you want people to use the "write more"?

Feels a bit fragile, maybe it's just my impression.

packages/gatsby-theme-docs/src/layouts/internals/layout-release-note.js
emmenko: Hmm but both are required fields, aren't they?

I would expect the logic to be (pseudo-code)

```
if hasMore
  render excerpt
  render read more
else
  render body
```

packages/gatsby-theme-docs/src/layouts/internals/layout-release-note.js
```diff
@@ -59,7 +66,9 @@ ReleaseNote.propTypes = {
   description: PropTypes.string.isRequired,
   type: PropTypes.oneOf(['feature', 'enhancement', 'fix']).isRequired,
   topics: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
-  body: PropTypes.string.isRequired,
+  body: PropTypes.string,
```
davifantasia, line 69: Since you're no longer querying for `body` from `src/templates/release-notes-list.js`, this prop is no longer necessary.

packages/gatsby-theme-docs/src/layouts/internals/layout-release-note.js
nkuehn: The excerpt is going through a different renderer now than the full body, so in the list view it's always render excerpt and the "If" just covers whether to show the link or not.  

If the body is made required, we would pump up the data size of the page query of the list view with all the unused bodies.

websites/docs-smoke-test/src/releases/excerpt-rendering-test.mdx
nkuehn: We went through that decisionmaking in a PR that was explicitly about agreeing on the format specification. 

That format is the current format (but that could be migrated).   One requirement is that it must be handled opaque by Markdown preview implementations, so it must be either a valid but invisible markdown element or an HTML comment.  

I proposed a different format, too (e.g. just a horizontal line) but the tech writers preferred this format becuause it's verbally explicit.

emmenko: Thanks for the heads up.

packages/gatsby-theme-docs/src/layouts/internals/layout-release-note.js
nkuehn: The component is used on the details page, too so it needs to be able to accept either a full body or just an excerpt. 

But I could refactor it in a way that the excerpt is passed as the body and in addition we pass just the boolean that tells which renderer to use.

emmenko: > But I could refactor it in a way that the excerpt is passed as the body and in addition we pass just the boolean that tells which renderer to use.

Yes that would be preferrable. This component should not concern itself where it's rendered, the parent knows the context.

nkuehn: @davifantasia  let's talk about this one in our half hour today - I'm confused about my own implementation now, maybe it has to do with work you did on the details pages. 

There isn't any second use of this component any more in the codebase.

packages/gatsby-theme-docs/src/layouts/internals/layout-release-note.js
```diff
@@ -60,6 +67,8 @@ ReleaseNote.propTypes = {
   type: PropTypes.oneOf(['feature', 'enhancement', 'fix']).isRequired,
   topics: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
   body: PropTypes.string.isRequired,
+  bodyRenderer: PropTypes.oneOf(['mdx', 'excerpt']).isRequired,
```
emmenko, line 70: Hmm, let me push a commit...

emmenko: So: https://github.com/commercetools/commercetools-docs-kit/pull/375/commits/3d32a5669530177268f50ef765a9e93870fde48a

TL;DR: The parent should just format the body and pass it as a children.

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -31,11 +32,13 @@ const ReleaseNotesListTemplate = (props) => (
...
-                  key={releaseNote.slug}
-                  bodyRenderer="excerpt"
-                  {...releaseNote}
-                />
+                <LayoutReleaseNote key={releaseNote.slug} {...releaseNote}>
+                  <Markdown.TypographyPage>
+                    <section>
+                      {markdownFragmentToReact(releaseNote.body)}
+                    </section>
+                  </Markdown.TypographyPage>
```
emmenko, line 40: The parent can decide how the render the body and pass it as a child.

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -31,11 +32,13 @@ const ReleaseNotesListTemplate = (props) => (
...
               props.data.allReleaseNotePage.nodes &&
               props.data.allReleaseNotePage.nodes.map((releaseNote) => (
-                <LayoutReleaseNote
-                  key={releaseNote.slug}
-                  bodyRenderer="excerpt"
-                  {...releaseNote}
-                />
+                <LayoutReleaseNote key={releaseNote.slug} {...releaseNote}>
+                  <Markdown.TypographyPage>
+                    <section>
```
emmenko, line 37: @nkuehn should these elements have some more contextual info? Like ids, labels, ...

packages/gatsby-theme-docs/src/layouts/internals/layout-release-note.js
emmenko: See https://github.com/commercetools/commercetools-docs-kit/pull/375#discussion_r411267318

packages/gatsby-theme-docs/src/layouts/internals/layout-release-note.js
```diff
@@ -45,24 +45,23 @@ const ReleaseNote = (props) => {
           ))}
         </Topics>
       )}
-
-      <div>
-        <MDXRenderer>{props.body}</MDXRenderer>
-      </div>
+      {props.children}
+      {props.hasMore ? <Link href={`${props.slug}`}>Read more...</Link> : null}
```
emmenko, line 49:
```suggestion
      {props.hasMore ? <Link href={props.slug}>Read more...</Link> : null}
```

PR #376, opened 2020-04-17
feat(docs-theme): make beta flag available in any content

nkuehn:
To make the beta flag usable in any content its size is changed
to a relative measure calculated from the design spec.

To align it visually straighter it's middle-aligned with the surrounding text now.

The change leads to factual beta badge size changes in the area of < 1px  because I wanted to keep re-using the design system values instead of calculating theoretical font sizes e.g. for the page header.  

It's good for me this way and overall the better solution than to have to maintain a badge style for every potential situation it's used in

### DIFF
--- a/packages/ui-kit/src/design-system.js
+++ b/packages/ui-kit/src/design-system.js
@@ -134,6 +134,11 @@ export const typography = {
     ultraSmall: pxToRem('10px'),
   },
 
+  relativeFontSizes: {
+    // design spec: 10px (ultrasmall) in beta flag if base font is 16px
+    badgeFontToSurroundingInline: '62.5%',
+  },
+
   fontWeights: {
     regular: '400',
     medium: '500',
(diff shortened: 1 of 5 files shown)


packages/ui-kit/src/design-system.js
```diff
@@ -134,6 +134,11 @@ export const typography = {
     ultraSmall: pxToRem('10px'),
   },
 
+  relativeFontSizes: {
+    // design spec: 10px (ultrasmall) in beta flag if base font is 16px
+    badgeFontToSurroundingInline: '62.5%',
```
davifantasia, line 139: This is clever.

emmenko: Hmm, you should use `em` which is relative to the parent size.

emmenko:
I pushed a commit to use `em`. It should be good now.

PR #384, opened 2020-04-21
fix(theme): sticky page nav and scrolling container

emmenko:
In #347 we attempted to fix the container `height` in order for the page to be printable.
However, the media query (including the `percy` agent) didn't really work, resulting in the `height: 100vh` to never be applied and thus the page nav on the right to not be sticky.

This PR removes the `percy` agent media query and uses instead a normal CSS selector in order for Percy to take full page snapshots.

It also fixes the page content to not scroll when the sidebar menu and top menu are open (with the overlay).

I also added a Cypress test with the intent of preventing regressions for the sticky behavior.

### DIFF
--- a/cypress/integration/docs-smoke-test/viewports.js
+++ b/cypress/integration/docs-smoke-test/viewports.js
@@ -10,7 +10,9 @@ describe('Viewports', () => {
       widths: [512, 956],
     });
   });
-  it('should take snapshots of sticky page navigation', () => {
+  // Currently snapshots with scroll postion are not supported by Percy.
+  // https://github.com/percy/percy-cypress/issues/2#issuecomment-533316240
+  it.skip('should take snapshots of sticky page navigation', () => {
     cy.visit(URL_DOCS_SMOKE_TEST);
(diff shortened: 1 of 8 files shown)


cypress/integration/docs-smoke-test/viewports.js
```diff
@@ -10,4 +10,13 @@ describe('Viewports', () => {
       widths: [512, 956],
     });
   });
+  it('should take snapshots of sticky page navigation', () => {
+    cy.visit(URL_DOCS_SMOKE_TEST);
+    cy.findByLabelText('Page Table of Contents Navigation').within(() => {
+      cy.findByText('Last section').click();
+      cy.percySnapshot(cy.state('runnable').fullTitle(), {
+        percyCSS: `#application { height: 100vh !important; }`,
```
emmenko, line 18: We need to keep the original "scrolling behavior" so that the right navigation is sticky.

Hopefully this will work.

emmenko:
Hmm this is not what I was hoping for... 🤔

PR #388, opened 2020-04-21
fix(theme-docs): Render mdx commonmark compliant

nkuehn:
Since all of our processing and especially prettier is following the commonmark spec the site needs to understand backslashes as newlines.

This requires passing the commonmark option to MDX which is passed down to all remark plugins by mdx.


emmenko:
Snapshots approved

PR #390, opened 2020-04-22
Support custom anchors in MD(X) using <Achor name="my-anchor" />

### LINKED ISSUE #337: Support custom anchors in MD(X) using <Achor name="my-anchor" />
*(formerly called "Support custom anchors in MD(X) using {#custom-id}")*

Create a custom component that's always available in MDX:

`<Anchor name="any-html-id-compatible-string" />`

Outputting:

`<a name="any-html-id-compatible-string" />`  

Why not `<span id="any-html-id-compatible-string" />`?  **HTML5 Spec says that ID  precedes over anchor names, so using the anchor name makes sure that user content cannot break / override anchor names that are fixed in the implementation of the site.**

The Anchor component must never have content,  even if the authors use it as  `<Anchor name="any-html-id-compatible-string" >I must not put text here</Anchor>` that must still output only the self-closing HTML tag with no content.   This makes sure it's not confused with Links and the CSS does not apply or conflict. 

If CSS conflicts come up,  the Anchor component can be `visibility:hidden`.  Let's try. 

### OLD CONTENT ---

The existing content base needs custom anchor IDs in headings and block elements like list items. 
It uses a custom liquid plugin to generate them.  Use Cases:
 * wanting to link e.g. to a specific error code in a list of codes. (that's the mass of the cases)
 * a page has two headings with identical text the anchor IDs are becoming unreliable because generating it from the heading text is not unique. 
  * re-naming a heading without having to go through houndreds of existing links pointing there.  

There is no official (commonmark) specification for it but a lot of discussions and voting in the spec site. 
 * Some favor this notation: `### My Great Heading {#custom-id}`  
 * But for consistency with custom IDs on list items, code blocks etc, having the custom ID block in front is favored by others:  
  * `### {#custom-id} My Great Heading`  -> `<h3 id="custom-id">My Great Heading</h3>`
  * `   * {#custom-id} My Great List Item` -> `<li id="custom-id">My Great List Item</li>`

e.g. https://github.com/imcuttle/remark-heading-id  is implementing it as a very simple remark plugin to take a look at. There are generic implementations to add any HTML-ish attributes beyond the ID in many places of the markdown but I am not really a fan of allowing that  - I prefer explicitly handling the elements where it's allowed and restricting it to #id attribution. 

The custom syntax in the jekyll site has two variations that are just historically different and both output is rather a workaround because it's not working against a proper AST of the markdown, so extra elements (span or a) are generated instead of just adding an ID to the current block scope. 
 * ` * {% anchor_link productvariantimportdraft-variantId %} fooBar` -> generates `<a href='##{@text}' name='#{@text}'></a>`
 * ` * {% id_anchor general-400-invalid-field %} fooBar` --> generates `<span id='#{@text}'></span>`

davifantasia:
Closes #337 

**Note: Changes should also reflected in ReadMe.**

---
*Everything below is no longer valid since the requirement changed*

This feature supports custom tags for the following elements:

* Headings
* Lists
* Thematic breaks

### Usage
Include `{#custom-id}` before the text of any of the previously mentioned elements, and an id will be generated for the element.

| Expected Markdown Element   | Expected Rendered Element     |
| :------------- | :---------- |
| `### {#custom-id} My Great Heading` | `<h3 id="custom-id">My Great Heading</h3>`  | 
| `* {#custom-id} My Great List Item`  | `<li id="custom-id">My Great List Item</li>` |
| `{#custom-id} ***` or  `{#custom-id} ---` or `{#custom-id} ___` | `<hr id="custom-id">` |


packages/gatsby-theme-docs/src/components/anchor.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+
+const A = styled.a``;
+
+const Anchor = (props) => <A name={props.name} />;
```
emmenko, line 7: Why do we need a React component when you can just use HTML/JSX in MDX?

```
<a id="foo" />
```

This is valid JSX 👆

davifantasia: I am not sure if @nkuehn has mentioned this to you but during my discussions with him, he has said something to the effect of the new docs tool will disallow html. Maybe discuss with him for details.

Also, one main requirement is that we only want to allow the `name` attribute to be used for this purpose.

websites/docs-smoke-test/src/data/navigation.yaml
```diff
@@ -19,6 +19,8 @@
       path: /components/images
     - title: Code Examples
       path: /components/code-examples
+    - title: Custom Anchor
+      path: /components/custom-anchor
```
emmenko, line 23: Nit: probably better to put those in the "Views" chapter, as they relate more to demoing the use cases (see links as an example)

packages/gatsby-theme-docs/src/components/anchor.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Anchor = (props) => <a name={props.name} />;
```
emmenko, line 4: Please add some comments/context to describe what this is used for.

packages/gatsby-theme-docs/src/components/anchor.js
emmenko: Hmm I see...

packages/gatsby-theme-docs/src/components/anchor.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * This component is exposed to be used by mdx authors. It returns an anchor with name, and only name attribute.
+ *
+ * See more usability info here - https://github.com/commercetools/commercetools-docs-kit/issues/337
+ *
+ * This component also exists since we don't allow html in authoring mdx documents.
+ */
```
emmenko, line 10: Thanks 👌

emmenko:
Snapshots approved ✅

packages/gatsby-theme-docs/README.md
```diff
@@ -148,6 +148,7 @@ The available JSX components are:
 - `<Warning>`: a notification message with warning colors
 - `<Error>`: a notification message with error colors
 - `<CodeExample file="example.js" title="JavaScript Code Sample" />`: loading a code block from a file in `src/code-examples/`, supports all parameters of the fenced code block.
+- `<Anchor>`: inserts a custom anchor on any part of the document, can be used with headers, lists, in paragraphs, etc, it is used for navigating to specific parts of the document.
```
nkuehn, line 151:
```suggestion
- `<Anchor>`: inserts a custom anchor on any part of the document, can be used with headers, lists, in paragraphs, etc, it is used for navigating to specific parts of the document that are not headings. Also useful when a document has multiple headings with the same text or when heading names change and old 3rd party links shall continue to work.  Cannot override ID generation of the site generator, this adds additional named anchors and IDs have precedence.   
```

websites/docs-smoke-test/src/content/views/custom-anchor.mdx
```diff
@@ -0,0 +1,114 @@
...
+<a name="any-html-id-compatible-string-expected-usage"></a>
+```
+
+## With Other Props
+
+[Anchor with other props links to the next line](#any-html-id-compatible-string-with-other-props)
+
+<Anchor name="any-html-id-compatible-string-with-other-props" id="any-html-id-compatible-string-with-other-props"/>
+
+### In mardown document
```
nkuehn, line 29:
```suggestion
### In markdown document
```

websites/docs-smoke-test/src/content/views/custom-anchor.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+```html
+<a name="any-html-id-compatible-string-expected-usage"></a>
+```
+
+## With Other Props
+
+[Anchor with other props links to the next line](#any-html-id-compatible-string-with-other-props)
+
+<Anchor name="any-html-id-compatible-string-with-other-props" id="any-html-id-compatible-string-with-other-props"/>
```
nkuehn, line 27: is it possible to disallow this or automatically reset / delete the ID?  The intended behavior of the tag is to explicitly _not_ allow overriding the ID generation, i.e. to only allow  named anchors that by HTML spec have lower precedence than IDs. 

I know that it can't generally be turned off but I'd like to avoid encouraging it especially on this tag.  So in any case it should not be on this page because this page is the de facto documentation of what you can do.

websites/docs-smoke-test/src/content/views/custom-anchor.mdx
```diff
@@ -0,0 +1,114 @@
...
+<a name="any-html-id-compatible-string-with-other-props"></a>
+```
+
+## With Children
+
+[Anchor with children links to the next line](#any-html-id-compatible-string-with-children)
+
+<Anchor name="any-html-id-compatible-string-with-children" >I must not put text here</Anchor>
+
+### In mardown document
```
nkuehn, line 50:
```suggestion
### In markdown document
```

websites/docs-smoke-test/src/content/views/custom-anchor.mdx
```diff
@@ -0,0 +1,114 @@
...
+* List 3
+* List 4
+* List 5
+* List 6
+* List 7
+* List 8
+* List 9
+* List 10
+* List 11
+* List 12
```
nkuehn, line 99:
```suggestion
```
let's not make the visual regression tests bigger than necessary.

PR #399, opened 2020-04-24
fix: inconsistent url and title or index nav

### LINKED ISSUE #392: Beta flags an other JSX in heading names break index nav
If a `<Beta />` or other JSX is added to a heading text, the index nav which is driven by gatsbyJS  (or the MDX plugin)  generates `the-heading-text-beta-` as the anchor to link to but our component implementation  generates  `the-heading-name`  as the target anchor. 

Most importantly they must be the same so the index nav works. 

If we have the choice, it's better to not have the `-beta-` as part of the anchor because that creates a ton of work when taking a feature out of beta (change all links pointing there).

davifantasia:
Fixes #392 

Since we use [rehype-slug](https://github.com/rehypejs/rehype-slug) to generate header IDs which are in turn used for header anchors, this change regenerates `tableOfContent` url using [github-slugger](https://github.com/Flet/github-slugger) which [github-slugger](https://github.com/Flet/github-slugger) uses to generate slugs. The change also includes stripping out JSX from titles of tableOfContent.

### Image of Problem

Screenshot 2020-04-24 at 09 51 52

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-node.js
+++ b/packages/gatsby-theme-docs/gatsby-node.js
@@ -54,6 +54,7 @@ const resolverPassthrough = ({
   typeName = 'Mdx',
   fieldName,
   resolveNode = identity,
+  processResult = identity,
 }) => async (source, args, context, info) => {
   const type = info.schema.getType(typeName);
   const mdxNode = context.nodeModel.getNodeById({
@@ -64,11 +65,7 @@ const resolverPassthrough = ({
     fieldName,
   });
 
-  if (fieldName === 'tableOfContents') {
-    return processTableOfContentFields(result);
-  }
-
-  return result;
+  return processResult(result);
 };
 
 exports.sourceNodes = ({ actions, schema }) => {
@@ -138,7 +135,10 @@ exports.sourceNodes = ({ actions, schema }) => {
               default: 6,
             },
           },
-          resolve: resolverPassthrough({ fieldName: 'tableOfContents' }),
+          resolve: resolverPassthrough({
+            fieldName: 'tableOfContents',
+            processResult: processTableOfContentFields,
+          }),
         },
       },
       interfaces: ['Node'],
--- a/packages/gatsby-theme-docs/gatsby-node.js
+++ b/packages/gatsby-theme-docs/gatsby-node.js
@@ -10,6 +10,7 @@ const crypto = require('crypto');
 const { createFilePath } = require('gatsby-source-filesystem');
 const { ContextReplacementPlugin } = require('webpack');
 const slugify = require('slugify');
+const processTableOfContentFields = require('./src/utils/process-table-of-content-fields');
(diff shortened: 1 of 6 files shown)


emmenko:
As for the link labels in the table of content, we should probably do it at the gatsby node level, or we can "strip" the label when we render it.

https://github.com/commercetools/commercetools-docs-kit/blob/0f08e36dd6cd902f24275761bb1da46ff42b9d29/packages/gatsby-theme-docs/src/layouts/internals/page-navigation.js#L106
Mapping over the table of contents is just unnecessary.

davifantasia:
> To be honest, I would just implement our own rehype-slug plugin. It's literally like 10 lines of code.
> Then we can add the "stripping HTML/JSX" logic there.
> 
> https://github.com/rehypejs/rehype-slug

I don't get what you mean - we already use rehype-slug. Even if we write our own, at what point do we use it?

emmenko:
We use it instead of the default one:

https://github.com/commercetools/commercetools-docs-kit/blob/0f08e36dd6cd902f24275761bb1da46ff42b9d29/packages/gatsby-theme-docs/gatsby-config.js#L128

It would basically be the same implementation but without the html/jsx part.
Isn't that what part of the issues is?

davifantasia:
> Isn't that what part of the issues is?

The problem is not the `rehype-slug` we configured, it's that mdx plugin generates different titles and urls in `tableOfContent`.

emmenko:
I guess I got confused from what you said:

> Since we use rehype-slug to generate header IDs which are in turn used for header anchors

So is it only about the table of contents links? Can we maybe try to "fix" that on the node level?

https://github.com/commercetools/commercetools-docs-kit/blob/0f08e36dd6cd902f24275761bb1da46ff42b9d29/packages/gatsby-theme-docs/gatsby-node.js#L127-L136

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -62,6 +63,11 @@ const resolverPassthrough = ({
   const result = await field.resolve(resolveNode(mdxNode), args, context, {
     fieldName,
   });
+
+  if (fieldName === 'tableOfContents') {
+    return processTableOfContentFields(result);
+  }
```
emmenko, line 69: Nitpick: the `resolverPassthrough` is a generic function, we shouldn't put specific logic into it.

Instead we can add a new option `processResult` that can be passed on demand.

```js
const resolverPassthrough = ({
  typeName = 'Mdx',
  fieldName,
  resolveNode = identity,
  processResult = identity,
}) => async (source, args, context, info) => {
  // ...
  return processResult(result);
}
```

Then in the `tableOfContents` schema resolver, you use it like

```js
resolve: resolverPassthrough({
  fieldName: 'tableOfContents',
  processResult: processTableOfContentFields,
})
```

PR #400, opened 2020-04-24
feat(theme): allow to override top-menu links

### LINKED ISSUE #396: Configurable Top Menu and Footer content (for migration)
As already discussed at the beginning of the kit implementation,  we are now nearing go live and it turns out we need to be able to configure the top nav and footer links - or at least toggle between "current" and "next" in the site options. 

Reasons:
 * I realized that it's very hard to get QA from externals if they get lost in navigating to the old site
 * Enabling link testing an generally full automated QA is running circles between excluding too much and then having issues there.  

We did discuss live download from a central location, that's an option but i would already be very happy with being able to pull the data from a file or a file in another package.

emmenko:
Fixes #396 

image

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -21,6 +21,7 @@ const defaultOptions = {
   excludeFromSearchIndex: true,
   createNodeSlug: undefined,
   additionalPrismLanguages: [],
+  overrideTopMenuLinks: false,
 };
 const requiredOptions = ['websiteKey'];
 
@@ -71,6 +72,7 @@ module.exports = (themeOptions = {}) => {
         options: {
           name: 'internalConfigurationData',
           path: path.join(__dirname, `./src/data`),
+          ignore: pluginOptions.overrideTopMenuLinks ? ['**/top-*'] : [],
         },
       },
       // Assets (e.g. images) used from the markdown pages
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -21,7 +21,7 @@ const defaultOptions = {
   excludeFromSearchIndex: true,
   createNodeSlug: undefined,
   additionalPrismLanguages: [],
-  overrideTopMenuLinks: false,
+  overrideDefaultConfigurationData: [],
 };
 const requiredOptions = ['websiteKey'];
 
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -71,6 +72,7 @@ module.exports = (themeOptions = {}) => {
         options: {
           name: 'internalConfigurationData',
           path: path.join(__dirname, `./src/data`),
+          ignore: pluginOptions.overrideTopMenuLinks ? ['**/top-*'] : [],
```
emmenko, line 75: This will basically ignore the `data/top*.yaml` files from the theme, thus allowing the files from the website `src/data/top*.yaml` to be solely used.

nkuehn:
@emmenko  thanks for the proposal.  It would work but I think a more generic flag that allows to override any of the default config would be more future-proof.  It's more a random artifact that  today I don't need to change the footer and meta links. 

What do you think about my alternative approach in #401   ?

emmenko:
If you do it like that, it means that you have to copy/paste and override every file in the `data` folder.

Is that what you really want?

Another approach would to pass the list of globs of the files that you want to ignore, so you have more control of you want but you can just as easily override every file.

nkuehn:
Yes, overriding the complete folder is actually more practical to me because to later make that "next" version of all the nav links the default version I just have to copy / paste that whole data folder to the theme without running risk of having lost something or have transferred it the wrong way. 

I already did the changes and 80% of the entries are affected - practically the complete nav.

emmenko:
My main concern here is to add a feature that we only need in the migration phase.
If that’s the case, I would just add it as a “hidden” feature so that we can remove it at some point without introducing a breaking change.

If we want to have something like this as a general thing, I would make it so that it‘s a sustainable solution, for example passing a list of glob files to ignore.
For instance, assuming that the docs are migrated and being developed in a normal way, what would be the chance of changing the complete navigation? Probably it would just be a bunch of links from time to time.

nkuehn:
FYI @emmenko  this is now the last blocker to get the migration set up automated including testing the links. 

Allowing to directly control the ignore option of the filesystem plugin sounds best to me, it's also best documented and specified.   I'll try it here.

nkuehn:
@emmenko  I just pushed the glob variant you proposed.  

There may be temporary hacks needed in the footer too (e.g. legal wants some imprint link changed ad hoc) and I expect more default configuration to come over time.

emmenko:
Thanks, this looks good to me now

PR #402, opened 2020-04-24
feat: add gatsby theme to work with constant values

### LINKED ISSUE #397: Component to place a data point (here: an API limit) into a page
The platform has various limits (mostly numeric data points, some regular expressions, some text) that are mentioned in many places in the documentation and also have a central listing in the API contract page. 

The existing internal implementation uses jekyll include tags  `{{ site.data.api-limits.foo }}` to place the value into arbitrary places in the documentation while remaining to be able to centrally change them.  The pattern has proven successful and it's useful to have one data file with the limits, for example to impose stricter 4-eyes priciple in changes and to be sure you haven't missed something somewhere. 

To have  a consistent syntax for authors but still retain the ability to have them in a central place and also be able to list them I can imagine the following **Idea**:

 * a data file `/src/data/limits.yaml`
 * a component that has the data available (staticquery hook pattern?) and can be placed anywhere.   `<Limit key="my-new-platform-limit" />`.  It outputs a plain `<span>16 MB</span>` or a visible error if the name cannot be found. 

YAML structure example (the current one is just a key-value map but that's not gatsby-source-yaml compatible anyways. 

```yaml
- key: my-new-platform-limit
  number: 16
  text: MB
```
- we need separate properties for numeric and text properties to be able to get it into the gatsbyJS graphQL. If both are set, number goes before text so text can be used for a unit

It's a very specific use case but if we want to be able to e.g. render a full table of limits we need to have a domain specific data file. 

Any feedback? Better ideas how to handle limits?

davifantasia:
Closes #397


packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -109,6 +109,12 @@ exports.sourceNodes = ({ actions, schema }) => {
       fix
       enhancement
     }
+
+    type LimitsYaml implements Node @dontInfer {
+      id: ID!
+      name: String!
+      value: String!
```
nkuehn, line 116: This is not the schema proposed in the issue (separate numeric and text keys there so they can be combined and we can apply number formatting as part of the UI implementation)

websites/docs-smoke-test/src/content/index.mdx
```diff
@@ -190,6 +190,10 @@ or taking the full width:
 
 ---
 
+## Limits
+
+This is an example of a limit -> <Limit name="numberLimit"/>
```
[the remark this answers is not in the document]
davifantasia: > can you use a test case where the limit has both a number and a text value?

Please give an example of what a limit with both number and text would look like.

packages/gatsby-theme-docs/gatsby-node.js
davifantasia: > This is not the schema proposed in the issue (separate numeric and text keys there so they can be combined and we can apply number formatting as part of the UI implementation)

I guess this is related to the previous comment so I'll just wait for your response on that one.

packages/gatsby-theme-docs/src/data/limits.yaml
```diff
@@ -0,0 +1,6 @@
+- name: numberLimit
```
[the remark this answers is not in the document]
davifantasia: > Please don't commit actual example data into the theme package, they will appear in all projects using it.

I totally missed that, thanks.

emmenko:
Q: are those "limits" functionalities something that we need for all websites, or is it more specific to the API docs?

packages/gatsby-theme-docs/src/components/limit.js
```diff
@@ -0,0 +1,26 @@
...
+    const message = `${props.name} limit is not found.`;
+
+    if (process.env.NODE_ENV !== 'production') {
+      return <ContentNotifications.Error>{message}</ContentNotifications.Error>;
+    }
+
+    throw new Error(message);
+  }
+
+  return <>{`${limit.number || ''} ${limit.text || ''}`.trim()}</>;
```
nkuehn, line 19: nitpick: Ideally there should be a `&nbsp;` between number and text if both are set. 
```suggestion
  return <>{`${limit.number || ''}${limit.text && limit.number ? "&nbsp;" : ''}${limit.text || ''}`.trim()}</>;
```
(code suggestion not tested, just typed in here

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -68,7 +68,7 @@ const resolverPassthrough = ({
   return processResult(result);
 };
 
-exports.sourceNodes = ({ actions, schema }) => {
+exports.createSchemaCustomization = ({ actions, schema }) => {
```
emmenko, line 71: Did we use it wrong this entire time?

Thanks for fixing it.

packages/gatsby-theme-docs/src/components/limit.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ContentNotifications } from '@commercetools-docs/ui-kit';
+import { useLimitByName } from '../hooks/use-limits';
+
+const Limit = (props) => {
```
emmenko, line 6: I'm struggling a bit with the naming here. What does limit imply? Can we find a more specific name?

packages/gatsby-theme-docs/src/components/limit.js
```diff
@@ -0,0 +1,32 @@
...
+
+    if (process.env.NODE_ENV !== 'production') {
+      return <ContentNotifications.Error>{message}</ContentNotifications.Error>;
+    }
+
+    throw new Error(message);
+  }
+
+  return (
+    <>
```
emmenko, line 20: Should we use a `span` here?

packages/gatsby-theme-docs/src/components/limit.js
```diff
@@ -0,0 +1,32 @@
...
+      {limit.number || ''}
+      {limit.text && limit.number ? <>&nbsp;</> : ''}
+      {limit.text || ''}
```
emmenko, line 23:
```suggestion
      {[limit.number, limit.text].filter(Boolean).join()}
```

websites/docs-smoke-test/src/data/limits.yaml
```diff
@@ -0,0 +1,7 @@
+- name: numberAndText
+  number: 60
+  text: megabytes
+- name: number
+  number: 7
+- name: text
+  text: "[a-zA-Z0-9_\\-]{2,256}"
```
emmenko, line 7: Can we name them properly with some "real" case examples?

packages/gatsby-theme-docs/src/components/limit.js
davifantasia: The name is an authoring thing, please speak with @nkuehn about that.

packages/gatsby-theme-docs/src/components/limit.js
davifantasia: I don't think it's necessary.

packages/gatsby-theme-docs/src/components/limit.js
davifantasia: > {[limit.number, limit.text].filter(Boolean).join('&nbsp;')}

Did you test that?

packages/gatsby-theme-docs/src/components/limit.js
nkuehn: Not now since we don't have any special styling instruction.  

But I can imagine that we later add more metadata to a limit (e.g. a label or a topic), generate a central overview table and have the component link to it.   But to start off and get the migration moving just text is fine.

nkuehn:
@emmenko  no, it's not API reference specific. Any piece of Software has limits and defaults that are key constants to know and that are better managed as data in documentation.   The limits are central to make sure that they are always up to date in tutorials, FAQ and other non-API-reference content, oo.  

The more interesting question is that the data is not website-specific.  In the commercetools-docs repo I would maintain the limits.yaml in the central theme wrapper that the setup neeeds anyways for other central data.

websites/docs-smoke-test/src/data/limits.yaml
nkuehn: @emmenko  this is a real example.

packages/gatsby-theme-docs/src/components/limit.js
nkuehn: Please - let's not make a code beauty contest out of this.  Thanks!

websites/docs-smoke-test/src/data/limits.yaml
emmenko: The **names**, not the values.

emmenko:
Yeah, I'm trying to understand if this needs to live "inside" the core theme, or can be a gatsby plugin, or part of the wrapper theme or something.
We have the ability to compose things, let's use that if and where we can.

emmenko:
Can we clarify my last comment before we merge this? I want to avoid merging this and then removing it again because we decided to move this functionality somewhere else.

davifantasia:
> Can we clarify my last comment before we merge this? I want to avoid merging this and then removing it again because we decided to move this functionality somewhere else.

If this is directed to me, I think @nkuehn indirectly gave a response to your comment [here](https://github.com/commercetools/commercetools-docs-kit/pull/402#issuecomment-620601206). And from what I understood, @nkuehn is saying that yes, the `Limit` component belongs in the core theme.

websites/docs-smoke-test/src/data/limits.yaml
nkuehn: right, the names would be better with real examples - I was looking at the result in the site an was initially irritated by the value myself

emmenko:
I pushed a commit to extract the data limits stuff into a gatsby transformer plugin.

emmenko:
Makes sense. I'll update the PR.

emmenko:
Ah shit, using multiple themes that shadow the same component results in only the component of the "last" theme being used. 🤯

I'll try to think of another solution

emmenko:
FYI: this is currently blocked by #422

nkuehn:
> FYI: this is currently blocked by #422

After merging your new architecture I now updated this to the new API and adjusted the documentation.  Builds are failing though due to the vale binary not downloading which seems completely unrelated to me.  Need to look at that tomorrow.

PR #407, opened 2020-04-27
fix(theme): render proper 404 page as a content page

### LINKED ISSUE #403: 404 layout broken
The layout of the gatsbyJS 404 page is broken,  in some responsive breakpoints it at least says "page not found" but in the main view it's completely broken. 

Please note:
 * this refers to production builds
 * the gatsbyjs 404 does not appear when directly entering a wrong URL.  But when navigating via back / forward the gatsbyJS client side 404 appears. 
 * ideally we also configure `now.json` to a proper 404.   Let's get it first set up correct in this repo and then do it in the private content repos. 

image

emmenko:
Fixes #403 

image


now.json
```diff
@@ -13,6 +13,11 @@
     },
     {
       "handle": "filesystem"
+    },
+    {
+      "src": "/([a-z]+)/.*",
+      "status": 404,
+      "dest": "/$1/404"
```
emmenko, line 20: This should do the trick for the ZEIT routing.

nkuehn: In my migration branch I opted to always point to the 404 of the `/docs` homepage path ( `/docs/404` ) - that also catches   paths like  `/12`  without showing the ZEIT 404.  

If we think there's a use case for microsite-specific 404s the approaches can be combined, too - add another route behind yours that points all else to  `/docs/404`

nkuehn: PS: obviously, here in the kit repo this is all not a problem.

now.json
```diff
@@ -13,6 +13,11 @@
     },
     {
       "handle": "filesystem"
+    },
+    {
+      "src": "/([a-z]+)/.*",
```
nkuehn, line 18: this does not catch our website names with dashes. Plus, you need to escape slashes.   

test case: https://commercetools-docs-kit-git-nm-404.commercetools.now.sh/docs-smoke-test/i-do-not-exist

```suggestion
      "src": "\/(\S+?)\/.*",
```
(any non-whitespace character, ungreedily matched until the next forward slash)

https://regex101.com/r/3ntqSz/1

emmenko: Thanks 😄

now.json
emmenko: Yeah, if we have a "home" microsite we can do that as well.

now.json
```diff
@@ -13,6 +13,11 @@
     },
     {
       "handle": "filesystem"
+    },
+    {
+      "src": "\/(\S+?)\/.*",
```
nkuehn, line 18:
```suggestion
      "src": ""\\\/(\\S+?)\\\/.*"",
```
JSON escaping, no questions...  That's what happens if you try to code in github PR comments :-(

emmenko: Escaping the slash should not be necessary

PR #408, opened 2020-04-28
feat: global language selection for code examples

### LINKED ISSUE #320: Global language selection for code examples grouped with MultiCodeExample
## Expected
When a language is selected from a dropdown in a rendered instance of `MultiCodeExample`, all instances of `MultiCodeExample` should use the selected language as the default.

This change is expected to affect `MultiCodeExample` instances on same page and on other parts of the website.

## Fallback
If preselected language does not exist for `MultiCodeExample`, then the default language is defined by the order as written by the author, that is, the language from the first `CodeExample` in the `MultiCodeExample` component.

davifantasia:
Closes #320

### DIFF
--- a/packages/gatsby-theme-docs/src/components/multi-code-example.js
+++ b/packages/gatsby-theme-docs/src/components/multi-code-example.js
@@ -12,12 +12,17 @@ function MultiCodeExample(props) {
   const codeExamples = useCodeExamples();
   const context = useCodeExampleLanguageContext();
 
+  function handleOnLanguageChange(newLanguage) {
+    localStorage.setItem('codeExampleLanguage', newLanguage);
+    context.setCodeExampleLanguage(newLanguage);
+  }
+
   try {
     return (
       <MultiCodeBlock
         title={props.title}
         selectedLanguage={context.codeExampleLanguage}
-        onLanguageChange={context.setCodeExampleLanguage}
+        onLanguageChange={handleOnLanguageChange}
       >
         {React.Children.map(props.children, (child, index) => {
           if (!child.props || child.props.mdxType !== 'CodeExample') {
--- a/packages/gatsby-theme-docs/src/components/multi-code-example.js
+++ b/packages/gatsby-theme-docs/src/components/multi-code-example.js
@@ -5,56 +5,72 @@ import {
   MultiCodeBlock,
   CodeBlock,
 } from '@commercetools-docs/ui-kit';
+import CodeExampleLanguageContext from '../hooks/use-code-example-language';
 import useCodeExamples from '../hooks/use-code-examples';
 
 function MultiCodeExample(props) {
   const codeExamples = useCodeExamples();
 
-  try {
-    return (
-      <MultiCodeBlock title={props.title}>
-        {React.Children.map(props.children, (child, index) => {
-          if (!child.props || child.props.mdxType !== 'CodeExample') {
-            throw new Error(
-              `Children of <MultiLanguageCodeExamples> must be a <CodeExample> component and not "${
-                child.props ? child.props.mdxType : child
-              }"`
-            );
-          }
-          if (!child.props.path) {
-            throw new Error(
-              `The component <CodeExample> is missing the required property "path".`
-            );
-          }
-          const codeExample = codeExamples.find((example) => {
-            return example.path === child.props.path;
-          });
-          if (!codeExample) {
-            throw new Error(
-              `The provided path "${child.props.path}" does not exist. Please make sure that the code example file exists at the provided path.`
-            );
-          }
+  return (
+    <CodeExampleLanguageContext.Consumer>
+      {({ language, setLanguage }) => {
+        function handleOnLanguageChange(lang) {
+          setLanguage(lang);
+        }
+        try {
           return (
-            <CodeBlock
-              key={`${codeExample.language}-${index}`}
-              content={codeExample.content}
-              language={codeExample.language}
-              highlightLines={child.props.highlightLines || []}
-              noPromptLines={child.props.noPromptLines || []}
-            />
+            <MultiCodeBlock
+              title={props.title}
+              selectedLanguage={language}
+              onLanguageChange={handleOnLanguageChange}
+            >
+              {React.Children.map(props.children, (child, index) => {
+                if (!child.props || child.props.mdxType !== 'CodeExample') {
+                  throw new Error(
+                    `Children of <MultiLanguageCodeExamples> must be a <CodeExample> component and not "${
+                      child.props ? child.props.mdxType : child
+                    }"`
+                  );
+                }
+                if (!child.props.path) {
+                  throw new Error(
+                    `The component <CodeExample> is missing the required property "path".`
+                  );
+                }
+                const codeExample = codeExamples.find((example) => {
+                  return example.path === child.props.path;
+                });
+                if (!codeExample) {
+                  throw new Error(
+                    `The provided path "${child.props.path}" does not exist. Please make sure that the code example file exists at the provided path.`
+                  );
+                }
+                return (
+                  <CodeBlock
+                    key={`${codeExample.language}-${index}`}
+                    content={codeExample.content}
+                    language={codeExample.language}
+                    highlightLines={child.props.highlightLines || []}
+                    noPromptLines={child.props.noPromptLines || []}
+                  />
+                );
+              })}
+            </MultiCodeBlock>
           );
-        })}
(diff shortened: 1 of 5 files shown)


packages/gatsby-theme-docs/src/hooks/use-code-example-language.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+const CodeExampleLanguageContext = React.createContext({
+  language: '',
+  setLanguage: () => {},
+});
+CodeExampleLanguageContext.displayName = 'CodeExampleLanguageContext';
```
emmenko, line 7: This is wrong, the context is not a component

packages/gatsby-theme-docs/src/components/multi-code-example.js
```diff
@@ -5,56 +5,72 @@ import {
...
-          const codeExample = codeExamples.find((example) => {
-            return example.path === child.props.path;
-          });
-          if (!codeExample) {
-            throw new Error(
-              `The provided path "${child.props.path}" does not exist. Please make sure that the code example file exists at the provided path.`
-            );
-          }
+  return (
+    <CodeExampleLanguageContext.Consumer>
```
emmenko, line 15: Use the hook:  `React.useContext(CodeExampleLanguageContext)`

packages/gatsby-theme-docs/src/components/theme-provider.js
```diff
@@ -4,9 +4,12 @@ import { useStaticQuery, graphql } from 'gatsby';
 import { Reset, Globals } from '@commercetools-docs/ui-kit';
 import { SiteDataContext } from '../hooks/use-site-data';
 import useAdditionalSiteData from '../overrides/use-additional-site-data';
+import CodeExampleLanguageContext from '../hooks/use-code-example-language';
 import ErrorBoundary from './error-boundary';
 
 const ThemeProvider = (props) => {
+  const [language, setLanguage] = React.useState('test');
```
emmenko, line 11: The name is too generic and it can be confused with "language/locale".
We should name it like `codeExampleLanguage`

packages/gatsby-theme-docs/src/components/multi-code-example.js
```diff
@@ -5,56 +5,72 @@ import {
...
+      {({ language, setLanguage }) => {
+        function handleOnLanguageChange(lang) {
+          setLanguage(lang);
+        }
```
emmenko, line 19: You don't need this, just pass the `setLanguage` as a prop

packages/gatsby-theme-docs/src/components/theme-provider.js
emmenko: Also remember to remove `test`

```suggestion
  const [language, setLanguage] = React.useState('');
```

emmenko: We should probably make the default language configurable for each website, as a theme option.

packages/gatsby-theme-docs/src/components/multi-code-example.js
```diff
@@ -5,56 +5,72 @@ import {
...
-              key={`${codeExample.language}-${index}`}
-              content={codeExample.content}
-              language={codeExample.language}
-              highlightLines={child.props.highlightLines || []}
-              noPromptLines={child.props.noPromptLines || []}
-            />
+            <MultiCodeBlock
+              title={props.title}
+              selectedLanguage={language}
+              onLanguageChange={handleOnLanguageChange}
```
emmenko, line 25: I would just move the hook into the `MultiCodeBlock` component, there is no need for this extra wrapping.

davifantasia: Please explain how that works.

packages/gatsby-theme-docs/src/hooks/use-code-example-language.js
davifantasia: You mean you prefer only components to have names?

If it is wrong, why is `displayName` a field you can set?

emmenko: This is JS, you can attach any field you want. The `displayName` is a static property used by the React devtools, for components.

packages/gatsby-theme-docs/src/components/multi-code-example.js
emmenko: Remove the wrapper with the consumer component, use `React.useContext` within the `MultiCodeBlock` component.

Not sure what else is not clear here.

packages/gatsby-theme-docs/src/hooks/use-code-example-language.js
davifantasia: > The `displayName` is a static property used by the React devtools, for components.

Sounds like a React Context is in fact a component.

emmenko: `Context.Provider` and `Context.Consumer` are components, not `Context`.

image

image

https://reactjs.org/docs/context.html#api

packages/ui-kit/src/components/multi-code-block.js
```diff
@@ -88,11 +88,34 @@ const languageDisplayNames = {
...
+
+      // will set selected language if
+      // - the language exists for this instance
+      // - this is not the first render.
+      // - this is not the instance that set the language
+      if (langIndex > -1 && selected !== props.selectedLanguage) {
+        setSelected(props.selectedLanguage);
+      }
+    }
+  }, [props.selectedLanguage, langs, selected]);
```
emmenko, line 114: Hmm I feel like this code-example <-> context setup is odd.

Ideally, the context should be initialized with the "selected" language. Then the `MultiCodeBlock` component simply reads from the context and updates the value only when the user selects the language.

Here we're trying to "shoot ourselves in the foot" by potentially running into an infinite loop, if we're not careful with state and the effect.

I'll have to think about it if there is a more safer approach.

packages/gatsby-theme-docs/src/components/theme-provider.js
```diff
@@ -4,9 +4,14 @@ import { useStaticQuery, graphql } from 'gatsby';
 import { Reset, Globals } from '@commercetools-docs/ui-kit';
 import { SiteDataContext } from '../hooks/use-site-data';
 import useAdditionalSiteData from '../overrides/use-additional-site-data';
+import CodeExampleLanguageContext from '../hooks/use-code-example-language';
 import ErrorBoundary from './error-boundary';
 
 const ThemeProvider = (props) => {
+  const [codeExampleLanguage, setCodeExampleLanguage] = React.useState(
+    localStorage.getItem('codeExampleLanguage')
```
emmenko, line 12: This doesn't work in SSR.

packages/ui-kit/src/components/multi-code-block.js
emmenko: (_talking out lout for myself_)

However, because of SSR, we can't know the value of the selected language (from localstorage) when the static pages are created.
So by default, the first language in the list will be always be rendered.
Then, the context gets initialized with the localstorage value.
At this point the component re-renders and the language context may contain a value. If the value is defined, the selected language should reflect the value from the context, but only if the language is available within the rendered code examples.

This is kind of tricky.

For instance, let's say we have code examples in 3 languages: java, javascript, php.
I would assume that ALL multi code examples should contain examples for each language, otherwise it would be weird if all examples on a page switch to php but some of them don't have an associated example file.
What should happen in this case? Should we fall back to one of the other examples or should we render a "blank" example?

Then maybe we need to fix the "multi code examples setup" first. For example, we need to specify which languages are supposed to be rendered, and that should be the same for all of them.

cc @nkuehn ?

packages/gatsby-theme-docs/src/components/theme-provider.js
```diff
@@ -29,9 +36,16 @@ const ThemeProvider = (props) => {
...
       <SiteDataContext.Provider value={siteData}>
-        <Reset />
-        <Globals />
-        {props.children}
+        <CodeExampleLanguageContext.Provider
+          value={{
+            codeExampleLanguage,
+            setCodeExampleLanguage,
+          }}
+        >
```
emmenko, line 44: FYI: now that the code-examples are an add-on, we need to find another solution for this. I would suggest maybe to re-open a new PR with a fresh approach.

davifantasia:
Multi-language samples generated from rmf-codegen won't be available in the near future. So we'll close this for now.

Issue #226, opened 2020-01-23
Rename design tokens colors to provide more meaningful context for usage


emmenko:
As the list of design tokens keeps growing, the current token names (especially for colors) are not super useful anymore.

I'd like to go over them (when I have some time) and provide more meaningful names.

Example of "good" names:

https://github.com/commercetools/commercetools-docs-kit/blob/d62f3b5a22e25023f704a5711f178b9bb0f6370a/packages/ui-kit/src/design-system.js#L58-L68

PR #411, opened 2020-04-28
feat(theme): allow to configure website primary color

### LINKED ISSUE #409: Theme: Configurable key color per website
Currently all the websites get the key color of the custom applications site because the second phase of designs was built on custom applications as the example content. 

But (@Shecki please confirm) the design concept is that there are multiple key colors and that in a "normal" platform documentation site the key color is the same green as the links (see the original design of the HTTP API site). 

To be decided:  
  a) Wrap it into a more general "style variation" setting that just allows to tell whether it's  e.g.  merchant center vs. api platform vs. open source offerings vs. custom applications  (or whatever separation we want.  
  b) explicitly allow to pass a color key (e.g. a named design token).

emmenko:
Closes #409

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -1,7 +1,7 @@
 /* eslint-disable global-require */
 
 const path = require('path');
-const { customProperties } = require('@commercetools-uikit/design-system');
+const defaultOptions = require('./default-options');
 
 // Proxy env variables needed for `gatsby-browser.js` and `gatsby-ssr.js`.
 // https://www.gatsbyjs.org/docs/environment-variables/#client-side-javascript
@@ -15,18 +15,7 @@ proxyEnvironmentVariables.forEach((envName) => {
   process.env[`GATSBY_${envName}`] = process.env[envName];
 });
 
-const defaultOptions = {
-  websiteKey: '',
-  websitePrimaryColor: customProperties.colorPrimary,
-  beta: false,
-  gaTrackingId: undefined,
-  excludeFromSearchIndex: true,
-  createNodeSlug: undefined,
-  additionalPrismLanguages: [],
-  overrideDefaultConfigurationData: [],
-};
 const requiredOptions = ['websiteKey'];
-
 const validateThemeOptions = (options) => {
   requiredOptions.forEach((option) => {
     if (!options[option]) {
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -1,6 +1,7 @@
 /* eslint-disable global-require */
 
 const path = require('path');
+const { customProperties } = require('@commercetools-uikit/design-system');
 
 // Proxy env variables needed for `gatsby-browser.js` and `gatsby-ssr.js`.
 // https://www.gatsbyjs.org/docs/environment-variables/#client-side-javascript
@@ -16,7 +17,7 @@ proxyEnvironmentVariables.forEach((envName) => {
 
 const defaultOptions = {
   websiteKey: '',
-  websitePrimaryColor: '',
+  websitePrimaryColor: customProperties.colorPrimary,
   beta: false,
   gaTrackingId: undefined,
   excludeFromSearchIndex: true,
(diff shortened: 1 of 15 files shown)


nkuehn:
Thanks, wow that was fast! 

I'm checking with design what the default color should be in the afternoon anyways, will come back with a review then.

packages/gatsby-theme-docs/src/layouts/internals/layout-application.js
```diff
@@ -36,12 +37,26 @@ const Container = styled.div`
...
 const LayoutApplication = (props) => (
-  <UiKitThemeProvider theme={designSystem.uikitTheme}>
+  <UiKitThemeProvider
+    theme={{
+      ...designSystem.uikitTheme,
+      // Docs specific theme properties
+      colors: {
+        light: {
+          primary:
+            props.websitePrimaryColor || designSystem.colors.light.primary,
```
nkuehn, line 47: https://github.com/commercetools/commercetools-docs-kit/blob/nm-website-primary-color/packages/ui-kit/src/design-system.js#L18  is not the default color sven posted in the issue ( #00B39E ).  Did you intend to not change it or is the intention to change it separately ( maybe during #226 ?).

emmenko: I guess the question is, should there be a default color or should we make the `websitePrimaryColor` option required? So that each website must specify the color.

@Shecki @nkuehn

Shecki: @nkuehn how many "colors" do we need in the beginning?
as a fallback, i think we should use the "key-Color"

it is not allowed to use "any-color" i would like to come up with a Selection of Colors.
First, i would use the most "prominent ones from our UI-KIT library.

emmenko: So should there be a default primary color in the theme?

- [ ] yes
- [ ] no

If so, is it the `#00B39E`?

Shecki: from my side, for the beginning yes.
Later i will look at all Products/Solutions and define the color.
As the colors have to be used in Banners/Graphics as well

emmenko: Ok thanks

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -16,6 +17,7 @@ proxyEnvironmentVariables.forEach((envName) => {
 
 const defaultOptions = {
   websiteKey: '',
+  websitePrimaryColor: customProperties.colorPrimary,
```
emmenko, line 20: The default primary color now is `#00B39E`

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -120,6 +120,7 @@ exports.sourceNodes = ({ actions, schema }) => {
         id: { type: 'ID!' },
         slug: { type: 'String!' },
         title: { type: 'String!' },
+        websitePrimaryColor: { type: 'String!' },
```
emmenko, line 123: In the schema/node the field is still required though.

packages/ui-kit/src/design-system.js
```diff
@@ -14,8 +14,6 @@ export const pxToRem = (px, suffix = 'rem') => {
 
 export const colors = {
   light: {
-    // Used to give the website a unique tone (e.g. page titles, navigation links)
-    primary: customProperties.colorAccent40,
```
emmenko, line 18: To avoid confusion, we should remove it from here and always pick it from the theme.

websites/api-docs-smoke-test/gatsby-config.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
       resolve: '@commercetools-docs/gatsby-theme-api-docs',
       options: {
         websiteKey: 'api-docs-smoke-test',
+        websitePrimaryColor: 'goldenrod',
```
emmenko, line 15: This is just a color example for the API smoke tests website.

emmenko:
We should be good to go with this now

emmenko:
So should I change the default color to black?

nkuehn:
@emmenko I don't mind both, black is more conservative as a default in case someone uses the kit outside the docs team.

PR #419, opened 2020-05-05
refactor(code-examples): make it a gatsby theme

emmenko:
~~The code-examples transformer package should/can contain also the UI components, so that we have all the logic encapsulated in one package.~~

See https://github.com/commercetools/commercetools-docs-kit/pull/419#pullrequestreview-405813014


packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -98,6 +98,14 @@ module.exports = (themeOptions = {}) => {
...
         },
       },
+      // Code examples
+      {
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'codeExamples',
+          path: path.resolve(`./src/code-examples`),
+        },
+      },
```
emmenko, line 108: All gatsby plugins do not ship with the `gatsby-config` included, so this needs to be defined by the theme/website itself.

packages/gatsby-theme-docs/src/markdown-components.js
```diff
@@ -4,12 +4,10 @@ import {
   ContentNotifications,
 } from '@commercetools-docs/ui-kit';
 import {
-  Link,
   CodeExample,
   MultiCodeExample,
-  BetaFlag as Beta,
-  Anchor,
-} from './components';
+} from '@commercetools-docs/gatsby-transformer-code-examples';
```
emmenko, line 9: We can import the components from the package.

emmenko:
Gatsby plugins can have component files as well. See for example: https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-google-analytics

The intent is to have all the related logic encapsulated into the package, so that you can potentially use it as a standalone package.

emmenko:
Alright, this is not a gatsby theme

emmenko:
@nkuehn is this good from your side as well?

emmenko:
Thanks!

Issue #363, opened 2020-04-14
Release Notes:  Build Filters


nkuehn:
Build the filter UI on the overview page 
  * not visible on the details
  * the overview page does not get a pager, i.e. all releases are there
  * to be decided  / analyzed: Statefulness.
     *  When navigating to the list through navigation (main menu, other links) the filters should be clear(ed)
     * When navigating back from a release notes details page the filters should remain set as they were

Behavior of the topic checkboxes:  OR, not AND

Design - https://commercetools.invisionapp.com/d/main?origin=v7#/console/19438556/408192223/preview?scrollOffset=215

> this can be delayed post launch if needed, it's not in the critical path to relauch!

PR #420, opened 2020-05-05
feat: release notes filters

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/363


davifantasia:
@emmenko is there a way to style the DateInput and CheckboxInput component from the ui-kit?

emmenko:
Why do you need to add custom styles?

packages/gatsby-theme-docs/src/components/release-notes-filter-topics.js
```diff
@@ -17,43 +18,64 @@ const ClearAll = styled.a`
...
+
+  const [topics, setTopics] = React.useState(checkedTopics);
 
   return (
     <Container>
       <SpacingsStack>
-        <SpacingsInline alignItems="center">
+        <SpacingsInline alignItems="center" justifyContent="space-between">
           <FilterTitle>Filter By Topics</FilterTitle>
           <ClearAll href="#" onClick={handleOnClearAll}>
```
emmenko, line 35: This is not a link, it should be a button.

packages/gatsby-theme-docs/src/components/theme-provider.js
```diff
@@ -27,13 +28,15 @@ const ThemeProvider = (props) => {
...
   };
   return (
-    <ErrorBoundary>
-      <SiteDataContext.Provider value={siteData}>
-        <Reset />
-        <Globals />
-        {props.children}
-      </SiteDataContext.Provider>
-    </ErrorBoundary>
+    <IntlProvider locale="en">
```
emmenko, line 31: For now, since we only have one use case, I would render this in the release notes filters only, not for the entire website.

packages/gatsby-theme-docs/src/hooks/use-all-release-notes.js
```diff
@@ -0,0 +1,40 @@
...
+          }
+        }
+      }
+    `
+  );
+
+  return queryResult.allReleaseNotePage.nodes;
+};
+
+export default useAllReleaseNotes;
```
emmenko, line 20: Use a named export for this as well, since you also have other exports (not just one default).

packages/gatsby-theme-docs/src/hooks/use-all-release-notes.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const useEarliestLatestDates = () => {
+  const allReleaseNotes = useAllReleaseNotes();
+  return {
+    latestDate: allReleaseNotes[0].date,
+    earliestDate: allReleaseNotes[allReleaseNotes.length - 1].date,
+  };
+};
+
+export const useReleaseNotesTopicsSet = () => {
```
emmenko, line 30:
```suggestion
export const useReleaseNotesTopics = () => {
```

Then return `return Array.from(topics)`

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -11,27 +11,68 @@ import markdownFragmentToReact from '../utils/markdown-fragment-to-react';
...
-          title={props.data.contentPage.title}
-          excludeFromSearchIndex={props.data.contentPage.excludeFromSearchIndex}
-        />
-        <MDXProvider components={markdownComponents}>
+const ReleaseNotesListTemplate = (props) => {
+  const [releaseNotes, setReleaseNotes] = React.useState(
+    props.data.allReleaseNotePage ? props.data.allReleaseNotePage.nodes : []
+  );
+  const [fromFilterDate, setFromFilterDate] = React.useState('');
+  const [toFilterDate, setToFilterDate] = React.useState('');
```
emmenko, line 19: Shouldn't there be some initial from/to dates? 🤔

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -11,27 +11,68 @@ import markdownFragmentToReact from '../utils/markdown-fragment-to-react';
...
+  ]);
+
+  return (
+    <ThemeProvider>
+      <LayoutReleaseNotesList
+        pageContext={props.pageContext}
+        pageData={props.data.contentPage}
+        handleOnFromFilterDateChange={setFromFilterDate}
+        handleOnToFilterDateChange={setToFilterDate}
+        handleOnFilterTopicsChange={setFilterTopics}
```
emmenko, line 59: "handle" is usually the function handling the callback. Here you want to pass an "on"-something prop. For instance, it's the same as `onClick`, `onChange`, etc.

```suggestion
        onFromFilterDateChange={setFromFilterDate}
        onToFilterDateChange={setToFilterDate}
        onFilterTopicsChange={setFilterTopics}
```

packages/gatsby-theme-docs/src/hooks/use-all-release-notes.js
```diff
@@ -0,0 +1,40 @@
...
+      }
+    `
+  );
+
+  return queryResult.allReleaseNotePage.nodes;
+};
+
+export default useAllReleaseNotes;
+
+export const useEarliestLatestDates = () => {
```
emmenko, line 22: Weird name 🤔

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
```diff
@@ -27,11 +31,9 @@ const ReleaseNotesFilterDates = () => {
       <SpacingsStack scale="xs">
         <DateLabel htmlFor="from-filter-date">From</DateLabel>
         <div>
-          <input
-            type="date"
+          <DateInput
```
emmenko, line 34: General question: since there are from/to date fields, is there any reason for not using the `DateRangeInput`?

https://uikit.commercetools.com/?path=/story/components-inputs--daterangeinput

davifantasia:
@emmenko please why did you include "WIP" label?

packages/gatsby-theme-docs/src/templates/release-notes-list.js
davifantasia: > Shouldn't there be some initial from/to dates? 🤔

That would mean the list would always be filtered, I don't see any benefit in that implementation.

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
davifantasia: See the [design](https://commercetools.invisionapp.com/d/main?origin=v7#/console/19438556/408192223/preview?scrollOffset=215)

emmenko: Thanks but that does not answer my question 😅

emmenko:
> why did you include "WIP" label?

So that in the PRs list it's more visually visible which PR is still a work in progress or not.

packages/gatsby-theme-docs/src/templates/release-notes-list.js
emmenko: I guess this is more of a general question to @nkuehn : if we have releases going back to 5 years ago, do we want to always show/render all of them?

I would image that we only want to show the more "recent" releases.
Not important for now, but seeing this popped up the question and we can discuss/look into this as a follow up.

emmenko:
@davifantasia FYI: I rebased the branch with `master` (and added a changeset for this PR). Make sure to pull the changes on the branch before continuing the work.

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
davifantasia: `DateRangeInput` does not meet the needs of the design implementation.

emmenko: Still not answering my question, which is not about the implementation 🙂

Let me rephrase that.

Given that that the filter is a "range" filter, what is the reason for not using a `DateRangeInput`? Is it just because of some design constraints or is the reason about usability?

davifantasia:
> * scroll to top of the list after filter changes done when scrolled into the list. maybe try passing the smooth option to the scrollTo browser API, let's see if it's too slow or not.

@nkuehn after taking a look at current implementation in the main documentation, we don't have smooth scrolling when scrolling to the top of the page. I remember that was a design decision because I did try to implement smooth scrolling with main docs (in the old dev repo) and had to remove it again. If we implement smooth scrolling on the release notes overview, it makes sense to implement it across all pages for a consistent behaviour. So, my question is do we now want smooth scrolling to top of the page on all pages?

nkuehn:
> @nkuehn after taking a look at current implementation in the main documentation, we don't have smooth scrolling when scrolling to the top of the page. ... So, my question is do we now want smooth scrolling to top of the page on all pages?

For the record what we spoke about: I did not mean to make anything on that page smooth (that's centrally defined in the link component anyways, but specifically when the scroll is happening because the user changed a filter.  That's because as opposed to links and nav entries, checkboxes and filters don't conventionally cause navigation to a different position so the list refresh is usually animated in some sort in SPAs.   To avoid going into animations in this initial PR I considered a smooth scroll. 

-->  leave it out for now or try quickly (it's just one parameter to the scrollTo DOM API) and decide what you like better.

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
nkuehn: @emmenko  please clarify this with design - they know the kit well enough and chose two separate boxes. 

I personally think two separate inputs are better because
 * if you squeeze the UI kit range component into the width of the index nav column the second date is partly not visible. 
 * typical use cases just need the start date of the list - forcing the user to select two although they don't need to is worsening the UX.

packages/gatsby-theme-docs/src/templates/release-notes-list.js
nkuehn: > Shouldn't there be some initial from/to dates? 🤔

No, there shouldn't be a preset.  It's a different state to have not selected anything vs. having selected today to e.g. send a link to this week's releases that will still work next week. Also, the date of the oldest release is some random date somewhere in the past that does not make sense to users.  

> if we have releases going back to 5 years ago, do we want to always show/render all of them?

Yes, also this question was already discussed and decided in the design process.  We have no problems with the current implementation (which shows even more than we migrate) and chose to not buy into the UI complexity of a paged or "show more" interface.  

If we get into performance issues we can choose to limit the initial list but that needs some UX and design consideration.   It would be a bit of a pity though to see that the performance is regressing.

packages/gatsby-theme-docs/src/components/release-notes-filter-topics.js
```diff
@@ -0,0 +1,90 @@
...
+  }));
+
+  const [topics, setTopics] = React.useState(checkedTopics);
+
+  return (
+    <Container>
+      <SpacingsStack>
+        <SpacingsInline alignItems="center" justifyContent="space-between">
+          <FilterTitle>Filter By Topics</FilterTitle>
+          <ClearAll onClick={handleOnClearAll}>Clear all</ClearAll>
```
emmenko, line 40: Nit: add also a `aria-label` attribute

packages/gatsby-theme-docs/src/components/release-notes-filter-topics.js
```diff
@@ -0,0 +1,90 @@
...
+  padding-top: ${designSystem.dimensions.spacings.m};
+`;
+const ClearAll = styled.button`
+  border: none;
+  cursor: pointer;
+  padding: 0;
+  color: ${designSystem.colors.light.textInfo};
+  font-size: ${designSystem.typography.fontSizes.extraSmall};
+  text-decoration: none;
+`;
```
emmenko, line 23:
```suggestion
  background-color: transparent;
`;
```

packages/gatsby-theme-docs/src/layouts/release-notes-detail.js
```diff
@@ -50,15 +50,21 @@ const LayoutReleaseNotesDetail = (props) => {
           <LayoutPageHeader>
             <SpacingsInline alignItems="center">
               <AngleLeftIcon size="medium" color="primary" />
-              <Link href="/releases" noUnderline={true}>
```
emmenko, line 53: Why did this change?

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -40,12 +52,55 @@ const ReleaseNotesListTemplate = (props) => (
...
-  </ThemeProvider>
-);
+            </SpacingsStack>
+          </div>
+        </Markdown.TypographyPage>
+      </LayoutReleaseNotesList>
+    </ThemeProvider>
+  );
+
+  function filterReleases(releases, filters) {
```
emmenko, line 62: You might move this outside of the react component.

packages/gatsby-theme-docs/src/utils/extract-query-parameters.js
```diff
@@ -0,0 +1,22 @@
...
+  if (queryString) {
+    const keyValueArray = queryString.split('&');
+    return {
+      fromFilterDate: decodeURIComponent(keyValueArray[0].split('=')[1]),
+      toFilterDate: decodeURIComponent(keyValueArray[1].split('=')[1]),
+      filterTopics: JSON.parse(
+        decodeURIComponent(keyValueArray[2].split('=')[1])
+      ),
+    };
+  }
```
emmenko, line 13: No need to parse this manually, I suggest to use the `qss` library and then the `decode` function

packages/gatsby-theme-docs/src/utils/navigate-with-filters.js
```diff
@@ -0,0 +1,21 @@
...
+    newQueryParameters.filterTopics
+  );
+
+  navigate(
+    `?fromFilterDate=${encodeURIComponent(
+      newQueryParameters.fromFilterDate
+    )}&toFilterDate=${encodeURIComponent(
+      newQueryParameters.toFilterDate
+    )}&filterTopics=${encodeURIComponent(newQueryParameters.filterTopics)}`
+  );
```
emmenko, line 18: Same here, don't build the query string on your own.

packages/gatsby-theme-docs/src/layouts/release-notes-detail.js
davifantasia: I needed the back button functionality/behavior of the browser to preserve the filters state when users use this link to return to release notes list, and we don't have back button functionality in the link component.

nkuehn: @davifantasia  it doesn't work as simple as replacing the link with a back navigation trigger.  If you open the details page URL in a fresh tab the link points into nothing / the user's empty tab page. 

Since the link is named "All Release Notes" it should stay as is and navigate to `/releases`.     

Since as far as I know the browser API does not allow to read the previous history state a "smart" link would be hacky, work only inside a gatsbyJS site etc.  -> I don't think it's worth trying a "smarter" link.

davifantasia: @nkuehn I can attempt to go in a little bit of a different direction - pass the filter states to details page. And when "All Releases" is clicked, then the state can be passed back to the list page. Would that work?

emmenko: I would suggest to leave this behavior for a follow up. Just keep the link as-is for now.

nkuehn: I also suggest to leave the "All" link to point to all.   Giving a page that does not have state URL state would just incur more follow-up issues, e.g. with search crawlers etc.  

Don't mess with the basic specs of the web ;-)

nkuehn:
> Can you guys clarify the need for having the filters as query params? I thought we didn't want to have global state 🤔

@emmenko to do what  URLs are good for:  
 * Link to specific content or views from other pages or by sending a link. 
 * Properly work with the browser history (back / forward navigation).  

The history navigation behavior is  a strong UX convention in filter interfaces, for example nearly all online shop filter UIs do it this way.   You don't want to get back to the amazon home page when pressing "Back" after having selected a subcategory.  You want to get back to the previous list view.

packages/gatsby-theme-docs/src/components/release-notes-filter-topics.js
```diff
@@ -0,0 +1,107 @@
...
+const ReleaseNotesFilterTopics = (props) => {
+  const location = useLocation();
+  const { filterTopics } = extractQueryParameters(location);
+  const topicsSet = useReleaseNotesTopicsSet();
+  const checkedTopics = topicsSet.map((topic) => ({
+    name: topic,
+    checked: isChecked(filterTopics, topic),
+  }));
+
+  function isChecked(filterTopicsArray, topic) {
```
emmenko, line 37: Nit: you can move this function outside of the component

packages/gatsby-theme-docs/src/hooks/use-all-release-notes.js
emmenko: Can you rename this function without the `Set` suffix part?

packages/gatsby-theme-docs/src/templates/release-notes-list.js
```diff
@@ -40,12 +52,20 @@ const ReleaseNotesListTemplate = (props) => (
...
+      </LayoutReleaseNotesList>
+    </ThemeProvider>
+  );
+
+  function scrollToTop() {
+    const layoutPage = document.querySelector('#top');
+    layoutPage.scrollIntoView({
+      block: 'start',
+    });
+  }
```
emmenko, line 67: I would define and use this function as an utility function, there is no need to pass it as a prop.

packages/gatsby-theme-docs/src/utils/navigate-with-filters.js
```diff
@@ -0,0 +1,33 @@
+import { navigate } from '@reach/router';
+import { encode } from 'qss';
+import extractQueryParameters from './extract-query-parameters';
+
+function navigateWithFilters(filters, location) {
+  const currentQueryParameters = extractQueryParameters(location);
+  const newQueryParameters = removeUnnecessaryParameters({
+    ...currentQueryParameters,
+    ...filters,
+  });
```
emmenko, line 10: I don't think there is a need to manually remove those params, the `encode` function does this for you.

image

packages/gatsby-theme-docs/src/utils/navigate-with-filters.js
```diff
@@ -0,0 +1,33 @@
...
+
+  const to =
+    Object.keys(newQueryParameters).length > 0
+      ? encode(newQueryParameters, '?')
+      : location.pathname;
```
davifantasia, line 15: > I don't think there is a need to manually remove those params, the encode function does this for you.

I have 2 reasons:
* to remove the questions mark from the url.
* the encode function does not remove fields with empty strings, so when date field is cleared, you end up with something like this in the url - `http://localhost:8000/releases?fromFilterDate=`

emmenko:
Let me push a commit to clean things up a bit, then we can merge.

Issue #414, opened 2020-04-30
Card component and its responsive behaviour


nkuehn:
The Card component is designed now and we need to 

 1. define how it's written in the author's perspective in MDX
 2. derive an implementation concept and decide what tradeoff(s) to take
 3. build and test.

I propose to do Step 1 in a first, separate PR to better define the requirements and component API - serves as the test "suite" too then. 

designs available to the team in invision.

nkuehn:
Done via #446 and further changes

PR #421, opened 2020-05-05
feat(theme): draft schema and model for cards

emmenko:
Proposal for modelling Cards (#414)

See also #415

---

## Motivation

Cards can be considered like MDX fragments.
Because cards have a deterministic behavior and appearance, we can model most of the information as frontmatter fields.
The body of the MDX file is the body of the card.

All cards information can then be queried in the UI components to programmatically build and render the cards.

image


emmenko:
What do you mean? Aren't cards supposed to be used in the landing page or intro pages?
If not, where can I find the usage requirements?

PR #422, opened 2020-05-06
feat: implement add-ons mechanism to merge markdown components from different add-on themes

emmenko:
## Problem

We have an issue at the moment where the additional theme packages (api-docs, code-examples, constants) are supposed to be used as complementary theme packages (or add-ons) on top of the core docs theme.
Those theme packages, among other things, inject React components to be used in the MDX provider. They do that using Gatsby's component shadowing.

However, when using multiple "stacked" themes, component shadowing will pick the "last" theme child, if all the themes try to shadow the same file.
In our scenario, using components shadowing is not what we want.

## Proposal solution

What we want is to merge the React components from the different "add-on" themes and inject them into the MDX provider of the core theme.
One way to do so is to auto-generate a "proxy" file that re-exports the markdown components from each theme, then use this file to load the components and pass them to the MDX provider.


emmenko:
Yeah, ideally we should be able to configure the "add-ons" only once. Not sure yet how that would work, I can look into it.

About the naming, ideally we should have something like `gatsby-theme-addon-xxx`, but that would mean renaming the packages again. I think it's fine to keep them as `gatsby-theme-xxx`.

emmenko:
I looked into passing the theme as a `plugins` options of the other theme. It doesn't work though, as Gatsby treats them as plugins, and not themes.

> A Gatsby theme is a type of plugin that includes a `gatsby-config.js` file and adds pre-configured functionality, data sourcing, and/or UI code to Gatsby sites.
> With a Gatsby theme, all of your default configuration (shared functionality, data sourcing, design) is abstracted out of your site and into an installable package. A theme might differ from a typical plugin in that it packages up usage of a plugin into a consumable API, making it easy to include functionality without having to type out all of the code by hand (such as GraphQL queries).
> https://www.gatsbyjs.org/docs/plugins-themes-and-starters/

The key part here is the usage of the `gatsby-config.js`, which our “add-ons” need for configuring other source plugins, etc. They also need to be a theme for the component shadowing to work.

As @nkuehn  pointed out, the only annoying thing is that you need to “duplicate” the name of the “add-on” in the website’s `gatsby-config.js`.
I couldn’t find a solution for that, so from my side I think it’s a little trade-off that we need to have.

emmenko:
Alright, I pushed a commit (https://github.com/commercetools/commercetools-docs-kit/pull/422/commits/f060ccfd8556679943afb93d46e93634dd456202) that helps configuring the plugin with add-ons using a configuration function, so less typing and duplication.

packages/gatsby-theme-docs/configure-theme.js
```diff
@@ -0,0 +1,25 @@
+const pkg = require('./package.json');
+
+function configureThemeWithAddOns(themeOptions = {}) {
+  const { addOnPlugins, ...coreThemeOptions } = themeOptions;
+
+  if (!addOnPlugins || !Array.isArray(addOnPlugins)) {
+    throw new Error(
+      'Missing or wrong format for "addOnPlugins" option. Expected a list of Gatsby theme add-on plugins.'
+    );
+  }
```
davifantasia, line 10: Does this imply that gatsby-theme-docs cannot be used as. a standalone?

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -27,10 +28,13 @@ exports.onPreBootstrap = ({ reporter }) => {
...
   requiredDirectories.forEach((dir) => {
     if (!fs.existsSync(dir)) {
-      reporter.info(`creating the ${dir} directory`);
+      gatsbyApi.reporter.info(`creating the ${dir} directory`);
       fs.mkdirSync(dir);
     }
   });
+
+  // Load theme add-ons
+  loadThemeAddOns(gatsbyApi, themeOptions);
```
davifantasia, line 37:
```suggestion
  createAddOnThemeFiles(gatsbyApi, themeOptions);
```

packages/gatsby-theme-docs/configure-theme.js
emmenko: No you can still use that. This is just an "helper function" to keep the configuration simple.

See detailed explanation: https://github.com/commercetools/commercetools-docs-kit/blob/2f6c3bf8b24ba6e81a456b5fc9e1d1db323bbf47/packages/gatsby-theme-docs/README.md#using-theme-with-add-ons

packages/gatsby-theme-docs/gatsby-node.js
emmenko: True. I renamed it to `bootstrap` instead of `load`, since it's also part of the `onPreBootstrap` step.

Hope it's fine for you.

packages/gatsby-theme-docs/utils/bootstrap-theme-addons.js
```diff
@@ -0,0 +1,45 @@
...
+}
+  `;
+  const addOnsFolderPath =
+    'src/@commercetools-docs/gatsby-theme-docs/overrides';
+  const markdownComponentsPath = path.join(
+    addOnsFolderPath,
+    'markdown-components.js'
+  );
+  if (!fs.existsSync(addOnsFolderPath)) {
+    fs.mkdirSync('src/@commercetools-docs');
```
nkuehn, line 22: On that occasion and since we're doing a breaking change anyways:  We could pull up the node.js Version requirement from 10 to 12 in package.json.    Then you can for example use the "recursive" option of fs.mkDirSync here, it was introduced in 10.12.0. 

Or, for a less intrusive change, set it to  `>=10.13.0` which was the start of the 10.x LTS releases and the current LTS until April 2020.

packages/gatsby-theme-docs/README.md
```diff
@@ -82,6 +82,8 @@ The project structure should contain at least the following files and folders:
 
   - `overrideDefaultConfigurationData` (_optional_, array of glob strings): allows to replace the configuration files in `src/data` instead of augmenting them. The option is passed to the `ignore` [option of the gatsby filesystem plugin](https://www.gatsbyjs.org/packages/gatsby-source-filesystem/#options). For example, by passing `['**/top-*']` and placing `top-menu.yaml` and `top-side-menu.yaml` files in the website's `src/data` folder the top navigation can be overridden completely. If this option is used, the files matching the glob patterns **must** be provided.
 
+  - `themeAddOns` (_optional_, array of Gatsby Theme plugin names): see [using Theme with Add-Ons](#using-theme-with-addons).
```
nkuehn, line 85: since it's in the context of the "theme-docs" plugin, I think the name of this can be only `addOns`.

packages/gatsby-theme-docs/README.md
```diff
@@ -152,6 +154,67 @@ The available JSX components are:
...
+  configureThemeWithAddOns,
+} = require('@commercetools-docs/gatsby-theme-docs/configure-theme');
+
+module.exports = {
+  plugins: [
+    ...configureThemeWithAddOns({
+      // Pass the normal theme options
+      websiteKey: 'my-website-key',
+      // Define the add-on plugins
+      addOnPlugins: [
```
nkuehn, line 178: it's really confusing that depending on the form of configuration the parameter has a different name.  The `configureThemeWithAddOns`  function would be possible also if both variations were just called `addOns` - the function could still handle it properly.

packages/gatsby-theme-docs/README.md
```diff
@@ -152,6 +154,67 @@ The available JSX components are:
...
+            // ...
+          },
+        },
+      ],
+    }),
+  ],
+};
+```
+
+Alternatively you should pass the Gatsby theme add-ons as normal plugins and additionally reference the package names in the `themeAddOns` option of the commercetools core theme.
```
nkuehn, line 192: (we spoke about this today, just for documentation here): I suggest to not advertise this third option, but to be clear that add ons have to be configured through the configuration function  if they are used.

packages/gatsby-theme-docs/configure-theme.js
```diff
@@ -0,0 +1,25 @@
+const pkg = require('./package.json');
+
+function configureThemeWithAddOns(themeOptions = {}) {
+  const { addOnPlugins, ...coreThemeOptions } = themeOptions;
+
+  if (!addOnPlugins || !Array.isArray(addOnPlugins)) {
```
nkuehn, line 6: Why would the function not work if no add on plugins are configured?  We could default the option to an empty array here, that would make copy / pasting and modifying site configurations (e.g. commenting out that line to test) more robust.

websites/api-docs-smoke-test/.gitignore
```diff
@@ -0,0 +1 @@
+@commercetools-docs
```
nkuehn, line 1: The gitignore should be autogenerated, located directly next to the generated file and cover only this exact file.  

That way it is not forgotten, does not interfere with other intentional .gitignore 's and still allows intended component shadowing in a given website.

packages/gatsby-theme-docs/configure-theme.js
```diff
@@ -0,0 +1,25 @@
...
+      themeAddOns: addOnPlugins.map((plugin) =>
+        typeof plugin === 'string' ? plugin : plugin.resolve
+      ),
+    },
+  };
+  // Flatten theme plugins, so that they are loaded by Gatsby as themes
+  return [coreThemeConfig, ...addOnPlugins];
+}
+
+exports.configureThemeWithAddOns = configureThemeWithAddOns;
```
nkuehn, line 25: Nitpick: can we re-export this from the `index.js` root of the package as you recommended generally?  That would make the configuration slightly more concise and we can refactor the actual location if we want.

packages/gatsby-theme-docs/README.md
emmenko: Correct. Now that we don’t document the “alternative” approach anymore, we can use the same value, otherwise it would have been confusing.

packages/gatsby-theme-docs/configure-theme.js
emmenko: Sure, if you prefer so. The check was meant to prevent the user to use the function without passing any add-on.

websites/api-docs-smoke-test/.gitignore
emmenko: We could also just commit the file in VSC, it shouldn’t be a big of a problem.
I’ll add the gitignore

packages/gatsby-theme-docs/configure-theme.js
emmenko: The `index.js` only exports React components, or things needed in the UI.
I wanted to keep that separate from the nodejs part.
The export file can still be there, even if we move things around. We just need to re-export things properly.

packages/gatsby-theme-docs/utils/bootstrap-theme-addons.js
emmenko: We'll do that in a separate PR

PR #427, opened 2020-05-08
feat(theme): add template with hero image for website homepage

emmenko:


packages/gatsby-theme-docs/src/layouts/content-homepage.js
```diff
@@ -0,0 +1,77 @@
...
+          <LayoutPageWithHero title={props.pageData.title}>
+            <LayoutPage>
+              <LayoutPageContent>
+                <PageContentInset id="body-content">
+                  {props.children}
+                  <ContentPagination slug={props.pageContext.slug} />
+                </PageContentInset>
+              </LayoutPageContent>
+            </LayoutPage>
+          </LayoutPageWithHero>
```
emmenko, line 52: This is the new component for the homepage layout

packages/gatsby-theme-docs/src/layouts/internals/layout-page-with-hero.js
```diff
@@ -0,0 +1,75 @@
...
+
+// TODO: for now we assume that the website provides a file `homepage-hero.svg`
+// in the `static` folder.
+const homepageHeroSvgUri = '/homepage-hero.svg';
+
+// The container is used to render a background image with for the full page width.
+const Container = styled.div`
+  grid-area: page-header;
+  position: relative;
+  background-color: ${(props) => props.theme.colors.light.primary};
```
emmenko, line 15: I assume that as a fallback the background color should be the website primary color

packages/gatsby-theme-docs/src/layouts/internals/layout-page-with-hero.js
```diff
@@ -0,0 +1,75 @@
...
+`;
+
+const LayoutPageWithHero = (props) => (
+  <>
+    <Container heroBackgroundURL={props.heroBackgroundURL}>
+      <ThemeProvider
+        theme={{
+          colors: {
+            light: {
+              primary: designSystem.colors.light.textInverted,
```
emmenko, line 58: I assume that the heading color should be the "inverted" one, so basically the "white" color, given that the websites primary colors are darker colors.

This needs to be checked with @Shecki though. If necessary, the text color should also be configurable.

websites/docs-smoke-test/src/content/index.mdx
```diff
@@ -1,6 +1,5 @@
 ---
-title: Full page example of markdown elements
-beta: true
+title: Explore how to build a documentation website for commercetools
```
emmenko, line 2: To give it a bit nicer title

websites/docs-smoke-test/src/content/views/beta.mdx
```diff
@@ -0,0 +1,4 @@
+---
+title: Beta page
+beta: true
```
emmenko, line 3: The homepage does not use the "beta" flag. For VRT purposes we have a new page now.

emmenko:
@nkuehn @Shecki I refactored the configuration set up for the website colors and introduced what I called "color presets".

https://github.com/commercetools/commercetools-docs-kit/tree/nm-homepage-hero/packages/gatsby-theme-docs/color-presets

For now there is one preset `base`, which is the black color.

packages/gatsby-theme-docs/src/layouts/content-homepage.js
```diff
@@ -0,0 +1,82 @@
...
+          siteTitle={siteData.siteMetadata.title}
+          excludeFromSearchIndex={props.pageData.excludeFromSearchIndex}
+        />
+        <LayoutPageWrapper id="top">
+          <LayoutPageWithHero
+            title={props.pageData.title}
+            heroBackgroundURL={props.heroBackground.publicURL}
+          >
+            <LayoutPage>
+              <LayoutPageContent>
```
nkuehn, line 48: I think we need to override the `LayoutPageContent`, too  or find another way to give it `grid-area: page-content / page-content / page-navigation / page-navigation`  instead of `grid-area: page-content`.   

The design intent was to have the most relevant cards above the fold and rather drop the index nav column for that.

emmenko: I don't get your point. There is no index nav in this page, right?

nkuehn: Yes, and the content column uses the width of the index nav for more side-by-side card or banner content in the design, but not yet in the implementation.

emmenko: Can you point me to the design that shows that?

websites/docs-smoke-test/src/@commercetools-docs/gatsby-theme-docs/overrides/markdown-components.js
```diff
@@ -0,0 +1,16 @@
...
+import { designSystem } from '@commercetools-docs/ui-kit';
+
+import addonComponents from './addon-components';
+
+const DummyCards = styled.div`
+  display: grid;
+  gap: ${designSystem.dimensions.spacings.m};
+  grid-auto-columns: 1fr;
+  grid-template-columns: ${(props) =>
+    `repeat(auto-fill, minmax(${props.narrow ? '200' : '250'}px, 1fr))`};
```
nkuehn, line 12:
```suggestion
    `repeat(auto-fill, minmax(${props.narrow ? '242' : '328'}px, 1fr))`};
```
This makes regular cards two-aside on landing pages and narrow cards three-aside in any landing page size and you can put 2,3,4,6 cards in a grid without whitespace.

It's not ideal in content but let's optimize for the landing page first. 

PS: If you think they aren't actually "Narrow" I would be fine swapping the parameter and rather having a "wide" option. 

PPS: I found two numbers that would give stable layouts in content too if the landing page had 16px more content width (248px / 378px), but that feels hacky - I would prefer to explicitly implement other sizes for cards that live in content (there should be only one card size in content pages because three-aside is too small anyways there)

.changeset/brown-mangos-hunt.md
```diff
@@ -0,0 +1,19 @@
+---
```
nkuehn, line 1: is the file name random or a reference to something I missed?

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -49,6 +49,14 @@ module.exports = (themeOptions = {}) => {
             node { etc...
       */
 
+      // Color presets
+      {
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'colorPresets',
+          path: path.join(__dirname, `./color-presets`),
```
nkuehn, line 57: to be on the safe side, I would ignore the readme and code to prevent them from being handled as content by other plugins that don't explicitly filter out the name of the source plugin (which most don't)

```suggestion
          path: path.join(__dirname, `./color-presets`),
          ignore: [`**/*.md`,`**/*.js`], 
```
PSEUDOCODE!

packages/gatsby-theme-docs/src/layouts/internals/layout-page-with-hero.js
```diff
@@ -0,0 +1,96 @@
...
+  background-image: ${(props) => `url('${props.heroBackgroundURL}')`};
+
+  @media screen and (${designSystem.dimensions.viewports.tablet}) {
+    background-size: auto;
+  }
+  @media screen and (${designSystem.dimensions.viewports.largeDesktop}) {
+    background-size: cover;
+  }
+
+  /* License with attribution of the background by SVGBackgrounds.com */
```
nkuehn, line 25: Wouldn't this comment better live in the folder where the images are?  It could be forgotten here once our own images are in.

packages/gatsby-theme-docs/src/layouts/internals/layout-page-with-hero.js
nkuehn: In our conversation with him he said that the text will always be pure white when over the background image.

.changeset/brown-mangos-hunt.md
emmenko: Yes, it's randomly generated

packages/gatsby-theme-docs/gatsby-config.js
emmenko: Good point 👌

packages/gatsby-theme-docs/src/layouts/internals/layout-page-with-hero.js
emmenko: I'd rather keep it here for now, as we need to somehow mention the license and that is "visible" on production

packages/gatsby-theme-docs/src/layouts/internals/layout-page-with-hero.js
emmenko: Yes, the `inverted` color points to: `color-surface`

image

PR #440, opened 2020-05-15
cards: added icons tests cases, and more specification for titles

davifantasia:


davifantasia:
@nkuehn I can't find the link of your comment on making cards more accessible, please include it here.

emmenko:
Can you clarify why you chose to pass icons as strings? I thought we wanted to go for components.

```
icon={<MyIcon />}
```

davifantasia:
> Can you clarify why you chose to pass icons as strings? I thought we wanted to go for components.
> 
> ```
> icon={<MyIcon />}
> ```

Fixed, thanks.

emmenko:
Ok.

However, this will fail because those components do not exist yet, that's why the icons examples have been removed 😅

davifantasia:
> Ok.
> 
> However, this will fail because those components do not exist yet, that's why the icons examples have been removed 😅

It builds according to Vercel :)

emmenko:
Yes, cause the icon is not rendered as the card component hasn’t been implemented yet.

Anyway, just wanted to make you aware of this

davifantasia:
> Anyway, just wanted to make you aware of this

Thanks. Please approve if you're good with the PR.

websites/docs-smoke-test/src/content/components/cards.mdx
```diff
@@ -172,3 +173,53 @@ would | work   |
 ## Output
```
emmenko: I think we forgot this last time

```suggestion
**Output**
```

PR #442, opened 2020-05-17
fix(theme): render release notes filters in overlay for small screens

emmenko:
image

### DIFF
--- a/.changeset/warm-kids-sparkle.md
+++ b/.changeset/warm-kids-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Render release notes filters in overlay for small screens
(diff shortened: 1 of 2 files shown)


emmenko:
I can take a look at that.

davifantasia:
Thanks for fixing this @emmenko .

I expected the filter overlay to have similar behaviours like the index nav overlay. But you included a close button and the transparent background does not close the overlay when clicked, is the different behaviour the expectation from design?

Screenshot 2020-05-18 at 10 30 27

emmenko:
I haven't checked with design yet but the behavior is indeed different.

In the content index nav, you have link. So when you click on a link you want to overlay to be closed immediately.

In the filters, you want to overlay to stay open, because you can configure multiple filters. Hence the need for a close button.

emmenko:
The last commit should fix things up: https://github.com/commercetools/commercetools-docs-kit/pull/442/commits/dd2cdd1a92a7e9e3006f71e17a339719f4f2a320

nkuehn:
approved by merging it :-).  Thank you!

PR #446, opened 2020-05-18
feat: Cards

davifantasia:
Note: The components are implemented in the main theme since `Link` component of the theme is a dependency. Please let me know if you have a better solution for this.

### Rendering Markdown In Body

These don't work:

* Bullet lists - the content is passed to the card as a string so when parsing with markdownFragmentToReact, only one all items are rendered as part of the first item of the list.
* Multiple paragraphs - not sure why but one paragraph is passed as a string and the other is passed as an object..


packages/gatsby-theme-docs/src/components/card.js
```diff
@@ -0,0 +1,99 @@
...
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import css from '@emotion/css';
+import { navigate } from '@reach/router';
+import SpacingsStack from '@commercetools-uikit/spacings-stack';
+import SpacingsInline from '@commercetools-uikit/spacings-inline';
+import Link from './link';
+
+const flatStyle = css`
+  border: 1px solid #cccccc;
```
emmenko, line 11: All the colors, sizes, etc must be derived from the design system. Please don't forget about that.

packages/gatsby-theme-docs/src/components/cards.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import { designSystem, ContentNotifications } from '@commercetools-docs/ui-kit';
+
+// Explanation about the following sizes:
+// https://github.com/commercetools/commercetools-docs-kit/pull/427#discussion_r425442556
+const cardNarrowMinWidth = '242px';
+const cardRegularMinWidth = '328px';
+const CardsContainer = styled.div`
```
emmenko, line 10:
```suggestion
const Container = styled.div`
```

packages/gatsby-theme-docs/src/components/cards.js
```diff
@@ -0,0 +1,54 @@
...
+  grid-template-columns: ${(props) =>
+    `repeat(auto-fill, minmax(${
+      props.narrow ? cardNarrowMinWidth : cardRegularMinWidth
+    }, 1fr))`};
+`;
+
+const Cards = (props) => {
+  try {
+    return (
+      <CardsContainer {...props}>
```
emmenko, line 23:
```suggestion
      <Container {...props}>
```

packages/gatsby-theme-docs/src/components/cards.js
```diff
@@ -0,0 +1,54 @@
...
+          if (!child.props || child.props.mdxType !== 'Card') {
+            throw new Error(
+              `Children of <Cards> must be a <Card> component and not "${
+                child.props ? child.props.mdxType : child
+              }"`
+            );
+          }
+          return React.cloneElement(child, [props], [...child.props.children]);
+        })}
+      </CardsContainer>
```
emmenko, line 34:
```suggestion
      </Container>
```

packages/gatsby-theme-docs/src/components/cards.js
```diff
@@ -0,0 +1,54 @@
...
+      <CardsContainer {...props}>
+        {React.Children.map(props.children, (child) => {
+          if (!child.props || child.props.mdxType !== 'Card') {
+            throw new Error(
+              `Children of <Cards> must be a <Card> component and not "${
+                child.props ? child.props.mdxType : child
+              }"`
+            );
+          }
+          return React.cloneElement(child, [props], [...child.props.children]);
```
emmenko, line 32: Why is `[props]` an array?

packages/gatsby-theme-docs/src/components/cards.js
```diff
@@ -0,0 +1,54 @@
...
+      props.narrow ? cardNarrowMinWidth : cardRegularMinWidth
+    }, 1fr))`};
+`;
+
+const Cards = (props) => {
+  try {
+    return (
+      <CardsContainer {...props}>
+        {React.Children.map(props.children, (child) => {
+          if (!child.props || child.props.mdxType !== 'Card') {
```
emmenko, line 25: Note that by using `mdxType` it implies that the `Cards` component can only be used within MDX files.
It would be better to check for example if the child component `displayName` is `Card`.

packages/gatsby-theme-docs/src/components/cards.js
```diff
@@ -0,0 +1,54 @@
...
+
+    throw e;
+  }
+};
+
+Cards.propTypes = {
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.element.isRequired),
+    PropTypes.element.isRequired,
+  ]).isRequired,
```
emmenko, line 51:
```suggestion
  children: PropTypes.node.isRequired,
```

packages/gatsby-theme-docs/src/components/card.js
```diff
@@ -0,0 +1,104 @@
...
+      {...props[0]}
+      onClick={props[0].clickable ? handleClick : undefined}
+    >
+      {props.icon ? renderCardContentWithIconLayout() : renderCardContent()}
+    </CardContainer>
+  );
+
+  function handleClick(e) {
+    e.preventDefault();
+    navigate(props.href);
```
emmenko, line 64: Why not wrapping the card with a `<Link>` component or something? It would also be semantically more correct.

packages/ui-kit/src/design-system.js
```diff
@@ -64,9 +64,12 @@ export const tokens = {
   borderRadiusForContentNotification: customProperties.borderRadius6,
   borderRadiusForImageFrame: customProperties.borderRadius6,
   borderRadiusForBlockquote: customProperties.borderRadius6,
+  borderRadiusForCard: customProperties.borderRadius6,
   shadowForBetaFlag: customProperties.shadow7,
   shadowForSearchDialog: customProperties.shadow2,
   shadowForPageContent: customProperties.shadow11,
+  shadowForClickableCard: `1px 1px 4px 0 rgba(0, 0, 0, 0.24),
+  -1px 1px 3px 0 rgba(0, 0, 0, 0.12)`,
```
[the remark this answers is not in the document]
davifantasia: I checked and did not find it.

packages/gatsby-theme-docs/src/components/cards.js
davifantasia: I'm not sure, I just used the code sample as is:

```
React.cloneElement(
  element,
  [props],
  [...children]
)
```

https://reactjs.org/docs/react-api.html#cloneelement

emmenko: 🤦‍♂️

Sorry but you can't just copy paste stuff like this without understanding what it does.
The docs example indicates that the arguments are optional, with the `[ ]` notation.

In this case you do it like this:

```js
React.cloneElement(element, props, children)
```

It would be better though to be explicit which props the parent should inject to the children.

```js
React.cloneElement(element, { foo: 'bar'}, children)
```

Which props does the `<Card>` component need from the parent?

packages/gatsby-theme-docs/src/components/card.js
nkuehn: Yes, it should be a real link but It could be that we have to refactor the Link code a bit to separate the behaviour from the Styling.  

It's not ideal for accessibility to the have the whole card content be the link text (the screenreader then reads out the whole text twice) but if the card does not have a text link at the bottom it would go completely unnoticed as a link so wrapping the whole thing is better IMHO.

davifantasia: There is an invalid DOM nesting problem of `<a>` of the text link being a child of the `<a>` from the `Link` component.

nkuehn: PS:  There's a formal problem though:  https://html.spec.whatwg.org/#the-a-element  

>"there must be no interactive content descendant, a element descendant, or descendant with the tabindex attribute specified."

So strictly spoken, since the content of the card can contain links, we would only be on the safe side not making the card wrapped in a link and tagging the javascript link using aria spec attributes:

Would the following pseudo-html make sense? 

```html
<ul class="cards-wrapper">
   <li class="card" 
        role="link" 
        aria-label="The Card Title (if available)" 
        onClick="theSameCodeWeUseInOurLinkComponent('/foo/bar')">
           <h6>The Card Title</h5>
           <p>All the <strong>formatted</strong> markdown card body</p>
           <a href="/foo/bar">The textLink if passed</a>
    </li>
 </ul>
```

emmenko: Well the solution is easy: wrap the card with a link if there is no "bottom" link, otherwise the card is a div

davifantasia: I guess I can just copy the link style instead and apply it to a different DOM element for the text link.

emmenko: Otherwise use aria attributes to indicate that the card is a link (if there is no "bottom" link).

nkuehn: > Well the solution is easy: wrap the card with a link if there is no "bottom" link, otherwise the card is a div

The card content is markdown, that's not safe either.  We'll have to live with it being a "fake" link.  It's not a problem for crawlers, the pages we want indexed are in the nav anyways.

nkuehn: What I have no idea how to solve cleanly is how to properly cover all the functional behavior the `PureLink` has (  https://github.com/commercetools/commercetools-docs-kit/blob/master/packages/gatsby-theme-docs/src/components/link.js#L97 ) 

If we don't want to duplicate it all here's an idea ( I have no idea whether this works in react): We could _always_ instantiate the textLink  at the bottom if an `href` is present.  `textLink` would then only trigger whether it's visible or not.   A click event on the card would be passed through to the text link,  which would make sure than any necessary behavior (including google analytics,  behavior buried in the GatsbyLink that it's using etc.) would be used and nothing is duplicated.  

@emmenko  crazy or useful?

nkuehn: > A click event on the card would be passed through to the text link,

PS: this was a "jQuery-ish" way to word it where you can randomly wire events inside the DOM without constraints.   My naive idea in React would be that the Card component acquires a ref to the virtual DOM element of the `<Link ...>` it has as a child and then calls the `click()` on that ref when the whole card is clicked.   But I bet there's more proper ways to handle it.

emmenko: It seems overly complicated to me. Why not “the card is a link or a div”?
Yes, the content might include links but is that even a use case we want to support?

nkuehn: Cards that are clickable and have a text link at the bottom that points to the same location are a known and intended use case because there's no hover effect on touch devices so for the primary call to action you want to have the link (and the text that makes clearer where you navigate to). 

But disallowing links in the card body content would be OK though (a.k.a. "it just fails", I would not even want to build the logic to catch it)

Same link on both the card and the bottom could be done by the "inverted" hack that the card is a link and the link itself isn't a link but just looks like one.  Not sure if that is nicer implementation wise, but it's clearly clean from the DOM and HTML perspective! 

So for illustration does this look acceptable as pseudo-html:

**Clickable:**
```html
<li class=card>
  <a href="/foo/bar" class="LinkComponent">
     <h6>Title</h6>
     <p>markdown <strong>formatted</strong> content but no links possible. 
     <span class="fake-link-that-changes-color-on-hover">Read more about cards...</span>
  </a>
</li>
```

**Not clickable:**
```html
<li class=card>
     <h6>Title</h6>
     <p>markdown <strong>formatted</strong> content and <a href="/baz" class="LinkComponent">links</a> possible. 
     <a href="/foo/bar" class="LinkComponent">Read more about cards...</a>
</li>
```

emmenko: Yes, the pseudo-html output is what I would expect as well 👌

packages/gatsby-theme-docs/src/components/card.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import css from '@emotion/css';
+import { navigate } from '@reach/router';
+import SpacingsStack from '@commercetools-uikit/spacings-stack';
+import SpacingsInline from '@commercetools-uikit/spacings-inline';
+import { designSystem } from '@commercetools-docs/ui-kit';
+// import markdownFragmentToReact from '../utils/markdown-fragment-to-react';
```
davifantasia, line 9: @emmenko I'm getting an error that prevents the card component from being exported when I import `markdownFragmentToReact`, can you please help with this?

emmenko: Can you be more specific? What error?

davifantasia: it's logged in the browser:

Screenshot 2020-05-20 at 11 16 25

Screenshot 2020-05-20 at 11 16 59

emmenko: Is this happening only when you import `markdownFragmentToReact`?

How are you using that function?

davifantasia: it happens only when I import.

emmenko: Just a guess maybe. I'm not sure this is supposed to work like this to be honest.
The card components are rendered within MDX, and here we're trying to render yet another MDX within MDX...it kind of feels like a vicious loop.

I suppose we just need to let MDX render the content, without us needing to interfere.

Again, I haven't looked into that yet, just some long guess.

@nkuehn what do you think?

davifantasia: @nkuehn found the problem :)

nkuehn: @emmenko cyclic import -> export -> import of the same thing.  never thought about that being an issue.  The card was imported into the fragement renderer via the MDX markdown components and then imported by the card.  

It wasn't correct from the beginning to have the markdown fragment renderer import the basic UI kit components via the MDX components provider instead of directly.  

Concerning "do we need this at all" -> yes, I'm running into more and more cases where it's becoming clear that you need different rendering if it's "squeezed inside some other box".   E.g. just opened #451

nkuehn:
@davifantasia  I think this is ready to go out of its draft status.  The result is fine for me if we don't find a solution for contextually applying different minimum widths.

packages/ui-kit/src/design-system.js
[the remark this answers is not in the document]
emmenko: Then please clarify this with @Shecki which one of the uikit shadow tokens should be used: https://uikit.commercetools.com/iframe.html?id=basics-tokens--all-tokens&viewMode=story#shadow

websites/docs-smoke-test/src/content/components/cards.mdx
```diff
@@ -23,6 +23,7 @@ title: Cards
  * `clickable` (boolean): if set, the whole area of all child cards is a link, the card is elevated and has a hover effect if a link url is set.
  * `narrow` (boolean): by default, two cards are shown side by side (except mobile viewports). if `narrow` is passed, the minimum card width is reduced so that three cards fit the content width on microsite landing pages.
  * `smallTitle` (boolean): lets all cards have a smaller title font.
+ * `fitContentColumn` (boolean): renders a little differently for cards used on content pages.
```
emmenko, line 26: We should mention that this is automatically set on content pages.

packages/gatsby-theme-docs/src/components/card.js
```diff
@@ -0,0 +1,155 @@
...
+`;
+const StackContainer = styled.div`
+  flex-grow: 1;
+  display: flex;
+  height: 100%;
+  flex-direction: column;
+  > * + * {
+    margin-top: ${designSystem.dimensions.spacings.m};
+  }
+`;
```
emmenko, line 51: Hmm sorry but we should be able to use the spacing components from the uikit. There is no need to re-implement them.
Can you clarify the reasoning here?

davifantasia: Screenshot 2020-05-25 at 12 14 40

For situations like in the image above, I needed to the link aligned to the bottom of the page. Using spacings the height and flex-grow needed to be styled, but I understand the ui-kit components are not meant to be styled from the last conversation I had with you, hence custom containers.

nkuehn: @emmenko  if you think it makes sense and is realistic to bring the requirement into the UI kit I can live with doing this sort of alignment later and wait for a UI Kit release.  

Use case: "I want a StackContainer to fill the available height and then be able to mark which of its children should fill the remaining height".    Basically tagging a specific child flex-grow: 1.  

Off-topic but FYI: From my API design experiments I have another use case that is pretty inconvenient with the UI kit since it was never intended for fully responsive :  Pass a responsive breakpoint to an inline stack to tell it from when on to wrap into one column (or just to generally wrap)

emmenko: I see, for this use case then it’s fine yes.
I thought you only wanted to define the spacing.

All good then 👌

packages/ui-kit/src/design-system.js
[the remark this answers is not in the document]
emmenko: @davifantasia have you clarified this with @Shecki ?

PR #454, opened 2020-05-22
feat(markdown): refactor content spacing to better render enclosed fragments

### LINKED ISSUE #451: docs UI Kit: Move Block Spacing to TypographyPage
The Information boxes and Subtitle Fragments can be used with plain-text content or by rendering the the content with the full MDX content renderer if an extra blank line is included before and after the content. 

This is the built-in behavior of MDX and is practical, also because custom rendering with the fragment renderer is only reliable for all use cases when the markdown is available programmatically - when passed from an MDX page, the string inside a tag is already trimmed and removed of all newlines when it reaches the component (HTML compatibility in JSX). 

When using MDX inside JSX Tags, we now realized the that UI kit "Markdown" components are applying sometimes large spacing that is specific to when they are used directly in the body text flow, but not when they are used inside another container. 

The UI kit "Markdown" component already does factor some of these things into the "TypopgraphyPage" Wrapper that can then be applied selectively but it's not consistent.  Some Block spacing, like the large space below a list, is directly inside the list component. 

For example, the following screenshot shows that a bullet list in an information box gets a way to large bottom spacing because that's intended in the body flow content. 

image

MDX:
```mdx
<Info>

- Inquire at [fooBar@commercetools.com](mailto:training@commercetools.com) or use the online [request form](https://ok.commercetools.com/training-offering).
- another bULLet
- Please stay healthy

</Info>
```

**Intended behavior**  No larges space in between or below components when used inside other wrapper boxes.

nkuehn:
Resolves #451 

---

As a counterpart to the `TypographyPage` wrapper, the Markdown UI components now export a `TypographyContainer` wrapper which is supposed to be used in use cases like Cards, Notifications, Subtitle, Table Cells etc. - overall, in places where markdown content is rendered not in the main body flow but inside a visual enclosure. 

In such enclosures the standard spacing is too large and some of it (like bottom margins) should never applied at all.  

Spacing that is specific to the main page content typography is pulled out of the specific components into the wrappers.  

The TypographyPage wrapper is to be applied to Card content, too once merged

---

In addition, this makes a small adjustment to the spacing of nested lists and lists:  Lists previously had very different spacing to the content above them depending on whether they are a a sub-list in a list or a list in the top body content.  Esp. sub-lists looked like a bug because there was zero spacing. 

because tech writers demanded that lists be visually "Bounded" to the paragraph  above or parent list item, which is usually the name of the list,  they now have a small spacing in both situations.

### DIFF
--- a/.changeset/refactor-content-spacing.md
+++ b/.changeset/refactor-content-spacing.md
@@ -1,0 +1,8 @@
+---
+"@commercetools-docs/ui-kit": minor
+"@commercetools-website/docs-smoke-test": patch
+---
+
+Markdown UI Kit: refactor content spacing to be able to render markdown inside
+
+Markdown UI Kit: refactor content spacing to be able to render markdown inside
--- a/.changeset/refactor-content-spacing.md
+++ b/.changeset/refactor-content-spacing.md
@@ -1,6 +1,6 @@
 ---
-"@commercetools-docs/ui-kit": minor
-"@commercetools-website/docs-smoke-test": patch
+'@commercetools-docs/ui-kit': minor
+'@commercetools-website/docs-smoke-test': minor
 ---
 
 Markdown UI Kit: refactor content spacing to be able to render markdown inside
--- a/.changeset/refactor-content-spacing.md
+++ b/.changeset/refactor-content-spacing.md
@@ -4,5 +4,3 @@
 ---
 
 Markdown UI Kit: refactor content spacing to be able to render markdown inside
-
-Markdown UI Kit: refactor content spacing to be able to render markdown inside
--- a/packages/ui-kit/src/components/markdown.js
+++ b/packages/ui-kit/src/components/markdown.js
@@ -3,12 +3,13 @@ import { css } from '@emotion/core';
 import styled from '@emotion/styled';
 import RibbonIcon from '../icons/ribbon-icon.svg';
 import { colors, dimensions, typography, tokens } from '../design-system';
-import { CodeBlockMarkdownWrapper as CodeBlock } from './multi-code-block';
+import {
+  CodeBlockMarkdownWrapper as CodeBlock,
+  Container as CodeBlockContainer,
+} from './multi-code-block';
 
 const headerStyles = () => css`
-  font-weight: ${typography.fontWeights.medium};
   line-height: 1.3;
-  margin: ${dimensions.spacings.m} 0 ${dimensions.spacings.s};
 `;
 
 const Paragraph = styled.p`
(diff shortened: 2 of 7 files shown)


packages/ui-kit/src/components/markdown.js
```diff
@@ -62,46 +51,57 @@ const H6 = styled.h6`
...
+/*
+The "container" styles have to be applied to containers that render markdown inside a surrounding
+visual box like a blockquote, notification box, card, or Subtitle.
+Heading margins are not set here because headings can and should not be used inside such containers.
+*/
+const containerStyles = () => css`
+  > * + * {
+    margin-top: ${dimensions.spacings.m};
+    margin-bottom: ${dimensions.spacings.m};
+  }
```
davifantasia, line 64:
```suggestion
  > * + * {
    margin-top: ${dimensions.spacings.m};
  }
```

This should be sufficient.

refactor-content-spacing.md
```diff
@@ -0,0 +1,8 @@
+---
+"@commercetools-docs/ui-kit": minor
+"@commercetools-website/docs-smoke-test": patch
+---
+
+Markdown UI Kit: Refactor content spacing to be able to render markdown inside visual boxes using a new `TypographyContainer` wrapper.
+
+Markdown UI Kit: Adjust margins of lists
```
davifantasia, line 8: This file should not exist twice - here and in `.changset` directory?

packages/ui-kit/src/components/markdown.js
```diff
@@ -62,46 +51,57 @@ const H6 = styled.h6`
...
+  > :first-child {
+    margin-top: 0;
+  }
+  > :last-child {
+    margin-bottom: 0;
+  }
```
emmenko, line 70: If you use the lobotomized owl selector, you only need this:

```suggestion
  > * + * {
    margin-top: ${dimensions.spacings.m};
  }
```

.changeset/refactor-content-spacing.md
```diff
@@ -0,0 +1,8 @@
+---
+'@commercetools-docs/ui-kit': minor
+'@commercetools-website/docs-smoke-test': minor
+---
+
+Markdown UI Kit: refactor content spacing to be able to render markdown inside
+
+Markdown UI Kit: refactor content spacing to be able to render markdown inside
```
emmenko, line 8: Duplicate sentence

```suggestion
Adjust content spacing for components that render MDX as children.
```

refactor-content-spacing.md
nkuehn: Good catch, I think I did not pay close enough attention while rebasing my branch.  Fixed it.

packages/ui-kit/src/components/markdown.js
nkuehn: Thanks for reminding me to the confusing parts of CSS fundamentals :-).  Vertical margin collapsing.

packages/ui-kit/src/components/markdown.js
nkuehn: ... except that it does not render the same in the smoke test site. 

One of the purposes of this wrapper is to _guarantee_ that the content does not have top or bottom margins, no matter what you throw in. 

But it's an indicator that the implementation is not consistent yet (e.g. the code example and  the headings still add margins by themselves).  I'll take a look whether it's realistic to cover all the cases in the components and then be able to not have to force-remove the very top and bottom margins.

nkuehn: OK, here's my issue:

The code block component is imported from a different module (a.k.a. file).  If I want to remove the margin that's hardwired into the component and instead address it as a child in the `TypographyChild` CSS to set the margin there, emotion fails because it does not seem to be able to resolve components that are symbols, i.e. only those that are in the same file. 

No idea, it's a CSS-in-JS specific problem that I could resolve by starting to work with manual classes or by pulling the CodeBlock into the Markdown file.  All not nice.

nkuehn: found it: https://github.com/styled-components/styled-components/issues/1326 

I'll address the emotion styled "Container" in the code block and not the CodeBlock itself

nkuehn:
FYI @davifantasia  will now take over this PR, review it and and apply it to the cards, too.

PR #455, opened 2020-05-23
API docs: experimental more compact Endpoint UI

nkuehn:
From my slight frustration over the usability regression in the API documentation and some practical issues I used my support on call shift to hack together this RFC how far we could go in making the API endpoint and type layouts more compact so that the actual information becomes more prominent and a typical API endpoint  fits on one desktop screen again.

some topics done here
 * back to regular off-the-shelf HTML tables that the browser optimizes for efficient space usage (since they have only two columns they still work decently on mobiles despite not being responsive)
 * Property name and type go into one line, only wrapping when needed. By using "inline-block" a lot it's guaranteed that whole segments of information wrap and not some random position between words. 
 * reduce spacing in some places
 * move response and OAuth scopes behind the label (all stay to the right of the label when wrapping)
 * word wrap optimization of the URL (wraps between host and path where the font changes, too)
 * move the GET / POST / DELETE badge from the title to the Endpoint block.  This reads more natural and fixes the problem that the layout is not working in the real APIs where endpoints typically don't have a separate display name and the name is in Markdown. 
 * labels consistently in light text (also in the table heads), previous pattern highlighted the labels and de-prioritized the actual content (e.g. the type name of the request body)
 * simplified label strings (e.g. "request body" and "response" ). 
 * include URI parameters into the visual block and remove the border around the colored API method block, only leaving the left bar. 
 * tune the wrapping behavior of the URL

**Consider this a starting point to discuss improvements and not an implementation to review on code level** 

TODO: somehow apply to the import API or a "real" platform API spec to get real-world input.


nkuehn:
@adinakleine  @stmeissner @Shecki  you may want to compare these two pages to get an impression of the difference in information density that could be possible if we want:

 * https://commercetools-docs-kit-git-nk-compact-api-docs.commercetools.now.sh/api-docs-smoke-test/endpoints/endpoints-for-resource
 * https://commercetools-docs-kit.commercetools.now.sh/api-docs-smoke-test/endpoints/endpoints-for-resource

I voluntarily tried to squeeze it as much as I can - it's easier to then open it up again than to have to imagine what would happen if.  

Here's two real-world examples for reference:
 * https://docs.commercetools.com/import-api/category-import#get-category-import-operations-of-a-sink
 * https://docs.commercetools.com/http-api-projects-carts#create-a-cart-in-a-store-beta (in comparison still pretty good and very concise)

stmeissner:
From the 1st look I like the dense version more

- the HTTP verb is next to the URL of the endpoint where it fits well and is also known from, e.g., swagger
- the data type is inline with the field name as in common code syntax, should still be on the next line when the field type is a Resource with a long name, like **CategoryKeyReference** and the field description contains enough text to fill more than one line
- also term **Request Body** feels more intuitive to me than **Request Representation**

Shecki:
i prefer the version with more white space.
it's way better for the eye.

nkuehn:
Sorry to have forgotten to tell that.  If possible please give separate and independent feedback on:

 * visual ("On the eye") look, for example density, light vs. heavy text 
 * structural:  what information is grouped into what block and "by meaning" positioned where
 * fundamental:  How important is the 
    * "micro" readability ("it's confusing visually and I can't see what label belongs to what")
    * vs.  "macro" understandability ("how do you want me to understand an API endpoint when its definition layout doesn't even fit a Full HD screen")

davifantasia:
My review from the perspective of a potential consumer of the API.

### Pros
* I like the colour coded left bar of each endpoint to be as long as the height of all information related to the endpoint. With the current design, sometimes I feel like I'm at the end of an endpoint when, in fact, there could be parameters and responses defined below.
* Moving the method type indicator like POST or GET below the description makes sense to me.

### Cons
* Endpoints are harder to read, I found that I needed to intentionally pay attention to where one logical part (e.g., oauth) ends and another (e.g., response) begins. So the spaces between are too small for me.
* I prefer that response and request types are rendered with all properties and not as a link. I wouldn't want to have to navigate to a different page to be able to use an endpoint.
* On certain screen sizes, for objects with one property, the row appears to have just one column and therefore one pice of information rather than 3 pieces of logical information. I prefer the current design and not to have all logical information on one row for scopes, responses, etc.
- This PR:
Screenshot 2020-05-25 at 16 26 44
- Current design implementation:
Screenshot 2020-05-25 at 16 27 16

### My suggestions
* In this PR, the location of the method type indicator (POST/GET) makes it look like it's out of place on smaller screens, I would try putting it above url.
* One way to compact is to reduce the font size, then a smaller padding would look better in my opinion.

nkuehn:
@adinakleine @stmeissner @Shecki @emmenko @TohruKohrita   thanks for your formal and informal feedback, which was all good.  I tried to incorporate it all now and start to feel good with the approach both visually and structurally.  

Happy to get another round of feedback!  I recommend to check more of the example content than just the two scenarios I linked above (e.g. ArrayTestType and all the Endpoints)

As a "side effect" I also realized a few missing features:
 * beta flag on properties
 * visual tags for the content constraints of properties
 * table for mobile viewports (no columns, one cell per type)
 * If the type of a request body is already documented in the site, it's not fully rendered but only the link (like in the existing docs)
 * render stuff as markdown that wasn't yet. 
 * allow passing a custom title if the type has no "displayName" set in the API spec
 * don't write an empty heading if neither title nor displayName is present

Notable visual changes that deviate from the design:
 * "secondary" instead of "muted" text color for the labels and hostname part of the URL
 * 95% white instead of 90% as the background for all inline code (that was a revelation in readability!)
 * "small" (14px) font instead of "body" (16px) for the property descriptions inside the tables.

davifantasia:
I suggest leaving the job of endpoint title text and placement to authors for these reasons

- Authors may want to include additional/unstructured information about an endpoint, and this typically should go after the title.
- It helps populate index navigation so end users can simply jump to an endpoint instead scrolling to find it.

nkuehn:
@davifantasia  I have implemented that the heading is only generated if a title is explicitly passed and if, then it's an H6.

nkuehn:
I have now also reverted the smaller font in the type property descriptions after consulting with @Shecki  that the small font is reserved for navigation and other site functionality and should not be used for content.

PR #456, opened 2020-05-24
feat(theme): use primary/secondary versions of prism themes for code blocks

emmenko:
image

### DIFF
--- a/websites/docs-smoke-test/src/content/components/code-examples.mdx
+++ b/websites/docs-smoke-test/src/content/components/code-examples.mdx
@@ -4,7 +4,7 @@ title: Code Examples
 
 # Multiple languages
 
-**Primary theme**
+**Primary theme (default)**
 
 <MultiCodeExample title="Multilanguage Code Samples">
   <CodeExample path="example.js" />
@@ -20,7 +20,7 @@ title: Code Examples
 
 # Single language
 
-**Primary theme**
+**Primary theme (default)**
 
 <CodeExample path="example.js" title="JavaScript Code Sample" />
 
--- a/websites/docs-smoke-test/src/content/components/code-examples.mdx
+++ b/websites/docs-smoke-test/src/content/components/code-examples.mdx
@@ -4,14 +4,14 @@ title: Code Examples
 
 # Multiple languages
 
-**Default theme: dark**
+**Primary theme**
 
 <MultiCodeExample title="Multilanguage Code Samples">
   <CodeExample path="example.js" />
   <CodeExample path="example.java"/>
 </MultiCodeExample>
 
-**With theme: light**
+**Secondary theme**
 
 <MultiCodeExample title="Multilanguage Code Samples" secondaryTheme>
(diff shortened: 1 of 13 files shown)


websites/docs-smoke-test/src/content/views/code-examples.mdx
```diff
@@ -0,0 +1,57 @@
+---
+title: Code Examples
```
emmenko, line 2: I moved this in the "views" section. There is still a simpler version of code examples in the "components" pages, which we use for VRTs.

websites/docs-smoke-test/src/content/components/code-examples.mdx
```diff
@@ -2,56 +2,28 @@
...
-## JSON
-
-<CodeExample path="example.json" />
-
-## PHP
-
-<CodeExample path="example.php" />
-
-## Shell
+<MultiCodeExample title="Multilanguage Code Samples" useThemeLight>
```
nkuehn, line 16: Sorry for being slightly pedantic, but I'd be happier with the grammatically correct word order in the author interface. The target group is tech writers, they are nerds about grammar ;-) 

```suggestion
<MultiCodeExample title="Multilanguage Code Samples" useLightTheme>
```

even better: don't introduce imperative words to the DSL with "use" but stick to declarative words like in the other components: 
```suggestion
<MultiCodeExample title="Multilanguage Code Samples" lightTheme>
```

emmenko: Makes sense, thanks. Will rename the prop.

emmenko: Thinking about it, I’m not sure using light/dark terminology is appropriate here and confusing with a possible website theme.

For instance, in case we do have light/dark theme for the entire website, would the light/dark theme for the code blocks still make sense as a name?

Instead maybe we can go with “primary” and “secondary” names. So the secondary version of code block is the “light” prism version.
The prop then would be `secondaryTheme`.

nkuehn: Yes, the name could get awkward with a dark theme, but I don't think it's as serious because a) we don't plan a dark theme for now and b) even if it would not be the default. 

Not sure if "secondaryTheme" is transmitting the meaning really good, but we haven't really defined the meaning.    If we go for a semantic flag I would not restrict it to the look / "theme" though and just call it a boolean "secondary". 

I can live with both "light" and "secondary",  let's use "secondary" (also for an easier path to later switch the default if users prefer light as primary).

emmenko:
Should we rename the prop to `secondary` or leave it as `secondaryTheme`?

davifantasia:
> Should we rename the prop to `secondary` or leave it as `secondaryTheme`?

I suggest `lightTheme`

emmenko:
I kept the prop as `secondaryTheme`. I feel like this is the less confusing name and a good middle ground.

PR #474, opened 2020-06-03
Better API example test cases

davifantasia:


websites/api-docs-smoke-test/source-raml/test/api.raml
```diff
@@ -147,33 +162,40 @@ uses:
...
         queryParameters:
           queryParameterOne:
             type: number
             description: some parameter to construct a combination of URI and query params
             required: false
+            example: 1
           queryParameterTwo:
             type: string
             required: false
+            example: string-path
```
nkuehn, line 181: For testing an implementation, it would be good to have a test case where one optional query parameter has no example set.   maybe this one?

websites/api-docs-smoke-test/source-raml/test/api.raml
```diff
@@ -53,30 +53,37 @@ uses:
...
           type: number
           required: false
+          example: 50
         offset:
           type: number
           required: false
+          example: 5
         collection:
           type: objects.ObjectTestType
           required: false
```
davifantasia, line 89: > For testing an implementation, it would be good to have a test case where one optional query parameter has no example set. maybe this one?

This has no example.

websites/api-docs-smoke-test/source-raml/test/api.raml
```diff
@@ -187,25 +209,32 @@ uses:
...
             type: string
+            example: bodyMassIndex
           filter.facets?:
             type: string
+            example: weight,height
           filter.query?:
             type: string
+            example: lastName = Doe
           facet?:
             type: string
```
davifantasia, line 233: > For testing an implementation, it would be good to have a test case where one optional query parameter has no example set. maybe this one?

This has no example.

PR #479, opened 2020-06-04
Index Nav:  Use available space, allow touch and wheel scrolling when sticky

nkuehn:
After running circles around trying to force all potential content pages into the available index nav with we (tech writers) realized that it's not actually making things better: 
  * smaller index nav is good, so we will try to make it fit for 90% of users and pages
  * reducing it to a bare minimum makes the site worse because the pages get harder to understand.  And there will always be someone with an older windows laptop and a fat task bar at the bottom that can't access the smallest index structure. 

**Decision**:  Allow scrolling if possible but don't show a scrollbar (design is not made for it) to get more of the edge case use cases covered without destroying the visual design. 

**Implementation**:   Not straightforward to find out but works fine.  CSS sticky is not exposing state about whether it's stuck or not.  So a special intersection observer is employed that observes a one pixel line over the top of the viewport.


emmenko:
Sorry it's a bit hard for me to understand the context here. Can you provide some visual clues as to what the problem is and how this PR solves that?

Thanks

nkuehn:
@emmenko I can show and explain you the problem live tomorrow in the call we proposed anyways.   Try this page to see the result: https://commercetools-docs-kit-git-nk-scollable-index-nav-fallback.commercetools.now.sh/docs-smoke-test/views/nested-headings 

Here's two gifs with just the sidebar (maybe still confusing but full screen would have gotten too big):

**Before**:  You can't find out what's at the end of the page and can't directly navigate there (screen shows the smallest we were able to make the Cart API docs to migrate but it's still to big):
cannot-access-indexnav

After: On normally small index navs there's not change in the UX or behavior but if a larger page doesn't fit on a smaller screen you can scroll inside the nav to directly see what's on the page:

allow-scrolling-when-index-nav-too-long

emmenko:
Thanks for the explanation, I get the point now.

I tested the behavior and I noticed a bunch of "regressions" when scrolling to the end of the page.
For instance, the index nav scrolls away instead of being "sticky".



Is this how it's intended to be?

nkuehn:
> For instance, the index nav scrolls away instead of being "sticky".

I don't think so, let me check.  I guess the culprit is that the height is set to `100vh` when set but obviously that does not work once the footer moves into the viewport and is visible only when  the index nav is small.

nkuehn:
@emmenko  I only gave the index nav a max-width of the viewport height now because scrolling is only needed when the index nav is higher than the page.  That fixes the problem you found. 

While testing I realized that it's still confusing when the index nav is long and decided to get back to normal accessibility behavior and show the scrollbars in that case.   They never appear when the nav fits the box and if they don't you actually need them as a visual hint to understand why some text ist cut out.

nkuehn:
PS @emmenko  and @davifantasia  please merge when you're ok with the result now.   
A release would help the docs team to see whether their migrated content works  if it contains this and the two-level heading frontmatter.

PR #483, opened 2020-06-05
fix(theme): release notes sort order

### LINKED ISSUE #481: Broken Sorting of Release Notes
I just realized a regression that must have come in in some past release.  

the Release notes are not sorted from newest to oldest any more.  In the smoke test page they follow no specific pattern, in the full migrated documentation they appear to be sorted from oldest to newest. 

Tasks here:
 * add release notes list page to the visual regression test
 * make sure that release notes are always sorted from newest to latest.

emmenko:
Fixes #481


emmenko:
Yes I wanted to do that as well.

PR #493, opened 2020-06-08
feat: card with image

davifantasia:


emmenko:
Probably it's better to wait for @nkuehn anyway to clarify those requirements. Just so that you don't have to spend time working on something that might not be valid.

nkuehn:
Hi, I'm back. 

The approach via a static query hook is consistent with how we handle such situations technically in other spaces, too so It's not bad per se.  The implementation is not really good though because the image names aren't safe to be unique without a path. 

But in this case there's two aspects that rather speak for passing a component:
 * it's not actually intended that arbitrary user provided images can be used as card images.  We rather intend to provide a finite predefined library like with the icons
 * the icon of the card is already implemented to be used via passing a component so the author experience would be more consistent if the image is passed the same way. 


Thanks!

packages/gatsby-theme-docs/src/components/card.js
```diff
@@ -23,6 +23,11 @@ const Container = styled.li`
   padding: ${designSystem.dimensions.spacings.m};
   border-radius: ${designSystem.tokens.borderRadiusForCard};
 `;
+const ImageContainer = styled.div`
+  svg {
+    width: 100%;
+  }
```
davifantasia, line 29: Currently, SVG image is centered instead of filling the width of the card. I found a solution [here](https://stackoverflow.com/a/9573878/2467003) but that means a `viewBox` needs to be included in the SVG.

nkuehn, line 29:
```suggestion
  svg, img {
    width: 100%;
    height: auto;
  }
```

The SVGs have a  height and widht hardcoded, that's how it often is when you get them from design.  If you just set the width to 100% but leave the hardwired height of 80px  then the height wins.  That's not SVG or ViewBox specific, that's plain old CSS. 

Plus, I suggest to addresss  too because the image component could be a pixel image, too.  Or, just target any first child.

websites/docs-smoke-test/src/@commercetools-docs/gatsby-theme-docs/overrides/markdown-components.js
```diff
@@ -1,4 +1,15 @@
...
+import CardBannerPlatformIcon from '../../../icons/card-banner-platform.svg';
 
-export default { ...addonComponents, DummyIcon: ConnectedSquareIcon };
+export default {
+  ...addonComponents,
+  DummyIcon: ConnectedSquareIcon,
+  CardBannerOSSIcon,
+  CardBannerMcDevsIcon,
+  CardBannerMcDocsIcon,
+  CardBannerPlatformIcon,
```
emmenko, line 14: I wouldn’t include the banner svgs in the core theme. For now they should be included and used only by the website itself. As far as I know, they are used only by the home website at the moment.

emmenko: Sorry never mind, I just realized this is the smoke test. 🤦‍♂️

Issue #477, opened 2020-06-03
Implement Release Notes Feed Component


nkuehn:
The home page will have the same preview aggregation of the release notes feeds like the existing home page has. 

Designed at the bottom here:  https://commercetools.invisionapp.com/d/main#/console/19262954/419244576/preview 

For the Merchant Center and Platform Release notes shown in the example design it would be possible to construct the data at build time because despite the home page being a separate microsite, it lives in the same repository.  

But that implementation would limit the implementation to such content and would impose future restrictions on the order the sites are built and also whether they are all monolithically built or not.  

@davifantasia and me therefore preferred an approach that reads the RSS feeds at runtime (means, client-side) to be able to later pull feeds from other sites and be able to decouple the builds if we want.   It's basically how Merchant Center also pulls the Release notes on the MC start page. 

Sine it's pretty generic then the component could be named just "RssFeed".  But we can call it ReleaseNotesFeed too if we think that helps. 

I don't think it's necessary to provide it in all MDX pages without an explicit import, if a use case comes up we can always use an import statement in an MDX page.

emmenko:
FYI: in the MC we fetch the RSS feed in the MC API and expose it through a GraphQL query.

We can maybe use the same API endpoint to fetch the data (no need to re-implement it).

nkuehn:
Do we really need to go through a separate API?   It feels a bit overengineered to me

 * I would like to avoid runtime and testing coupling of the documentation to the availability and hosting of the platform. 
 * the RSS is hosted on the same domain and (usually) hosting provider and can be safely fetched same-domain.  Also, they both fail or not.  

```
fetch("/api/release-notes.xml")
  .then(response => response.text())
  .then(str => new window.DOMParser().parseFromString(str, "text/xml"))
  .then(data => {
    const items = data.querySelectorAll("item");
    items.forEach(el => {
        // DO JSX FOO HERE and read name and url from item. 
        // {el.querySelector("link").innerHTML}
        // {el.querySelector("title").innerHTML}
    });
  });
```
Or, if we're lazy we can do it even less LOC and add another dependency:  https://www.npmjs.com/package/rss-parser

(wrapped in some sort of waiting spinner - maybe that part is available in the UI kit? ).

PR #497, opened 2020-06-11
feat: rss feeds

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/477

This uses [SWR](https://www.npmjs.com/package/swr) to get the feeds, and then parses the result with [rss-parser](https://www.npmjs.com/package/rss-parser)


nkuehn:
The header bar background is square and overlays the rounded area.   Otherwise LGTM.

Issue #470, opened 2020-06-02
Support HubSpot Tracking in addition to Google Analytics


nkuehn:
We're asked to also support the Hubspot tracking code in the documentation to give sales and marketing a strong signal to which users have been using the product documentation a lot in their evaluation of interesting prospects.  

Configuration should be analogous to the Google Analytics config we are including and there's little plugin for gatsby that looks decent and has no additional dependencies:  https://www.gatsbyjs.org/packages/gatsby-plugin-hubspot/

The tracking ID is not hardwired, we'll add it in the respective documentation repositories.

PR #498, opened 2020-06-12
feat: hubspot tracking

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/470


packages/gatsby-theme-docs/README.md
```diff
@@ -65,6 +65,8 @@ The project structure should contain at least the following files and folders:
 
   - `gaTrackingId` (_optional_): this is the Google Analytics tracking ID. For all sites hosted on the `docs.commercetools.com` domain the ID must be: `UA-38285631-3`.
 
+  - `hubspotTrackingCode` (_optional_): this is HubSpot tracking code.
+
     > For test websites the field should not be set.
```
nkuehn, line 70: This comment was originally belonging to gaTrackingId, but now belongs to the hubspotTrackingCode.    Please either reword it to GA and hubspot or remove it.   

A better solution would be to tell that if the config is not set nothing is tracked in both flags. .

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -198,6 +198,14 @@ module.exports = (themeOptions = {}) => {
           exclude: [],
         },
       },
+      pluginOptions.hubspotTrackingCode && {
+        resolve: 'gatsby-plugin-hubspot',
+        options: {
+          trackingCode: pluginOptions.hubspotTrackingCode,
+          respectDNT: true,
```
nkuehn, line 205: Please set this to false like we do in GA - there's no point in having differing config and in analytics it's important that all sites use the same config (all other ct sites are not setting the respectDNT flag, too).

Issue #406, opened 2020-04-27
Hide Empty Table Heads OR Grid view for many small items


nkuehn:
A common pattern in authoring markdown tables is to leave the header row of a table empty if you don't need a header.   Some renderers do not render the header row at all when the pattern is detected, but remark complies stricter and does not do that. 

I don't think it's a good idea to change the remark output but the Table frontend component could hide the header row if it's empty. 

Here's an example from the current docs:

![image](https://user-images.githubusercontent.com/4006139/80388654-6f384d00-88aa-11ea-8c84-8f2595b2cb22.png)

The underlying use case of the example is rather a grid than a table but I would prefer to use standard markdown for where it's possible over complex nested markup in JSX.

nkuehn:
Here's a proposal that would be semantically cleaner and could be rendered using something like a flexbox with row wrap

If we introduce such a custom variant  I suggest to design a custom style too to better match the semantics (e.g. striped lines aren't needed if lines don't mean that items belong together any more). 

The following would give any direct children of the Grid the row-flowing behavior, in this example the children are paragraphs. 

```mdx
<Grid>

[link 1](/foo/bar)

[link 2](/foo/bar)

[link 3](/foo/bar)

[link 4](/foo/bar)

[link 5](/foo/bar)

[link 6](/foo/bar)

[link 7](/foo/bar)

[link 8](/foo/bar)

[link 9](/foo/bar)

</Grid>
```mdx

Alternatively, a fixed convention that the items are in a list (but it would ignore all non-list content which is not ideal because it could confuse authors): 
```mdx
<Grid>

 - [link 1](/foo/bar)
 - [link 2](/foo/bar)
 - [link 3](/foo/bar)
 - [link 4](/foo/bar)
 - [link 5](/foo/bar)
 - [link 6](/foo/bar)
 - [link 7](/foo/bar)
 - [link 8](/foo/bar)
 - [link 9](/foo/bar)

</Grid>
```mdx

emmenko:
Yes I would go for the `Grid` component as well.

nkuehn:
Spoke to the design master today about the topic while speaking about something else.  If we build it, it would be fine to have no borders or nothing and have 16px vertical padding and 48px horizontal between the entries.

PR #499, opened 2020-06-12
feat: table with no column header in markdown is rendered without header

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/406

Implementation is different from expectation but I believe it is worth considering.


nkuehn:
@davifantasia  this has a linter error - please check the output of the checks yourself.

PR #504, opened 2020-06-16
feat: render rss feed using swr

davifantasia:
This uses [SWR](https://www.npmjs.com/package/swr) to get the feeds, and then parses the result with [rss-parser](https://www.npmjs.com/package/rss-parser)


nkuehn:
@davifantasia  is this intended to be a PR?   I suggest you finish the implementation in one branch of your choice.  

I'll make this a draft for now to avoid accidental merges. 

PS: as far as I know you are getting  a preview build for any push to a branch, no need to open a PR.

PR #521, opened 2020-06-25
fix(bullet lists): define list styling inside containers like card

### LINKED ISSUE #513: Bullets inside Cards: start on second level ("hollow" circle)
For some reason, bullet lists inside card bodies are visually starting on the second level as if they were nested bullets.  I think they should have less margin to the left when enclosed in a box and that they should be the normal first-level style.    

image

The reason why they are circles and not dots is that they are technically on the second level (the cards are a list for accessibility already).  I think the correct solution is to explicitly set the bullet style in our base components instead of leaving the user agent default.  

image

nkuehn:
fixes #513  (see issue for details)

 * a more robust and browser compatible implementation of indentations in lists
 * I chose to reduce the left padding of unordered list to have the identical style everywhere.  Numbered lists keep the existing padding to have speace for two-digit numbers.


nkuehn:
@davifantasia  spacing back to previous now - approval ok?

Issue #515, opened 2020-06-23
RssFeed(s) component:allow custom title text


nkuehn:
The RSS feeds component takes the complete title of the Feed for its title, but in the context of e.g. the home page that name contains too much redundancy:
 - "commercetools"
 - "Release Notes"
(see image below)

It would make the copywriting of the page cleaner if the title of the Feed could be overridden via an optional "title" property 

![image](https://user-images.githubusercontent.com/4006139/85455037-d36f4880-b59d-11ea-9917-da3675f6680c.png)

PR #522, opened 2020-06-25
rss feeds with custom title

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/515


.changeset/late-queens-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+"@commercetools-docs/ui-kit": patch
+"@commercetools-website/docs-smoke-test": patch
```
nkuehn, line 3: We're not "releasing" the test site so it does not need a patch change here

Issue #278, opened 2020-02-06
Implement "4-column" side-by-side body content layout


nkuehn:
### Background

The UX Design includes a layout for wide screens so that a typical browser window on a "Full HD  CSS pixels" monitor can show all examples next to the respective specification instead of under the spec.  

The design proposes the pattern based on user research. it is specified in the API documentation designs, but should be available as a general pattern of "spec and example side by side if the space allows".  

It's been designed using the API documentation as an initial pattern but it's intended to be used more broadly for body UI use cases that have "Pairs" of content that are not sequential.  For example, the interactive playground for App Kit components could be easier to use if the controls were next to the preview and not under it:  https://docs.commercetools.com/custom-applications/components/info-dialog

In addition, there are content elements that would profit from being allowed to take a very broad width in general if it's available:
 * Tables (readability does not suffer from being wide, it rather profits if there are many columns)
 * Card Grids
 * (more ideas?  e.g. Level 1 sections should go full width so that the divider line is full width)

**NOT** part of this despite being useful: 
 * actual "asides" (blocks that are allowed to float to the left of the block above them) - this is about blocks that are allowed to take full width themselves. 

### Implementation approach

I personally would start with an approach that generally increases the maximum width of the body column in the highest breakpoint and at the same time constrains all direct children of a `<section>` to the current maximum width.  Components can then indidvidually "take" more width by allowing themselves to become `max-width: 100%`.  This basically delegates the actual width usage to the individual components.   

### (Optional / Idea) allow manual / hand-written Side-by-Side content via MDX

The following Pseudo-MDX would be kind-of-nice for authors. It would instruct the layout to put its direct child blocks side by side and makes sure the column proportion is correct. 

````mdx
<SideBySide>
  
 * Random Markdown here
 * more random Markdown here

```javascript
const foo = "bar";
console.log(foo);   
```

</SideBySide>
````

nkuehn:
FYI  @emmenko  @davifantasia  it's not the first priority at the moment but would be good if you can "let it sink" and think about implementation possibilities and constraints

nkuehn:
Shot from the design for illustration:

![image](https://user-images.githubusercontent.com/4006139/73933496-dc0bfe80-48dc-11ea-9fb0-c1fcd8391b14.png)

nkuehn:
Finding from one hour tech time daddling:  

[...]

PR #529, opened 2020-06-29
Four-column / "largeDesktop" breakpoint and layout

nkuehn:
Addresses #278 

To get a better feeling for whether and how it works and how it could be achieved here's an implementation of the "largeDesktop" a.k.a "four-column" layout.   

It's configurable on site and page level, default is to not use the extrawide layout.  

It's implemented by giving the content column the full width but giving all the direct blocks in the body content a maximum width.  Individual components can then override that default and "take the full width".


packages/ui-kit/src/utils/side-by-side-content-styles.js
```diff
@@ -0,0 +1,21 @@
...
+    grid-template-columns: 1fr ${dimensions.widths.pageContentSmall};
+    grid-template-rows: auto;
+    grid-column-gap: ${dimensions.spacings.xl};
+    grid-row-gap: ${dimensions.spacings.xl};
+    justify-items: stretch;
+    align-items: start;
+  }
+`;
+
+export default sideBySideContainerStyle;
```
emmenko, line 21: PS: this should be a styled component, no need to export only the styles.

```
const SideBySide = styled.div`
  ...
`
```

packages/ui-kit/src/design-system.js
```diff
@@ -129,9 +130,14 @@ export const dimensions = {
...
-    pageContentSmallWithMargings: `calc(${pageWidthSmall} + 16px * 2)`,
-    pageContentWithMargingsAndPageNavigation: `calc(${pageWidth} + 32px * 2 + ${navbarWidth})`,
+    pageContentWide: `calc(${pageWidthSmall} + ${asideColumnWidth})`,
+    pageContentWideMax: `calc(${pageWidth} + ${asideColumnWidth} + 32px)`,
+    pageContentWithMargins: `calc(${pageWidth} + 32px * 2)`,
+    pageContentSmallWithMargins: `calc(${pageWidthSmall} + 16px * 2)`,
+    pageContentWideWithMargins: `calc(${pageWidthSmall} + ${asideColumnWidth} + 32px * 3)`,
+    pageContentWideWithMarginsMax: `calc(${pageWidth} + ${asideColumnWidth} + 32px * 3)`,
+    pageContentWithMarginsAndPageNavigation: `calc(${pageWidth} + 32px * 2 + ${navbarWidth})`,
+    pageContentWideWithMarginsAndPageNavigation: `calc(${pageWidthSmall} + ${asideColumnWidth} + 32px * 3 + ${navbarWidth})`,
```
emmenko, line 140: Are you using all those new sizes?

packages/ui-kit/src/utils/side-by-side-content-styles.js
nkuehn: > this should be a styled component, no need to export only the styles.

It's used in more than one component (here e.g. the API type and the manual `<SideBySide>` Block, plus more in the future) _and_ is sensitive to the block nesting levels (it needs to be the direct element under the MDX `<section>` wrappers to override the 770px constraint I liked it better this way, more like a mixin. 

I can try whether it works as a styled component without having to refactor too much though, maybe it was theoretical.

packages/ui-kit/src/design-system.js
nkuehn: Good question :-).  They aren't all new, just the ones with "Wide" in them.  Not sure why all lines are marked as new in the diff. I'll cross check, would not be a surprise if some turned out to not be used.

nkuehn:
> In the side-by-side layout, is the text supposed to span across?

No, that's a bug.  That specific text is not markdown but the description of the type.  Thanks!

packages/ui-kit/src/utils/side-by-side-content-styles.js
emmenko: In the current implementation you're passing the styles to a div element anyway, so from that point of view nothing changes if you export it as a component and replace the div elements.

nkuehn: done, it's a component now

packages/ui-kit/src/design-system.js
nkuehn: @emmenko  I found out why the all look new:  I refactored a spelling error "Margings->Margins"  in the existing names, so they are all changed.  Plus, two were actually not used any more, I removed them.

nkuehn:
@emmenko @davifantasia I'm unsure how to best make it an optional "feature".  There seem to be too many possible ways - or I'm just confused. 

If it's an option of the theme, how would I get it through the chain of graphQL, hooks / gatsby-browser.js to the sTyles?

If it's an option for a given chapter in a site, how could that be passed into the CSS of the page inset layout?

davifantasia:
@nkuehn 

> If it's an option of the theme, how would I get it through the chain of graphQL, hooks / gatsby-browser.js to the sTyles?

This field would be queried for in the GetSiteData query, which can then be accessed by any component using the `useSiteData` hook. The GetSiteData query is located in the `ThemeProvider` component.

> If it's an option for a given chapter in a site, how could that be passed into the CSS of the page inset layout?

It can be passed through frontmatter. Which is queried in the templates. From The templates the values can then be passed down eventually to the component that needs to render differently.

> A third (and one with easiest implementation) option is passing a boolean prop to the component (like `ApiType`) that needs to be rendered a little differently. This can also be implemented along with the previously mentioned approach top.

**Note**:
Using the prop in styled component is a little different than regular react components so here's an example of usage:

```
// styled component
const H1 = styled.h1`
  color: ${(props) => props.useBlue ? 'blue' : 'red'}
`;

// regular react component
const Title = (props) => <H1 useBlue={props.useBlue}>The is a Title</H1>
```

packages/gatsby-theme-api-docs/src/components/description.js
```diff
@@ -0,0 +1,26 @@
...
+      <Markdown.TypographyContainer>
+        {markdownFragmentToReact(props.children)}
+      </Markdown.TypographyContainer>
+    ) : (
+      props.children
+    )}
+  </DescriptionContainer>
+);
+Description.propTypes = {
+  children: PropTypes.any,
```
davifantasia, line 23:
```suggestion
  children: PropTypes.node,
```

packages/gatsby-theme-api-docs/src/components/type/children.js
```diff
@@ -22,30 +28,14 @@ const Children = ({
...
-        <Properties
-          apiType={apiType}
-          parentDiscriminator={parentDiscriminator}
-        />
+      {apiType.description && renderDescriptionBelowProperties ? (
+        <Description>{apiType.description}</Description>
       ) : null}
 
-      {apiType.examples ? <Examples examples={apiType.examples} /> : null}
+      {apiType.enumeration ? <Enum values={apiType.enumeration} /> : null}
```
davifantasia, line 38: Previously, descriptions and enums were either rendered on top or below properties, so does this change mean we want descriptions to always be rendered last?

packages/gatsby-theme-docs/src/layouts/internals/layout-page-navigation.js
```diff
@@ -34,13 +34,6 @@ const GridContainer = styled.div`
   @media screen and (${designSystem.dimensions.viewports.largeTablet}) {
     display: block;
     grid-area: page-navigation;
-    width: ${designSystem.dimensions.widths.pageNavigation};
-  }
-  @media screen and (${designSystem.dimensions.viewports.laptop}) {
-    width: ${designSystem.dimensions.widths.pageNavigationSmall};
-  }
-  @media screen and (${designSystem.dimensions.viewports.desktop}) {
-    width: ${designSystem.dimensions.widths.pageNavigation};
```
davifantasia, line 43: Please what is the reason these are being removed?

packages/gatsby-theme-api-docs/src/components/description.js
nkuehn: I chose any because it can be a string or a node.  But maybe I'm wrong here and a string would go through as a node.

packages/gatsby-theme-api-docs/src/components/type/children.js
nkuehn: This was more an accident I think .  I suggest to leave it unchanged here and you do it proper in your Enums PR

packages/gatsby-theme-docs/src/layouts/internals/layout-page-navigation.js
nkuehn: When digging into the whole grid logic and how space is used I realized that on tablet and laptop the constraint is redundant because the space is already constrained by the grid. 

On desktop, I found that it's not necessary to force a smaller navigation when the space is available - i find it easier on the eye and visually nicer if the entries are one line:

image

nkuehn:
@davifantasia  I addressed all your feedback, OK for an approval so it can go into design review next week?

packages/gatsby-theme-docs/src/components/side-by-side.js
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+const SideBySide = (props) => {
+  const pageData = usePageData();
+  if (pageData.allowWideContentLayout)
+    return <SideBySideContainer>{props.children}</SideBySideContainer>;
+  return <>{props.children}</>;
+};
+SideBySide.propTypes = {
+  children: PropTypes.node,
```
emmenko, line 38:
```suggestion
  children: PropTypes.node.isRequired,
```

packages/gatsby-theme-docs/src/components/side-by-side.js
```diff
@@ -0,0 +1,57 @@
...
+const FullWidthContainer = (props) => {
+  const pageData = usePageData();
+  return (
+    <div css={pageData.allowWideContentLayout ? fullWidthStyle : null}>
+      {props.children}
+    </div>
+  );
+};
+FullWidthContainer.propTypes = {
+  children: PropTypes.node,
```
emmenko, line 54:
```suggestion
  children: PropTypes.node.isRequired,
```

packages/gatsby-theme-docs/src/layouts/internals/page-content-inset.js
```diff
@@ -1,9 +1,16 @@
 import styled from '@emotion/styled';
-import { designSystem } from '@commercetools-docs/ui-kit';
+import { designSystem, Markdown } from '@commercetools-docs/ui-kit';
 
 const PageContentInset = styled.div`
```
emmenko, line 4: Nit: can you maybe leave some comments explaining the reasons for the `section` selectors? So that we know in the future what the reason for that is. Thanks

packages/gatsby-theme-docs/src/templates/homepage.js
```diff
@@ -4,32 +4,37 @@ import { graphql } from 'gatsby';
...
-          <div>
-            <MDXRenderer>{props.data.contentPage.body}</MDXRenderer>
-          </div>
-        </Markdown.TypographyPage>
-      </MDXProvider>
-    </LayoutContentHomepage>
+    <PageDataContext.Provider value={props.data.contentPage}>
+      <LayoutContentHomepage
+        pageContext={props.pageContext}
+        pageData={props.data.contentPage}
```
emmenko, line 17: Nit: since you've put the data into a react context, we might not need to explicitly pass the same data as a prop.
Could be also be refactored as a follow up.

websites/api-docs-smoke-test/src/content/large-desktop-layout.mdx
```diff
@@ -0,0 +1,57 @@
+---
+title: Large Desktop Layout
+description: Test components that can flow into two content columns
+wideLayout: true
```
emmenko, line 4: This shouldn't be needed, since the website has already the `allowWideContentLayout` option enabled.

websites/api-docs-smoke-test/src/content/regular-desktop-layout.mdx
```diff
@@ -0,0 +1,58 @@
+---
+title: Regular Desktop Layout
+description: Test components that can flow into two content columns when not allowed to
+wideLayout: false
+---
+
+<Subtitle>
+
+Tests the responsive behavior in "largeDesktop" responsive sizes. It has the `widelayout` frontmatter set to `true`.
```
emmenko, line 9:
```suggestion
Tests the responsive behavior in "largeDesktop" responsive sizes. It has the `widelayout` frontmatter set to `false`.
```

websites/api-docs-smoke-test/src/content/regular-desktop-layout.mdx
```diff
@@ -0,0 +1,58 @@
+---
+title: Regular Desktop Layout
+description: Test components that can flow into two content columns when not allowed to
```
emmenko, line 3:
```suggestion
description: Test components that can not flow into two content columns when the option is disabled
```

emmenko:
@nkuehn also remember to add a changeset

packages/gatsby-theme-api-docs/src/components/description.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import { designSystem, Markdown } from '@commercetools-docs/ui-kit';
+import { markdownFragmentToReact } from '@commercetools-docs/gatsby-theme-docs';
+
+const DescriptionContainer = styled.div`
```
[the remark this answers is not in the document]
nkuehn: I find the specific names much more helpful when inspecting the component tree at runtime, just like we name file names rather specific and not assuming you know the folder path because you often see them standalone with no context.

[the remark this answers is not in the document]
emmenko: Hmm that's debatable 😄

You name files specifically because it's easier to find them when searching e.g. with the editor.
When you inspect the component tree, you always look for the main component first, then then components you see underneath it are obviously part of the component.

There is also another thing to consider. The generated CSS classes have much longer and repetitive names like this.

packages/gatsby-theme-api-docs/src/components/description.js
```diff
@@ -0,0 +1,26 @@
...
+    ) : (
+      props.children
+    )}
+  </DescriptionContainer>
+);
+Description.propTypes = {
+  children: PropTypes.node,
+};
+
+export default Description;
```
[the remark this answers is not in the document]
nkuehn: It's used for all plain text descriptions in various API definition concepts (endpoints, types, etc). And it lives in the api-docs package.   It's something like a "Paragraph", that's it.   Not sure what a more specific but still correct name would be.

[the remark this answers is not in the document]
emmenko: `ApiDescription`?

It's just a suggestion though, I'm fine to keep it as-is if it makes more sense to you.

packages/gatsby-theme-docs/src/templates/homepage.js
nkuehn: I first started that too but then went back because there was not practical advantage (the props are just passed down one time) and this PR has enough "Noise" already. 

What also made me go back is the disadvantage that you don't have prop type validation in the child component any more (resisted throwing typescript at the whole thing ;-)

websites/api-docs-smoke-test/src/content/large-desktop-layout.mdx
nkuehn: It's a test case to have the "true" option at least parsed and passed down once - if it's not enabled site-wide you can override it per page. A better test case would be to have one page in the `docs-smoke-test` use the frontmatter with "true" though.

packages/gatsby-theme-api-docs/src/components/description.js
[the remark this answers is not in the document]
nkuehn: addressed since it's just a code style preference. Changed a few other existing places too on that occasion for consistency.

packages/gatsby-theme-api-docs/src/components/description.js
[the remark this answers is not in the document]
nkuehn: An Api is an Endpoint, bu this applies to Types, Enums, Security Schemes etc. too in RAML so I struggled.  Will leave it this way for lack of better ideas.

packages/gatsby-theme-docs/src/components/side-by-side.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import { css } from '@emotion/core';
+import { dimensions } from '@commercetools-docs/ui-kit/src/design-system';
+import { usePageData } from '../hooks/use-page-data';
+
+// allows the first and second child of this container to float side by side in large desktop sizes
+const SideBySideContainer = styled.div`
+  margin-top: 0 !important; /* a stupid "important-war" with the UI kit's stack components */
```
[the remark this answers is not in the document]
nkuehn: Using important is bad practice (IMHO) but to be able to use UI kit stacks I had to use !important to override the !important in the UI Kit.  Which is a bad idea, too concerning what a UI kit is good for.  

A Stack that's more optimized for responsive body content would be helpful instead of having to do hacks like this.  

If you have a good idea how to solve this layout situation more elegantly I'm very happy, as you see in the comment it's not making me happy.  I wanted to use UI kit stuff because you asked for it in all previous code but it's not an ideal fit here. 

**Problem Statement**: The Side By Side content is sometimes an entry in the UI Kit stack (e.g. in API docs).  the UI kit stack forces the top spacing and uses !important. 
But the SideBySide is responsive, i.e. the second item can sometimes be to the right of the first and sometimes under it.  So we need different spacing "logic" here - the UI Kit stacks and inlines are not intended for responsive content (without this override the right item would not be top aligned with the left item).

packages/gatsby-theme-docs/src/components/side-by-side.js
```diff
@@ -0,0 +1,57 @@
...
+    display: grid;
+    grid-template-columns: 1fr ${dimensions.widths.pageContentSmall};
+    grid-template-rows: auto;
+    grid-column-gap: ${dimensions.spacings.xl};
+    grid-row-gap: ${dimensions.spacings.xl};
+    justify-items: stretch;
+    align-items: start;
+    > :only-child {
+      /* if there aren't actually two, let the first take the space: */
+      grid-column: auto / span 2;
```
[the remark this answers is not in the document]
nkuehn: If there aren't two children of a side by side it does not force it into two columns although the right one is empty.  Instead, it gives it the regular pageContent max width. 

the API docs are data driven, if an endpoint or type does not have an example there is nothing for the right column (this is not my invention, it's in production data).

nkuehn:
@emmenko  I tried to address all feedback, the remaining thing that is not nice is the !important CSS.  

Thanks for the thorough feedback!

emmenko:
@nkuehn thanks, this looks good to me. I pushed a commit to improve a bit the comments and docs around this feature.

packages/gatsby-theme-docs/src/components/side-by-side.js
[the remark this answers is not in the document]
emmenko: Thanks for the explanation. I added the reasoning as a comment so that it's clear what's it for.

PR #525, opened 2020-06-29
feat: include canonical link in page header

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/512

PR #531, opened 2020-06-30
feat(theme): allow to optionally enable canonical urls

### LINKED ISSUE #512: Implement Canonical Links in HTML Head
Currently, the Head HTML does not contain a "canonical" URL that instructs search crawlers what the official unique address is and that way makes sure that e.g. our preview addresses never get indexed separately. 

Also, it allows to make sure filter states in the release notes don't get indexed separately by not including the query. 

It's possible to do it directly in the templates ( via React Helmet) or  using a plugin like https://www.gatsbyjs.org/packages/gatsby-plugin-react-helmet-canonical-urls/

emmenko:
Superseeds #525 
Closes #512


emmenko:
Yeah, I suppose the linkinator needs to be configured. Do you want to open an issue first? We can also contribute to the feature if it's deemed appropriate.

nkuehn:
> Do you want to open an issue first? We can also contribute to the feature if it's deemed appropriate.

I can ask whether we can sponsor something, but we can also try to directly contribute. The maintainer works at google so I guess it's not a sponsoring-driven model.  As opposed to the previous library we used the codebase looks pretty sane to me, so contributing would be nice. 


existing issues in the repo that matter to us:
 * https://github.com/JustinBeckwith/linkinator/issues/166  (CLI output without localhost, nice to have)
 * https://github.com/JustinBeckwith/linkinator/issues/79  (fragment identifier / anchor support, that's **pretty important** to us - in the current repo 99% of the link check failures are wrong fragment identifiers)

nkuehn:
PS: I opened an issue in the linkinator project:  https://github.com/JustinBeckwith/linkinator/issues/175

nkuehn:
PPS:  HTMLProofer has a fancy script documented that dynamically ignores new files but i find it insufficient:  https://github.com/gjtorikian/html-proofer#ignoring-new-files

emmenko:
Let's put this PR on hold / draft until we decide how to proceed.

cc @nkuehn

nkuehn:
@emmenko : @davifantasia  and me decided to try to contribute the necessary feature to linkinator.  If they are responsive we need to contribute more anyways (hash validation).

davifantasia:
@nkuehn @emmenko enabling canonical links works with the skip attribute. I believe we can now review and close this.

For the hash fragment validation I would open a different PR for that.

nkuehn:
> @nkuehn @emmenko enabling canonical links works with the skip attribute. I believe we can now review and close this.

Sorry, @davifantasia  - can you clarify what you mean with this?  getting a link checker that does not fail on these would require doing a change here to add the skip attribute but why close this PR altogether?

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -163,6 +164,12 @@ module.exports = (themeOptions = {}) => {
        */
       'gatsby-plugin-sharp',
       'gatsby-plugin-react-helmet',
+      pluginOptions.enableCanonicalUrls && {
+        resolve: 'gatsby-plugin-react-helmet-canonical-urls',
+        options: {
+          siteUrl: `https://${productionHostname}`,
```
nkuehn, line 170:
```suggestion
          siteUrl: `https://${productionHostname}`,
          noTrailingSlash: true,
          noHash: true,
          noQueryString: true,
```
https://www.gatsbyjs.com/plugins/gatsby-plugin-react-helmet-canonical-urls/#how-to-use

Esp. for the release notes overview we must not have query strings in the URL.

packages/gatsby-theme-docs/utils/default-options.js
```diff
@@ -11,6 +11,7 @@ const defaultOptions = {
   additionalPrismLanguages: [],
   overrideDefaultConfigurationData: [],
   addOns: [],
+  enableCanonicalUrls: false,
```
nkuehn, line 14: default should be true since it's a best practice to have canonical links

nkuehn (changes requested):
Result of analysing the situation with @davifantasia :
 * we prefer to continue working with this plugin
 * that means we cannot add a custom skip attribute to the canonical link
 * that means in linkinator we don't need support for skip attributes but rather explicitly for canonical links
  * the linkinator exclusion of "docs.commercetools.com" is necessary evil now but needs to go away because we need it for e.g.  linking to custom apps docs.

nkuehn:
@davifantasia  do you want to take over checking my feedback here - this is also needed for migration

PR #533, opened 2020-06-30
refactor: rename oppressive terms ✊🏾

emmenko:


.changeset/six-cups-promise.md
```diff
@@ -0,0 +1,8 @@
+---
+'@commercetools-docs/gatsby-theme-api-docs': patch
```
nkuehn, line 2: Since it's changing the configuration interface, isn't this strictly spoken a major release?

emmenko: Not really, the deprecated option is still there for backwards compatibility

nkuehn: ok, I oversaw that.

Issue #526, opened 2020-06-29
API docs: Basic Enum Type Styling


nkuehn:
If a Type is just an enumeration of Values (e.g. Strings) the current output is just a plaintext list.  For example see the test page `/types/enum-test`

As far as I remember the design proposed a definition list style for the values and their respective descriptions (for the descriptions we had defined an annotation syntax).  Currently the implementation is re-used in places that actually need the comma separated string without descriptions so maybe some refactoring is needed.

PR #535, opened 2020-07-02
refactor: render enums as a definition list

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/526


packages/gatsby-theme-api-docs/src/components/type/children.js
```diff
@@ -15,29 +15,39 @@ const Children = ({
...
-      {apiType.description && renderDescriptionBelowProperties ? (
-        <Description>{apiType.description}</Description>
-      ) : null}
-
-      {apiType.enumeration ? <Enum values={apiType.enumeration} /> : null}
+      {renderDescriptionBelowProperties && renderDescriptionAndEnums()}
     </SpacingsStack>
   );
+
+  function renderDescriptionAndEnums() {
```
emmenko, line 36: Nit: if you're rendering it like this, you can just make it a component.

```js
function DescriptionAndEnums() {}

<DescriptionAndEnums />
```

Issue #538, opened 2020-07-03
Allow centering the top nav when expanded (like search, for centered home page)


nkuehn:
We oversaw that the top nav has a similar issue to the search results when used on a custom centered layout: It's left-aligned on the whole screen width.  Either it's possible directly in the custom page or we need to provide a prop from the kit to center it.

PR #544, opened 2020-07-08
feat: option to center top menu

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/538

### DIFF
--- a/packages/gatsby-theme-docs/src/components/top-menu.js
+++ b/packages/gatsby-theme-docs/src/components/top-menu.js
@@ -1,6 +1,8 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { useStaticQuery, graphql } from 'gatsby';
 import styled from '@emotion/styled';
+import { css } from '@emotion/core';
 import SpacingsInline from '@commercetools-uikit/spacings-inline';
 import SpacingsStack from '@commercetools-uikit/spacings-stack';
 import { designSystem } from '@commercetools-docs/ui-kit';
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-docs/src/components/top-menu.js
```diff
@@ -169,7 +177,7 @@ const TopMenu = () => {
           event.stopPropagation();
         }}
       >
-        <ContentGrid>
```
emmenko, line 172: This grid component should stay, as all the children are positioned accordingly to the grid. If you change the container to a block element, the children will be break the layout most likely.

packages/gatsby-theme-docs/src/components/top-menu.js
```diff
@@ -169,7 +177,7 @@ const TopMenu = () => {
...
+        <div css={props.centered ? centeredContainerStyle : contentGridStyle}>
```
davifantasia, line 180: @emmenko the grid still remains but through css prop now.

emmenko: But the centered container style is not a grid 🧐

davifantasia: I'm not sure what you mean, but currently, the children's layout is not affected from what I see - I get exactly what I want by making it a block element, and children are properly centered.

packages/gatsby-theme-docs/src/components/top-menu.js
nkuehn: @emmenko I just inspected the built site: The content has its own grid in the "Columns" component.  

This "ContentGrid" is an extra wrapper to position it the same way the main page content is positioned.

PR #552, opened 2020-07-10
test: initial e2e test setup for gatsby-theme-api-docs

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/555


packages/gatsby-theme-docs/src/components/link.js
```diff
@@ -96,10 +96,10 @@ export const ExternalSiteLink = (props) => (
  */
 const PureLink = (extendedProps) => {
   const siteData = useSiteData();
-  const { location, noUnderline, ...props } = extendedProps;
+  const { location, noUnderline, ariaLabel, ...props } = extendedProps;
```
emmenko, line 99: There should be no need to pass an `aria-label`. You can access the link by the text, for example `findByText('This is a link')`

cypress/integration/api-docs-smoke-test/api-type-soft-linking.js
```diff
@@ -0,0 +1,14 @@
+import { URL_API_DOCS_SMOKE_TEST } from '../../support/urls';
+
+describe('API type soft linking', () => {
+  it('should go to page of a type, then find the page of linked property type', () => {
+    cy.visit(URL_API_DOCS_SMOKE_TEST);
+    cy.findByLabelText('Open ObjectTestType').should('exist');
```
emmenko, line 6: Using `findBy` implicitly checks for the element existence, so there is no need to specify that. You can just click on it.

.changeset/polite-melons-hide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@commercetools-docs/gatsby-theme-api-docs": patch
+"@commercetools-docs/gatsby-theme-docs": patch
+---
+
+chore: initial e2e test setup for gatsby-theme-api-docs
```
emmenko, line 6: There should be no need to release a new version right? We're only adding e2e tests here.

cypress/integration/api-docs-smoke-test/api-type-soft-linking.js
```diff
@@ -0,0 +1,14 @@
...
+    cy.findByLabelText('Open ObjectTestType').click();
+    cy.findByLabelText('ObjectTestType definition').should('be.visible');
+    cy.findByLabelText('Link to ArrayTestType').should('exist');
+    cy.findByLabelText('Link to ArrayTestType').click();
+    cy.findByLabelText('ArrayTestType definition').should('be.visible');
+    cy.percySnapshot();
```
emmenko, line 12: Just to confirm, this will take a snapshot of the "ArrayTestType" page. Is this what you want to have?

packages/gatsby-theme-docs/src/components/link.js
davifantasia: > There should be no need to pass an `aria-label`. You can access the link by the text, for example `findByText('This is a link')`

A link text can exist in the main navigation as well as a page, doing this for me ensures I am getting the right text.

Do you still think an aria-label is not needed?

cypress/integration/api-docs-smoke-test/api-type-soft-linking.js
davifantasia: > Just to confirm, this will take a snapshot of the "ArrayTestType" page. Is this what you want to have?

Yes

packages/gatsby-theme-docs/src/components/link.js
emmenko: If there are multiple links in the page with the same text, you can narrow down the selection by selecting links within certain containers. There are different ways to do that with Cypress.

Still, an `aria-label` is still not necessary, as you have probably the same issue selecting the right link.

Issue #556, opened 2020-07-13
API docs: Render constant (single value enum) pattern in type properties


nkuehn:
E.g. in discriminator values of specific types we're just rendering the generic information of the discriminator field instead of the specific string to be sent. 

(use case: CategoryImportSink)

davifantasia:
Just leaving this here for my use:

Use ImportRequest and CategoryImportRequest.

Previous discussion about rendering constants:

Use setChanges in table design - https://commercetools.invisionapp.com/d/main#/console/17784193/390031793/preview

PR #561, opened 2020-07-15
API docs: Render constant (single value enum) pattern in type properties

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/556


packages/gatsby-theme-api-docs/src/components/type/properties/rows/description.js
```diff
@@ -38,30 +37,46 @@ const DescriptionText = styled.span`
...
-              <Info key={index}>
-                {info}
-                <InfoValue>{value}</InfoValue>
+      {isConstantLike ? (
+        <DescriptionText>
+          <div>
+            <Markdown.InlineCode>{property.enumeration[0]}</Markdown.InlineCode>
+          </div>
+          <div>
+            Only &quot;{property.enumeration[0]}&quot; value must be assigned to{' '}
```
nkuehn, line 52: These properties can have a normal description text like any others and it must be rendered here like a normal description.  

instead of the div:
```jsx
<DescriptionText>
      {markdownFragmentToReact(property.description)}
</DescriptionText>
```

websites/api-docs-smoke-test/source-raml/test/types/unions.raml
```diff
@@ -3,25 +3,28 @@ usage: Unions test types.
...
+
 types:
   UnionLikeType:
     description: |
       A pattern of inheritance that is to be represented like a union, i.e. this parent type would not be shown with a property list on its own.
     type: object
     discriminator: kind
     properties:
-      kind: string
+      kind: enums.EnumWithDescriptionExample
```
nkuehn, line 16: Can you set up the examples of the constantLike in a way that is independent of the UnionLike and cover a case where the type of the constant is not a string but for example an integer?

packages/gatsby-theme-api-docs/src/components/type/properties/rows/description.js
```diff
@@ -38,30 +37,46 @@ const DescriptionText = styled.span`
...
+        </DescriptionText>
+      ) : (
+        <>
+          <DescriptionText>
+            {markdownFragmentToReact(property.description)}
+          </DescriptionText>
+          <SpacingsInline>
+            {renderEnums ? (
+              <Info>
+                Value must be one of: {property.enumeration.join(', ')}
```
nkuehn, line 64: - the text is still cumbersome.  For inspiration: Stripe has a mix of automated and hand-authored wordings here: https://stripe.com/docs/api/accounts/object    "Possible values: X, Y, Z",  "Values may be X, Y or Z", "Can be X, Y or Z".   For a visual component like here I like "Possible values: " best (better than the "Allowed values: " I originally proposed). 
- Please try to render the values depending on the property type (string vs. number) and also apply that logic in other places like "default".

PR #578, opened 2020-07-22
Content spacing that binds elements better

### LINKED ISSUE #575: Design review: content spacing
* smaller space between headers and first block after the header
* list elements - smaller spacing above and regular spacing below

nkuehn:
resolves #575 

- only small spacing after headings
- small spacing above lists and regular below  (instead of normal above and XXL below).
- no extra spacing after code blocks

Code order needed to be changed to support nesting

@Shecki can't preview it in the full site easily but all changes can be found in the smoke tests.


packages/ui-kit/src/components/markdown.js
```diff
@@ -119,6 +98,31 @@ const Dd = styled.dd`
...
+  ${containerStyles};
+  background-color: ${colors.light.surfaceQuote};
+  border-left: 1px solid ${colors.light.borderHighlight};
+  border-radius: 0 ${tokens.borderRadiusForBlockquote}
+    ${tokens.borderRadiusForBlockquote} 0;
+  color: ${colors.light.textFaded};
+  font-size: ${typography.fontSizes.small};
+  padding: ${dimensions.spacings.xs} ${dimensions.spacings.s};
+`;
+
```
nkuehn, line 125: unchanged, had to be moved behind "containerStyles"

packages/ui-kit/src/components/markdown.js
```diff
@@ -111,6 +75,21 @@ const Dl = styled.dl`
...
+*/
+const containerStyles = () => css`
+  > * + * {
+    margin-top: ${dimensions.spacings.m};
+  }
+  > * + ${Ul}, > * + ${Ol}, > * + ${Dl} {
+    margin-top: ${dimensions.spacings.s};
+  }
+`;
+
```
nkuehn, line 92: "containerStyles" had to be moved down here to be able to have access to the Ul, Ol and Dl variables for nested lists in containers.

nkuehn:
I reviewed the result in a call wit Sven - OK this way.

Issue #558, opened 2020-07-14
Rewrite e2e test for soft linking to address type below the fold


davifantasia:
Right now, the test uses a page with just one type, and tests to ensure type is rendered on the page. This does not address the issue of types that may lie deep somewhere in the page - if a linked anchor is not working properly, for example, a page is linked but anchor not found, the current test does not cover that.

Issue #559, opened 2020-07-14
Decision: Consider dropping URN as an anchor pattern


davifantasia:
We recently experienced a broken page using soft linking on api site in commercetools-docs repo after an update, this is because the urn pattern we use in our anchor links is not valid when querySelector is used. Error returned was:

`DOMException: Failed to execute 'querySelector' on 'Document': '#ctp:import:type:ImportResourceType' is not a valid selector.`

See these link for more information on the reason this happened:
* https://stackoverflow.com/a/37271406/2467003

nkuehn:
@davifantasia I'm not sure what this issue "wants to be" - fix the implementation or reconsider the anchor pattern to avoid future similar changes. 

the colons are the issue but they are the definition of how a URN is built. So if this is about changing the pattern then the title would rather be "consider dropping URN as an anchor pattern"

davifantasia:
> the colons are the issue but they are the definition of how a URN is built. So if this is about changing the pattern then the title would rather be "consider dropping URN as an anchor pattern"

Thanks @nkuehn , this is a much better title.

nkuehn:
@davifantasia  from our conversation:  The remaining problem is in cypress only, here's the FAQ for the topic: https://docs.cypress.io/faq/questions/using-cypress-faq.html#How-do-I-use-special-characters-with-cy-get

nkuehn:
That means: no, we're not changing the spec for now.

PR #579, opened 2020-07-23
test: improve integration test for soft linking

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/558


davifantasia:
On hold until API type id is standardised.

This is needed because Percy does not currently support scroll positions (https://github.com/percy/percy-cypress/issues/2#issuecomment-533316240). The alternative is to test using element id, but the id cannot be found by Cypress, the [issue](https://github.com/commercetools/commercetools-docs-kit/issues/559) related to this ID needs to be resolved first.

davifantasia:
> On hold until API type id is standardised.

After https://github.com/commercetools/commercetools-docs-kit/pull/580 is merged, this PR can be reviewed.

PR #580, opened 2020-07-24
In ctp-urn util replace colon with hyphen, and lowercase letters

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/559


davifantasia:
I opened this PR because of a problem with element selection using Cypress, there is already a solution for that here:

https://github.com/commercetools/commercetools-docs-kit/issues/559#issuecomment-663527099

So this will be closed.

Issue #565, opened 2020-07-22
Design review: gatsby-theme-docs tasks that can be combined into 1 issue


davifantasia:
### Top menu dropdown

Should have on hover effect of blue text and bottom border:

![Screenshot 2020-07-22 at 10 48 43](https://user-images.githubusercontent.com/8238495/88155863-ef035880-cc08-11ea-9a4f-d6932de2e9e5.png)

### Index Menu

Use appropriate spacing, see design here - https://commercetools.invisionapp.com/share/GZY2OGRU2EK

### Release notes table

Header background is too bright, change from  from F2F2F2 to E6E6E6

PR #581, opened 2020-07-24
Design review: gatsby-theme-docs tasks that can be combined

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/565


packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -140,85 +143,110 @@ const SearchContainer = styled.div`
...
-              onClick={props.closeTopMenu}
+            <SwitcherButton
+              role="button"
+              aria-label={
+                props.isTopMenuOpen ? 'Close Top Menu' : 'Open Top Menu'
+              }
+              isActive={props.isTopMenuOpen}
+              onClick={props.toggleTopMenu}
+              onMouseOver={handleOnMouseOverTopMenu}
+              onMouseLeave={handleOnMouseLeaveTopMenu}
```
emmenko, line 182: Why do we need mouse events? We should be able to do this with CSS only.

davifantasia: I don't have a solution for css only. Do you have any suggestion?

emmenko: I can take a look later but I would prefer to not use event listeners if not necessary.

emmenko: I pushed a commit with the changes: https://github.com/commercetools/commercetools-docs-kit/pull/581/commits/af399b88b5a17790f82360aec1bcb206258817f3

I also took the chance to improve the a11y of the button element (see gif, where I use tab navigation to focus on the element and then toggle the menu using the spacebar and the enter keys).

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -116,17 +120,21 @@ const DocumentationSwitcherContainer = styled.div`
...
+  :hover,
+  :focus {
+    border-bottom: 2px solid ${designSystem.colors.light.linkNavigation};
+    color: ${designSystem.colors.light.linkNavigation};
+    svg {
+      * {
+        fill: ${designSystem.colors.light.linkNavigation};
+      }
+    }
+  }
```
emmenko, line 132: This is the CSS part to handle hover and focus effects.

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -116,17 +120,21 @@ const DocumentationSwitcherContainer = styled.div`
...
+
   @media screen and (${designSystem.dimensions.viewports.desktop}) {
     margin: 0 0 0 ${designSystem.dimensions.spacings.xl};
   }
 `;
-const SwitcherButton = styled.a`
```
emmenko, line 123: This is not necessary anymore, I "merged" it into the `DocumentationSwitcherContainer`, which is now `DocumentationSwitcherButton`

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -140,31 +148,48 @@ const SearchContainer = styled.div`
...
+      event.key === 'Enter' ||
+      event.key === ' ' ||
+      event.key === 'Spacebar' ||
+      event.which === 13 ||
+      event.which === 32;
+    if (enterOrSpace) {
+      event.preventDefault();
+      props.toggleTopMenu(event);
+    }
+  };
```
emmenko, line 163: To handle the spacebar and enter keys for a11y.

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -140,31 +148,48 @@ const SearchContainer = styled.div`
...
+                  display: none;
+                }
+              `}
+            />
+            <LogoButton />
+          </LogoContainer>
+          <DocumentationSwitcherButton
             role="button"
+            id="top-menu-switcher"
+            tabIndex="1"
```
emmenko, line 185: To make the element accessible

emmenko:
@davifantasia can you check the API e2e tests?

Issue #568, opened 2020-07-22
Design review: back to top functionality


davifantasia:
The page jumps to the page title instead of the very top which includes the top menu

emmenko:
Why? The link refers to the page heading, not the top bar.
From a user perspective, no one knows that this is a "back to top" link as intended by the design. It's not clear and thus the expectations are obviously different.

A "back to top" as you want it needs to be explicitly mention in the text link, otherwise it's not clear at all what it should do.

PR #586, opened 2020-07-28
back to top now linked to header instead of content

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/568


emmenko:
Is this the expected behavior now? Isn't this confusing to the user?

@nkuehn has this been decided with Sven?

nkuehn:
> @nkuehn has this been decided with Sven?

As previously said in other places: Yes.  He has explicitly put it into his design review results document - it was unspecified before, so the implementation was not a UX input but an assumption.

PR #594, opened 2020-08-04
API Request / Response Examples

### LINKED ISSUE #445: API docs: Include Request Examples for Endpoints
RAML allows "example" and "examples" properties for query parameters, POST payloads etc ,too but we have not yet defined the usage pattern in example files.  

Some constraints need to be clarified, e.g. how to connect query, path and request body examples that could have differing names or some be a single example and others multiple.   

Also, the examples would have to be converted into e.g. a single cURL call example.

### LINKED ISSUE #550: API Endpoint:  Request payload with example broken in wide layout
Situation:
 * API endpoint 
 * request body not placed somewhere else on the site but rendered directly in the endpoint layout
 * request body type has an example
 * wide layout

Current:  request body type and its example are rendered side-by-side but don't have the space, so the type is compressed to a thin column
 
Intended: request body type and its example can take the full width.  I.e. the complete API endpoint needs to allow the width and constrain its children to 770px except when they are side-by-side.   It might require using a different stack implementation, let's see.

nkuehn:
resolves #445 
resolves #550 

 * updates rmf-codegen to get curl endpoint examples
 * refactors some of the DOM structure and the sideBySide implementation
 * don't separately show request body type example in endpoints, it's part of the endpoint curl example
 * moves the endpoint description into the colored block (not fully defined in design, it's a proposal)
 * some spacings tuning

Blockers / TODO:
 * add changeset (multiple packages involved!)
 * curl generated by rmf-codegen not polished enough yet
    * space between -X and the method would make it better readable
    * missing ending quotation marks after token header
    * replace "EOF" token with "DATA" (can be chosen freely)
    * add hostname including https://
    * test that it works
    * consider rendering actual example values  in path and query variables to cURL if RAML has them set.  Could be confusing, too and misinterpreted as a fixed value. 
    
Steps to final design:
 1. merge, release and apply this to the documentation (only used in unreleased docs until now)
 2. get design review on the unreleased preview docs (agreed this way with sven)
 3. iterate here in new PR and branch given design feedback


nkuehn:
@davifantasia can you take this over when you're back please?  @jenschude  asked us to directly submit a PR to the template code in rmf-codegen to get the curl calls right.  Once that is released, update the version here again.  

The whole chain to putting it into the real preview docs, then getting design review etc. would take too long if it waits for me. 

Plus: consider what of this requires test coverage beyond introducing VRT. 

Thank you!

packages/gatsby-theme-docs/src/components/side-by-side.js
```diff
@@ -3,38 +3,28 @@ import PropTypes from 'prop-types';
...
  * Allows the first and second child of this container to float side by side in large desktop sizes.
- *
- * NOTE about the `!important` override.
- * The <SideBySide> content is sometimes an entry in the UI Kit stack (e.g. in API docs).
- * The UI kit stack forces the top spacing and uses `!important`.
- * However, given that the SideBySide is responsive, the second item can sometimes be
- * to the right of the first and sometimes under it.
- * Therefore, we need different spacing "logic" here - the UI Kit stacks and inlines are
- * not intended for responsive content. As a consequence, without this override, the right
- * item would not be top aligned with the left item.
```
davifantasia, line 18: @emmenko Nikolaus also removed it here.

davifantasia:
@jenschude 

Could you please take a look at these changes Nikolaus suggested in the description of this PR?

* curl generated by rmf-codegen not polished enough yet
- space between -X and the method would make it better readable
- missing ending quotation marks after token header
- replace "EOF" token with "DATA" (can be chosen freely)
- add hostname including https://
- test that it works
- consider rendering actual example values in path and query variables to cURL if RAML has them set. Could be confusing, too and misinterpreted as a fixed value.

jenschude:
>       * replace "EOF" token with "DATA" (can be chosen freely)
>       * add hostname including https://

already done

jenschude:
>       * missing ending quotation marks after token header

fixed

>       * consider rendering actual example values in path and query variables to cURL if RAML has them set. Could be confusing, too and misinterpreted as a fixed value.

I wouldn't do this atm as it would be harder to identify placeholder values. What could be done is to replace them with shell variables like:

`https://api.europe-west1.gcp.commercetools.com/${PROJECT_KEY}/categories/key=${KEY}`

Btw i also added required query parameters to the examples

nkuehn:
Thank you!

> What could be done is to replace them with shell variables like:

pro:
 * maybe / hopefully syntax highlighting picks them up
 * for basics like the token and project ID actually useful to be able to actually copy / paste and run some examples  (not as useful for resource specifics like 

con:
 * confusing / "yet another syntax" in the same screen 
    * we're not using a dollar in the documentation next to it at the moment: 
    * different naming convention (dromedar case vs. uppercase)

if we do it, the token and other header values should be done that way, too. 

**NEW DOCS example**:

image 

**OLD DOCS example**:

image

PR #615, opened 2020-08-19
Render endpoint code examples

davifantasia:


packages/gatsby-theme-api-docs/src/components/resource/method/request-response-examples.js
```diff
@@ -0,0 +1,81 @@
...
+  requestCodeExamples: PropTypes.arrayOf(
+    PropTypes.shape({
+      language: PropTypes.string.isRequired,
+      value: PropTypes.string.isRequired,
+    }).isRequired
+  ),
+  responses: PropTypes.arrayOf(
+    PropTypes.shape({
+      code: PropTypes.number.isRequired,
+      body: PropTypes.object,
```
emmenko, line 76: Can we be more specific about the object shape?

emmenko (commented):
Looks good as far as I can tell 👌

packages/gatsby-theme-api-docs/src/components/resource/method/request-response-examples.js
```diff
@@ -0,0 +1,89 @@
...
+      code: PropTypes.number.isRequired,
+      body: PropTypes.oneOfType([
+        PropTypes.node,
+        PropTypes.shape({
+          applicationjson: PropTypes.shape({
+            builtinType: PropTypes.string.isRequired,
+            type: PropTypes.string.isRequired,
+          }),
+        }),
+      ]),
```
davifantasia, line 84: A response body could be a json object or scalar.

Issue #600, opened 2020-08-07
Bug: Visible focus state border on the top menu


nkuehn:
![image](https://user-images.githubusercontent.com/4006139/89635747-c71e2100-d8a7-11ea-878c-f7b4601cab35.png)

The top menu has a visible focus state since it got the hover effect.  That's not intended.

PR #621, opened 2020-08-21
fix: focus styling active only when top menu is open

davifantasia:
Fixes https://github.com/commercetools/commercetools-docs-kit/issues/600

### DIFF
--- a/packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
+++ b/packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
@@ -122,6 +122,7 @@ const DocumentationSwitcherButton = styled.div`
 
   :hover,
   :focus {
+    outline: none;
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -161,6 +162,12 @@ const LayoutHeader = (props) => {
       props.toggleTopMenu(event);
     }
   };
+
+  const handleOnMouseLeave = (e) => {
+    if (!props.isTopMenuOpen) {
+      e.target.blur();
+    }
+  };
```
emmenko, line 170: I'm not sure we need this. Isn't the `outline: none;` enough?

davifantasia: > I'm not sure we need this. Isn't the `outline: none;` enough?

We still need the focused state when the menu is open.

Issue #567, opened 2020-07-22
Design review: index menu item not highlighted consistently on scroll when content menu is used.


davifantasia:
When we have content menu (see https://new-commercetools-docs.vercel.app/api/projects/categories#representations), scrolling through content that follows removes highlighted menu on the index menu.

emmenko:
As far as I can tell, this is not possible to fix with the current implementation.
The problem is that the (sub) sections are flat, meaning they are sibling elements. And the anchor links are also flat. There is no hierarchical notion of parent-child.

PR #630, opened 2020-08-28
fix: index menu inconsistent highlighting when content menu is used

davifantasia:
Fixes https://github.com/commercetools/commercetools-docs-kit/issues/567


cypress/integration/docs-smoke-test/content-menu.js
```diff
@@ -0,0 +1,26 @@
+import { URL_DOCS_SMOKE_TEST } from '../../support/urls';
+
+describe('Content menu', () => {
```
nkuehn, line 3: The component is not called "content menu".  The term isn't used anywhere yet.   I suggest to use the specific name of the component tested - also for the test file name.

packages/gatsby-theme-docs/src/layouts/internals/page-navigation.js
```diff
@@ -35,10 +35,17 @@ function flatten(items) {
...
+  let itemAndChildUrls;
+
+  if (item.items && recursive) {
+    const flattenedItems = flatten(item.items);
+    const flattenedItemUrls = flattenedItems.map((flatChild) => flatChild.url);
+    itemAndChildUrls = [item.url, ...flattenedItemUrls];
+  } else {
+    itemAndChildUrls = [item.url];
+  }
+
```
nkuehn, line 48: just to be sure: beyond rewriting it in a less compressed way, does this represent a change?  I can't see one.

packages/gatsby-theme-docs/src/layouts/internals/page-navigation.js
```diff
@@ -182,15 +189,37 @@ const Container = (props) => (
...
+            !item.items[0].items[0].title &&
+            item.items[0].items[0].items
+        );
+      }
+
+      const isActive = getIsActive(
+        props.activeSection,
+        item,
+        skipsOneLevelOfNav
+      );
```
nkuehn, line 213: Note to self: This is the actual functional change.

websites/docs-smoke-test/src/content/views/nested-headings.mdx
```diff
@@ -109,3 +109,41 @@ Content.
...
+
+# Content Menu With First Level Heading
+
+<ChildSectionsNav parent="content-menu-with-first-level-heading"/>
+
+#### Lorem
+
+Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur scelerisque metus nec nibh faucibus, sed faucibus ipsum rutrum. Nullam nibh lectus, eleifend a nulla sit amet, pretium hendrerit purus. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Suspendisse sit amet eros ut felis iaculis pharetra at condimentum odio. Duis gravida orci non lacus dignissim, in laoreet ex ultrices. Cras consectetur dolor risus, nec porttitor velit vehicula volutpat. Pellentesque sit amet finibus nulla, ut egestas lacus.
+
+Cras pulvinar est eget nibh vestibulum, id ultrices orci auctor. Nullam auctor imperdiet ligula, ac auctor ligula aliquet sed. Suspendisse potenti. Vivamus tristique suscipit arcu, eu auctor libero venenatis sed. Maecenas ut cursus justo. Fusce euismod augue vitae pellentesque porttitor. Proin consectetur libero tempus nisi interdum auctor.
```
nkuehn, line 121: image

This scenario is invalid content and does not need to be "tested" (it doesn't actually test something because the index nav isn't shown anyways in that scenario).   Skipping a heading level is only allowed level two downwards at the moment.

nkuehn (changes requested):
Functionality seems correct now, thanks for digging in. 

Two formal change requests though (invalid test content and a test that invents a new name for an existing component).

packages/gatsby-theme-docs/src/layouts/internals/page-navigation.js
davifantasia: No functionality change, this is just rewriting to make it easily readable.

nkuehn (approved):

PR #654, opened 2020-09-18
refactor: spacing in code example title bar

### LINKED ISSUE #528: Code Examples: Text in title bar gets only 50% of space
For example in the API docs example page `/types/examples-test-type` there is an example with a longer title.  That name wraps in the header line although there is plenty space available. 

The reason is that the header is divided 50/50 between the language selector and title. That is not ideal and the browser has the layout capacities to distribute the width as the content needs it.

TimonRey:
closes #528


nkuehn (approved):

nkuehn:
pls add a changeset and merge

PR #663, opened 2020-09-23
feat(docs-theme): render images in fragments

### LINKED ISSUE #543: Images in Release Notes not embedded in frame like in content
Images in Release Notes should look like images in regular content pages.   They currently don't have the round box around the image.   It's visible in some Merchant Center Release notes. 

It might require some refactoring or a second implementation of the image CSS because in content pages the images are rendered by a remark images plugin.  Alternatively, maybe employ that plugin too when rendering MDX in release notes?

### LINKED ISSUE #596: Optimize gatsby-remark-images configuration (build speed and quality)
https://www.gatsbyjs.org/packages/gatsby-remark-images/  is not yet configured optimally in the docs theme and we are seeing suboptimal screenshot qualities in practice as well as slow build times due to the number of image variants generated. 

currently the only setting is a maxWidth of 770px, with the other resolutions being the default "0.25x, 0.5x, 1x, 1.5x, 2x, and 3x sizes" of the plugin.  Here's the effective output:

```html
<img class="gatsby-resp-image-image" alt="product panel" title="product panel" 
src="/merchant-center/static/87bf57dd14b1a52a788848164c3c46eb/f4b77/product_panel.png" 
srcset="
/merchant-center/static/87bf57dd14b1a52a788848164c3c46eb/6d2d1/product_panel.png 193w,
/merchant-center/static/87bf57dd14b1a52a788848164c3c46eb/409e6/product_panel.png 385w,
/merchant-center/static/87bf57dd14b1a52a788848164c3c46eb/f4b77/product_panel.png 770w,
/merchant-center/static/87bf57dd14b1a52a788848164c3c46eb/b7936/product_panel.png 1155w,
/merchant-center/static/87bf57dd14b1a52a788848164c3c46eb/2b608/product_panel.png 1540w,
/merchant-center/static/87bf57dd14b1a52a788848164c3c46eb/7f7f8/product_panel.png 2396w" 
sizes="(max-width: 770px) 100vw, 770px" 
loading="lazy" 
style="width: 100%; height: 100%; margin: 0px; vertical-align: middle; position: absolute; top: 0px; left: 0px; opacity: 1; transition: opacity 0.5s ease 0s; color: inherit; box-shadow: white 0px 0px 0px 400px inset;">
```

Issues to be addressed:
 * 770px is not actually the max width of the image - images have a grey border and are in fact smaller - screenshots on low res monitors are sensible to such "just a few pixels" deviations and show artifacts)
 * The six other default resolutions aren't actually all needed and not set optimally. 
 * default lazy loading is less important in a docs site than having to wait for an image.  I suggest to use "auto" for the lazy attribute - the browser is usually smarter than us. 

The actually available size is 754px.  The small end of the fluid content column results in an image width of 576px.  Below that there's only mobiles where the image space ranges between 312 and 366 CSS pixels (the latter is an iphone 8 plus for example)., but these are all 2x or 3x device pixel ratio screens, so they would actually need 624 to 732px or more.  

Since the "small end" of the content column size isn't even "hit" by users that don't proactively try to find that spot before the next responsive breakpoint hits, resizing for that size does not really help either since it won't be used. 

**TL:DR**: 
 * modern mobiles pull about the same image sizes as an office monitor, we don't actually need images for them
 * we need our known upper and lower bounds of the content columns plus 2x multipliers of these. 
 * we don't need strictly 3x multipliers - we take the original shots on retina macbook screens and they are 2x. Plus, no added value in the case of a screenshot (when opening the original on click you get full res anyways). The only 3x devices are mobiles, which are less wide and little used on our site)
 
suggested settings (feedback welcome):
 * `maxWidth:  754`
 * `loading: "auto"`
 * `srcSetBreakpoints: [ /* 754, */  1508 ]` (maxWidth is added automatically, put in a comment here for transparency. Original resolution is retained automatically, too )

Additional upside is that this reduces the number of generated images from six to two (or three including the original resolutio), hence reducing build times, too.

### LINKED ISSUE #660: "null render" images in the fragment renderer.
to quick fix the error symbols in the release note list, inject a `img` component that does nothing.  

This is to be able to go live at all.   Next step is to build  working image component (separate issue)

### LINKED ISSUE #661: Render images in markdown fragment renderer
The markdown fragement renderer needs a working image component that accepts the path convention used in other MDX pages. 

this graphqL query can provide data for a static query hook that helps providing the right srcSet etc for the rendering

```graphql
{
  allImageSharp {
    nodes {
      fixed{
        src
        srcSet
      }
      parent {
        ... on File {
          relativePath
          relativeDirectory
        }
      }
    }
  }
}

```

The visual design should have the same border, title positioning etc. as the one in full pages and have a figure / figcaption structure to properly render it.   To have only one implementation, we should try to use the class names used for styling the gatsby-remark-images  images (if possible - it might not work du to sizing logic etc)

```html
<span
  class="gatsby-resp-image-wrapper"
  style="
    position: relative;
    display: block;
    margin-left: auto;
    margin-right: auto;
    max-width: 770px;
  "
>
  <a
    class="gatsby-resp-image-link"
    href="/static/284be2a1809043196635c48b745e0002/5b503/image.png"
    style="display: block"
    target="_blank"
    rel="noopener"
    data-link-type="image-link"
  >
    <span
      class="gatsby-resp-image-background-image"
      style="
        padding-bottom: 41.9689%;
        position: relative;
        bottom: 0px;
        left: 0px;
        background-image: url('...');
        background-size: cover;
        display: block;
        transition: opacity 0.5s ease 0.5s;
        opacity: 0;
      "
    ></span>
    <img
      class="gatsby-resp-image-image"
      alt="Delete button"
      title="Delete button"
      src="/static/284be2a1809043196635c48b745e0002/f4b77/image.png"
      srcset="
        /static/284be2a1809043196635c48b745e0002/6d2d1/image.png  193w,
        /static/284be2a1809043196635c48b745e0002/409e6/image.png  385w,
        /static/284be2a1809043196635c48b745e0002/f4b77/image.png  770w,
        /static/284be2a1809043196635c48b745e0002/b7936/image.png 1155w,
        /static/284be2a1809043196635c48b745e0002/2b608/image.png 1540w,
        /static/284be2a1809043196635c48b745e0002/5b503/image.png 1976w
      "
      sizes="(max-width: 770px) 100vw, 770px"
      style="
        width: 100%;
        height: 100%;
        margin: 0px;
        vertical-align: middle;
        position: absolute;
        top: 0px;
        left: 0px;
        opacity: 1;
        transition: opacity 0.5s ease 0s;
        color: inherit;
        box-shadow: white 0px 0px 0px 400px inset;
      "
      loading="lazy"
    />
  </a>
</span>

```

nkuehn:
* closes #596
 * closes #661
 * closes #543 
 * closes #660 (obsoletes it)

Notes:
 * re-using the existing classes only as far as it makes sense
 * chose to implement the width behavior as intended in the design (background always wide in small images) 
 * chose to directly address the sizing and scaling because it had to be defined anyways in the image data hook


davifantasia (approved):
Looks good to me.

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -141,7 +141,9 @@ module.exports = (themeOptions = {}) => {
                 // It's important to specify the maxWidth (in pixels) of
                 // the content container as this plugin uses this as the
                 // base for generating different widths of each image.
-                maxWidth: 770,
+                maxWidth: 754,
+                loading: 'auto',
+                srcSetBreakpoints: [754, 1508],
```
emmenko, line 146: Let's add a comment/reference that these values are also used in the static query `use-resized-images.js`, in case we need to change the values.

packages/gatsby-theme-docs/src/components/image.js
```diff
@@ -0,0 +1,41 @@
...
+  display: block;
+  background-color: #fff;
+  margin: auto;
+`;
+
+const Image = (props) => {
+  const resized = useResizedImagesByPath(props.src);
+  return resized ? (
+    <figure className="gatsby-resp-image-figure">
+      <a href={resized.originalImg} target="ct-docs-original-size-image">
```
emmenko, line 16: What does the `ct-docs-original-size-image` target do?

packages/gatsby-theme-docs/src/hooks/use-resized-images.js
```diff
@@ -0,0 +1,44 @@
+import nodeJsPath from 'path';
```
emmenko, line 1: Let's not use a nodejs API  in a browser environment

packages/gatsby-theme-docs/src/hooks/use-resized-images.js
```diff
@@ -0,0 +1,44 @@
...
+    `
+  );
+  return queryResult.allImageSharp.nodes.map((node) => {
+    return {
+      ...node.fluid,
+      path: nodeJsPath.join(
+        '/images/',
+        node.parent.relativeDirectory,
+        node.parent.relativePath
+      ),
```
emmenko, line 35:
```suggestion
      path: [
        '/images/',
        node.parent.relativeDirectory,
        node.parent.relativePath
      ].join('/'),
```

packages/gatsby-theme-docs/src/hooks/use-resized-images.js
```diff
@@ -0,0 +1,44 @@
...
+    };
+  });
+};
+
+export const useResizedImagesByPath = (path = '') => {
+  return useResizedImages().find((resizedImage) => {
+    return resizedImage.path === path;
+  });
```
emmenko, line 43:
```suggestion
  const allResizedImages = useResizedImages();
  return allResizedImages.find((resizedImage) => resizedImage.path === path);
```

packages/gatsby-theme-docs/src/hooks/use-resized-images.js
```diff
@@ -0,0 +1,44 @@
...
+      path: nodeJsPath.join(
+        '/images/',
+        node.parent.relativeDirectory,
+        node.parent.relativePath
+      ),
+    };
+  });
+};
+
+export const useResizedImagesByPath = (path = '') => {
```
emmenko, line 40: Why the empty string as default value?

```suggestion
export const useResizedImagesByPath = (path) => {
```

packages/gatsby-theme-docs/src/components/image.js
```diff
@@ -0,0 +1,41 @@
...
+        <Img
+          alt={props.alt}
+          title={props.title}
+          src={resized.src}
+          srcSet={resized.srcSet}
+          width={resized.presentationWidth}
+          height={resized.presentationHeight}
+        ></Img>
```
emmenko, line 24:
```suggestion
        />
```

emmenko (changes requested):
Thanks, I guess as a first step is good enough.

packages/gatsby-theme-docs/src/components/image.js
emmenko: Let's use `target="_blank"`

```suggestion
      <a href={resized.originalImg} target="_blank">
```

emmenko (approved):
Thanks!
VRTs are also fine, you can approve them.

PR #664, opened 2020-09-24
auto-generate src/images/releases directory

davifantasia:
For authors, all release notes images must go into the `src/images/releases` directory. So this change addresses that need.


emmenko:
Hmm why? Can you provide some context please?

davifantasia:
> Hmm why? Can you provide some context please?

I added some description. This came out of the discussion we had this morning in the docs team. Since we will be training users/contributors as part of the prelaunch process, lectors would point to this directory for contributors to use.

emmenko (approved):

PR #665, opened 2020-09-24
fix: add fixed date column width in RSS feed

### LINKED ISSUE #657: fixed date column width in RSS feed widget
When using multiple RSS feeds above each other it's visually improper to automatically size the columns of the table (shot below)

image

I suggest to hardwire the width of the left (date) column so that the dates just fit in and leave the rest of the space to the right column. The date is implemented to not break into multiple lines anyways.

TimonRey:
closes #657

### DIFF
--- a/.changeset/lovely-suits-judge.md
+++ b/.changeset/lovely-suits-judge.md
@@ -1,6 +1,5 @@
 ---
 '@commercetools-docs/ui-kit': patch
-'@commercetools-docs/writing-style': patch
 ---
 
 Set fixed width for the date column in the RSS Feed widget
--- a/.changeset/lovely-suits-judge.md
+++ b/.changeset/lovely-suits-judge.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-docs/ui-kit': patch
+'@commercetools-docs/writing-style': patch
+---
+
+Set fixed width for the date column in the RSS Feed widget
--- a/packages/ui-kit/src/components/rss-feeds.js
+++ b/packages/ui-kit/src/components/rss-feeds.js
@@ -87,7 +87,11 @@ const RssFeeds = (props) => {
           <tbody>
             {data.items.map((item, index) => (
               <tr key={`${item.title}${index}`}>
-                <td style={`width: ${dimensions.widths.rssFeedDateWidth}`}>
+                <td
+                  css={css`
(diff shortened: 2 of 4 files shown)


davifantasia (dismissed):
Why is this file included in the changes?

`packages/writing-style/vale-2.1.1`

Also, there is a problem with the actual page, but I'm not sure yet:

Screenshot 2020-09-24 at 13 44 47

emmenko:
The error is

```
The `style` prop expects a mapping from style properties to values, not a string. For example, style={{marginRight: spacing + 'em'}} when using JSX.
```

packages/ui-kit/src/components/rss-feeds.js
```diff
@@ -87,7 +87,7 @@ const RssFeeds = (props) => {
           <tbody>
             {data.items.map((item, index) => (
               <tr key={`${item.title}${index}`}>
-                <td>
+                <td style={`width: ${dimensions.widths.rssFeedDateWidth}`}>
```
emmenko, line 90:
```suggestion
                <td css={css`width: ${dimensions.widths.rssFeedDateWidth}`}>
```

emmenko (changes requested):
- use either the css prop or a style object (better the css prop)
- remove the binary file
- add a changeset

packages/writing-style/.gitignore
```diff
@@ -1,2 +1,3 @@
 *.zip
 *.tar.gz
+vale-*
```
emmenko, line 3: I'm not sure we need this. You just need to remove your binary file from disk, as it's moved to another location.

TimonRey: @nkuehn and I discussed it...

.changeset/lovely-suits-judge.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-docs/ui-kit': patch
+'@commercetools-docs/writing-style': patch
```
emmenko, line 3: Bumping the writing-style package is not necessary

packages/writing-style/.gitignore
emmenko: Hmm ok

emmenko (approved):
Thanks

PR #669, opened 2020-09-24
fix: load only images related to release notes

emmenko:
We have an issue where some images don't seem to be properly loaded.

```
Original error:
  pngload: non-recoverable state
  vips2png: unable to write to target target
```

For our use case in release notes list, we should only load images that are used in release notes. Since we have the convention of `/images/releases` we should load only images in that folder.


nkuehn (changes requested):
Hi, the images used in release notes do not yet live properly inside a subfolder and we can't guarantee that too.  

The subfolder is a strong convention to not completely lose track of what's where but can't be guaranteed (plus, authors may choose to use the same image that is also in the content page in the release note and that's actually OK to not duplicate.

emmenko:
I think we should make it a convention, otherwise we just load more data then necessary especially if there a lot of images.

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -80,6 +80,13 @@ module.exports = (themeOptions = {}) => {
         },
       },
       // Assets (e.g. images) used from the markdown pages
+      {
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'imagesReleases',
+          path: path.resolve(`./src/images/releases`),
```
nkuehn, line 87: I don't think we do us a performance favor by having two graphql File nodes for every images when "mounting" the same folder twice (once directly and once as part of `/images`).  

If we want to really fix this we would have to move the release notes images into a completely different conventional place, which I think it's too late for now.

emmenko: We can add the `ignore` option for the `/images` source location to ignore the `releases` folder.

I would try to be careful on loading only the necessary things, as we already do that in a lot of places and I feel like it's easy to get it out of hand if we don't look out for those things.

I mean, this is anyway a workaround and I hope we can find a better solution for the release notes list.

nkuehn: Ignoring the /images/releases folder sounds like a good thing to me, but I can't fully see the impact yet. 

I suggest to first check with the @commercetools/docs-lectors (tech writers) whether doing any sort of technical separation of images in release notes and other content is a good idea at all.   If it's intended that any image can be used anywhere then we're doing better in not hardwiring any assumptions here.

TimonRey:
Percy shows one less picture than before...is that intended?

TimonRey (commented):
Percy shows one less picture than before...is that intended?

nkuehn:
Hi, can we separate topics into PRs?  

We cannot merge and use the separation of release notes images into own folders before being done with the migration.  The auto migrated content from contenful cannot do that separation, only humans can.

emmenko:
Have you seen my message and the other PR?

nkuehn:
@emmenko  If I'm not confusing  things we can close this now because it was decided to not render images in the MC release notes at all. Right?

emmenko:
Yes we don't need this anymore.

Superseded by #702

PR #670, opened 2020-09-25
fix: release notes image aspect ratio and HTML structure

emmenko:
Same [fix as in the other PR](https://github.com/commercetools/commercetools-docs-kit/pull/669/commits/ea56a3c7f066198541d39ce75ab9675513833e36), but without the release notes image folder changes (in case we want to release this first):


packages/gatsby-theme-docs/src/utils/markdown-fragment-to-react.js
```diff
@@ -70,7 +70,7 @@ const markdownFragmentToReact = (markdownString, customElements) =>
         strong: Markdown.Strong,
         delete: Markdown.Delete,
         hr: Markdown.Hr,
-        img: Image,
+        img: ReleaseNotesImage,
```
davifantasia, line 73: If the image component must be renamed it should be `MarkdownFragmentToReactImage`.

davifantasia (changes requested):

packages/gatsby-theme-docs/src/utils/markdown-fragment-to-react.js
nkuehn: `MarkdownFragmentImage` would suffice there but I agree it's not a generic sort of image but specific to being in Markdown

nkuehn (commented):
To be honest I'm not a fan of adding this mess of a DOM here which is introducing a known bug and unintended layout (with small images) into more places while we are otherwise trying to get rid of it. 

If we can have small images display properly at least in a few places that's clearly better than having a bad behavior everywhere just for consistency.

emmenko:
> which is introducing a known bug

Are you referring to the `figure` element? It's not a bug if we render the HTML structure as we want.

> To be honest I'm not a fan of adding this mess of a DOM

I agree but the goal here is to (try) to have the same render behavior as with the `gastby-remark-images` plugin, I don't see why we should do it differently.

---

I'll take a look at the rendering though, since it does not seem to work as it should (see VRT).

nkuehn:
> but the goal here is to (try) to have the same render behavior as with the `gastby-remark-images` plugin

no, it's not.  gatsby-remark-images is not rendering as we want it to render and we just live with it because we came to the conclusion that we can't control it unless the plugin becomes more flexible.

emmenko:
That's a different thing. We shouldn't have different image rendering behavior in the website, even if we know it's not how we would like it to be.

emmenko:
Ok the fix now fixes also the "wrong" behavior of `gatsby-remark-images` coincidentally.

image

@nkuehn is that ok with you now?

emmenko:
image

packages/gatsby-theme-docs/src/components/markdown-fragment-image.js
```diff
@@ -0,0 +1,89 @@
...
+
+const MarkdownFragmentImage = (props) => {
+  const image = useResizedImagesByPath(props.src);
+  if (!image) {
+    return null;
+  }
+
+  // NOTE: the HTML structure below is taken from the `gatsby-remark-images` plugin,
+  // as we want to replicate the same behaviour and structure.
+  const ratio = `${(1 / image.aspectRatio) * 100}%`;
```
emmenko, line 14: This is taken from here: https://github.com/gatsbyjs/gatsby/blob/416dca7f67894dc34a48df4435b587ec3f2ac251/packages/gatsby-remark-images/src/index.js#L281

davifantasia (approved):
Looks good to me

PR #679, opened 2020-09-29
refactor(linkinator): exclude static assets from being checked

emmenko:
I don't think it makes sense to include in the link checker the static assets generated by Gatsby. For example:

```
  [SKP] http://localhost:5757/docs-smoke-test/component---packages-gatsby-theme-docs-src-templates-release-notes-list-js-906fbc4cbdfdc28f1849.js
  [SKP] http://localhost:5757/docs-smoke-test/page-data/releases/page-data.json
  [SKP] http://localhost:5757/docs-smoke-test/page-data/sq/d/590814825.json
  [SKP] http://localhost:5757/docs-smoke-test/static/587804a5968509430a052a1d8f11f257/47498/1200x300.jpg
  [SKP] http://localhost:5757/docs-smoke-test/static/29b5057b12f2e61e010113a49551d5e0/79056/transparent-bkg.png
  [SKP] http://localhost:5757/docs-smoke-test/static/03fedb3125cfc6d36d3de3a46a7434bf/f836f/200.jpg
  [SKP] http://localhost:5757/docs-smoke-test/static/587804a5968509430a052a1d8f11f257/5a944/1200x300.jpg
```

### DIFF
--- a/linkinator.config.json
+++ b/linkinator.config.json
@@ -2,6 +2,7 @@
   "recurse": true,
   "silent": true,
   "skip": [
+    ".(json|js|png|jpg|svg|css)(\\?(.*)+)?$",
     "https://docs.commercetools.com",
     "https://commercetools.com/",
     "https://support.commercetools.com",


nkuehn:
What is the problem you are trying to solve in the first place?  Are you getting false negatives or do you have a performance problem? 

Excluding all image files flat out is not a good approach to me.  It can actually happen if we have a bug or when an author takes an image location and uses it literally

emmenko:
The intent is to avoid checking the generated data from Gatsby, which is impossible for it to be broken.

> when an author takes an image location and uses it literally

We can include these checks I guess. How would the URL be like? Something like this?

```
http://localhost:5757/docs-smoke-test/images/my-image.jpg
```

nkuehn:
> impossible for it to be broken.

other peoples' software isn't bug free either.  I've seen enough shit to not drop a dead simple and fast test that covers tons of potential cases just because the intended case should typically work.

PR #681, opened 2020-09-29
a little less image resolution to transform at dev time

nkuehn:
This is reducing the number of transform by ca. a third in my tests 

(excluding the separate problem we introduced with the images in the react fragment renderers - these make up 50% at the moment because every image is processed twice).

### DIFF
--- a/.changeset/afraid-readers-notice.md
+++ b/.changeset/afraid-readers-notice.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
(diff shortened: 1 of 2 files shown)


emmenko:
> the separate problem we introduced with the images in the react fragment renderers - these make up 50% at the moment because every image is processed twice

Should we restrict the query to only load images in the `images/releases` folder, as proposed here #669?
Or is the plan to remove the images in the release notes list altogether?

emmenko (approved):

PR #687, opened 2020-09-30
fix: nav text outside the window in wide layout

### LINKED ISSUE #672: index nav text flowing outside window in wide layout
image

TimonRey:
closes #672


emmenko (changes requested):
The issue is not in the margin but rather in the width of the central content container.



The index nav has a min width of `225px`.

See here how it keeps the same width and it does not overflow. In the wide layout it should work the same way.

nkuehn:
@emmenko I don't think the fingerpointing is as simple.  The browser's 100% viewport does not include any scrollbars so it is the correct behavior when our grid gives the center column the automatic available width and the side columns have defined width.  

what @TimonRey did here is the same as what our CSS is already doing in smaller breakpoints, too - add a margin to the ToC itself to prevent it from flowing under the scroll bar.  16px is not enough though, most scrollbars are rather 24px.  The following screenshot is from the current live site and does the same thing (not sure where it is in the codebase though). It's only relevant in two two viewport size brackets that are really filling 100%.  In the rest we just don't realize. 

image

> The index nav has a min width of `225px`.

Side note: it has two sizes depending on breakpoint: 224 and 200. 
image

Side note 2: there's a layout glitch when using a non-touch input device - the index nav has a left grey border then.  Different topic for a different issue.

emmenko:
Yes we can of course add the margin, but my point was that the "overflow" issues is still not fixed.

Maybe we can split this into 2 tasks:
- add the marging
- investigate and fix the overflow issue with the responsive widths

nkuehn (approved):
Thanks! The combination works fine - I tested with a non-touch scrollbar, the extra margin is still needed. 

Please remove the duplicate changeset though, they have identical meaning in the text.

emmenko (approved):

emmenko:
@nkuehn can you review the VRTs and approve if they're fine for you?

emmenko (approved):

PR #692, opened 2020-10-01
endpoint design implementation

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/684


packages/gatsby-theme-api-docs/src/components/resource/method/url.js
```diff
@@ -47,27 +41,25 @@ const SlashWrappingPath = (props) => {
...
+        css={css`
+          background-color: ${methodColor};
+        `}
+      >
+        {method}
+      </Type>
+      <SpacingsStack scale="xs">
+        <BasePath>
+          <SlashWrappingPath>
+            {uris.baseUri.replace(/^(https:\/\/)/, '')}
```
nkuehn, line 55:
```suggestion
            {uris.baseUri}
```
that was omitted in the design, but one reason to give the hostname always its own line was to have the proper URL including protocol again

packages/gatsby-theme-api-docs/src/components/resource/method/index.js
```diff
@@ -1,5 +1,3 @@
 import Method from './method';
 
-const MethodFromIndex = Method;
-
-export default MethodFromIndex;
+export default Method;
```
emmenko, line 3: Usually you can simply write `export { default } from './method';`

emmenko (approved):
Code and VRTs look good from my side 👌

nkuehn (changes requested):
Compared to the visual design and found some differences:
 1. the text color of the hostname and the "OAuth2 Scopes", "Path Parameters" etc. headings is still the previous grey - the new design has them in darker to have enough contrast over the grey background, 
 2. the (red monospace) code blocks still have a background and border which is visible (esp. the border) if you look closely - and it add extra unneeded spacing.   When they are on the grey background they should have no background or border at all.
 3. It's not explicitly in the design, but I think the header row of the Response examples should have the darker grey defined under "White Table inside Endpoint" in the invision provided designs.  The two greys do not really contrast good enough.  (this might be annoying because it would be a third theme for the code examples next to dark and light - can be discussed).   I'll send sven the question.

davifantasia:
> 3. It's not explicitly in the design, but I think the header row of the Response examples should have the darker grey defined under "White Table inside Endpoint" in the invision provided designs.  The two greys do not really contrast good enough.  (this might be annoying because it would be a third theme for the code examples next to dark and light - can be discussed).   I'll send sven the question.

Created an issue for this here - https://github.com/commercetools/commercetools-docs-kit/issues/712

nkuehn (approved):
Good for now, the response example box header will get a different color in a separate step.

PR #701, opened 2020-10-08
refactor: add spacing between description and API type in wide layout

### LINKED ISSUE #697: visual polish in wide layout: description above API type has no spacing to type
There's not even minimal spacing below a description above an API type, let's add a little (no explicit situation in the designs, but it needs to behave as if the text would be in markdown above it instead of being the type description. 

Occurs only in wide layout! 

image

TimonRey:
closes #697


.changeset/gold-games-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Set spacing between description and API type in wide layout
```
nkuehn, line 5:
```suggestion
Do not remove top margin for any side by side content.  Resolves the API type having been too near to the description and manual side by side content being too near to the heading above.
``

**Review by nkuehn — APPROVED** [review]
@TimonRey  see my proposal for an improvement of the release text.  Please take care in not only describing the event that triggered the fix but also all other effects it has (was visible in the percy tests here that it also changes other situations)

File: .changeset/gold-games-admire.md (hunk 1 above)
Comment (TimonRey) [reply .changeset/gold-games-admire.md:5]: Makes sense...Thanks for the suggestion!  I will remember that 👍

File: .changeset/gold-games-admire.md [hunk 2]
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Do not remove top margin for any side by side content.  Resolves the API type having been too near to the description and manual side by side content being too near to the heading above.
```
nkuehn, line 5:
```suggestion
Do not remove top margin for any side by side content. Resolves the API type having been too near to the description and manual side by side content being too near to the heading above.
```

PR #705, opened 2020-10-09
fix(theme): render links for static files as static HTML links

### LINKED ISSUE #680: Links to static downloads "caught" by react and showing a 404.
For example in https://docs.commercetools.com/tutorials/extensions#validate-maximum-number-of-ten-items-in-the-cart  there are links to static downloads.   GatsbyJS uses the "static" folder next to "src" and the links actually work. 

But if the link is clicked inside the content generated by MDX it's caught by the `Link` component that tries to navigate to the location using react router (without loading a page). 

So we need a way to 
 - either make the link component understand that the links to the static stuff under "downloads" is not handled, i.e. handled as a regular link tag and not as a 
- or find a content pattern in MDX that instructs the link to open in a new tab or something else that leaves the react page context

emmenko:
Fixes #680 

> Note that serving static `.html` pages does not work in development mode. See https://github.com/gatsbyjs/gatsby/issues/13072

This PR settle on a convention that links referencing files, according to the following rules, are rendered as "static" HTML links.

- all files within the `/downloads` folder
- any HTML file (file path ending with `.html`)

Those files MUST be located in the `static` folder (https://www.gatsbyjs.com/docs/static-folder/).

> In the API docs, we already have the files located in the `static` folder.

### DIFF
--- a/packages/gatsby-theme-docs/README.md
+++ b/packages/gatsby-theme-docs/README.md
@@ -46,13 +46,15 @@ The project structure should contain at least the following files and folders:
 ├── .eslintrc.yml
 ├── gatsby-config.js
 ├── package.json
-└── src
-    ├── content
-    │   ├── files
-    │   └── index.mdx
-    ├── images
-    └── data
-        └── navigation.yaml
+├── src
+│   ├── content
+│   │   ├── files
+│   │   └── index.mdx
+│   ├── images
+│   └── data
+│       └── navigation.yaml
+└── static
+    └── downloads
 ```
 
 - `.eslintrc.yaml`: in case you're using a monorepository, you need to provide this file with an empty object `{}`, otherwise provide a valid ESLint configuration.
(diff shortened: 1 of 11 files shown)


packages/gatsby-theme-docs/README.md
```diff
@@ -149,6 +151,11 @@ These are required directories:
   - chapter-title: {} # another chapter, and so on...
   ```
 
+- `static`: this folder should contain files that do not need to be processed by Gatsby and will be served as-is. See [Gatsby static folder](https://www.gatsbyjs.com/docs/static-folder/).<br/>
+  Note that any `.html` file that is referenced as a link within the `*.mdx` content files is opened as a "static" HTML link, so when clicking on it the browser opens the link as a normal page.
+
+- `static/downloads`: this folder should contain static files that can be referenced in the links within the `*.mdx` content files. All links starting with `/downloads` will be rendered as "static" HTML links, so when clicking on it the browser opens the link as a normal page.
```
emmenko, line 157: This is the convention we want to settle for.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -26,6 +26,8 @@ exports.onPreBootstrap = (gatsbyApi, themeOptions) => {
     'src/content',
     'src/content/files',
     'src/releases',
+    'static',
+    'static/downloads',
```
emmenko, line 30: I think it makes sense to bootstrap those folders, so it's clear what the convention is

TimonRey (approved):
Looks good from my side 🙌 💯

nkuehn (approved):

PR #706, opened 2020-10-09
UI change: move beta flag from main nav to index nav

### LINKED ISSUE #700: UI change: Move beta flag from main (left) to index (right) nav
We have decided to implement moving the page-level beta flags from the left hand (main) nav to the right hand (index) nav's top area. 

While touching the top right corner anyways, the icon for "back to top" link is also changed to a more explicit upward pointing arrow. 

The data sourcing is not changing, the information is available on graphQL in the page metadata. 

Inspectable design prototype here ( closed account in commercetools):  https://commercetools.invisionapp.com/console/share/WP1MKLPW6A

Screenshot:
image

TimonRey:
closes #700


packages/gatsby-theme-docs/src/layouts/content-homepage.js
```diff
@@ -23,7 +23,6 @@ const LayoutContentHomepage = (props) => {
         {...layoutState.sidebar}
         {...layoutState.searchDialog}
         siteTitle={siteData.siteMetadata.title}
-        isGlobalBeta={props.pageData.isGlobalBeta}
```
emmenko, line 26: The global beta flag is a different thing and should stay.

image

Only the "chapter" page beta flags should be moved to the page index nav.

emmenko (changes requested):

packages/gatsby-theme-docs/src/layouts/content-homepage.js
TimonRey: Wasn't sure about that...thanks 👍

nkuehn: For clarification:  "Page" beta, not Chapter.

nkuehn (commented):
Please also update the smoke test page for "beta" to have 
 - significant content and a few chapters (can be all lorem ipsum), so that the index nav appears at all
 - a longer page name so you see how it behaves with line wrapping in the top right element of the index nav. 

Thanks !

nkuehn:
@TimonRey  please remind me via Slack when you think this is ready for review

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -234,7 +234,6 @@ const SidebarChapter = (props) => {
       <SpacingsStack scale="s">
         <LinkItem>
           <LinkTitle>{props.chapter.chapterTitle}</LinkTitle>
-          {props.chapter.beta && !props.isGlobalBeta && <BetaFlag />}
```
emmenko, line 237: I suppose we don't need the `beta` flag on the chapter level anymore. @nkuehn ?

We should also remove it from the graphql query then.

emmenko (commented):
Code looks good 👌

TimonRey:
@nkuehn I changed now the beta smoke test page to see the difference in the index nav. Looks good from my side so if you have time you can review it 👍

emmenko (approved):
UI looks nice!

nkuehn:
Thanks, it looks nice!  

But unfortunately the badge is so small that it's only hardly readable any more on a non-retina screen:

image

I'll check back with Sven, this is the same problem as in previous designs of the index nav - the font sizes were changed for a smaller font during implementation.

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
nkuehn: Correct, I forgot we have that feature at all. 
FYI @TimonRey

nkuehn (approved):

emmenko (changes requested):
You need to tell Percy to snapshot the beta page: https://github.com/commercetools/commercetools-docs-kit/blob/7f73056c2fc9ee1df717d2af08195c752f63994a/.percy.yml#L11

Also, is the **font size** of the beta flag ok like this? cc @nkuehn 

image

nkuehn:
> Also, is the **font size** of the beta flag ok like this?

I pushed a change here to set the font size of the flag to "small" instead of "extraSmall".  if that's still the case it's ok

emmenko:
I don't see any commit from you here... 🧐

emmenko (changes requested):
@davifantasia @TimonRey guys, the increased font size is still missing! Please fix that (which apparently was a commit from @nkuehn) before requesting a new review.
Thanks

nkuehn:
> which apparently was a commit from @nkuehn

Big apologies from me - I simply did not push it and moved on to other tasks.  Done now, I'll check percy etc one the build is through.

PR #731, opened 2020-10-27
chore: add directory in repository field for each package

emmenko:
To keep things consistent, and to point to the directory of the monorepo where the package lives


nkuehn (approved):
nice, I did not know this exists!

PR #737, opened 2020-10-30
fix(theme): load polyfill for IntersectionObserver API, when necessary

### LINKED ISSUE #733: (Sentry): guard usage of IntersectionObserver for stone age browsers (or ignore them)
https://sentry.io/organizations/commercetools/issues/?project=1819068&query=is%3Aunresolved&sort=user&statsPeriod=7d 

(multiple slightly varying entries that look like they are the identical root cause).

some code (likely in the index nav) is using the intersection observer browser API and there really are ie 11 and other stone age users on the site, so we have stuff in Sentry.   It should be easy to guard with a check whether the API is present to make Sentry not spammed with such stuff. 

@emmenko  alternatively we could try excluding IE 11 and such candidates from Sentry completely. It's not a development target but since it's a static site it's still usable even if a user has JS errors.  The Sentry JS has a "beforeSend" callback that can be used to filter out stuff (you can't do it server side).

emmenko:
Fixes #733

### DIFF
--- a/.changeset/perfect-drinks-rescue.md
+++ b/.changeset/perfect-drinks-rescue.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Load polyfill for `IntersectionObserver` API.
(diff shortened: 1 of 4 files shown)


davifantasia (approved):
Looks good to me.

Issue #689, opened 2020-09-30
Readme: Both Release Note links are broken


cneijenhuis:
I get a 404 if I click on either :(

cneijenhuis:
To clarify, I'm referring to this section: https://github.com/commercetools/commercetools-docs-kit/blob/master/packages/gatsby-theme-docs/README.md#writing-release-notes

PR #742, opened 2020-11-02
fix: broken links to release notes samples

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/689


.changeset/afraid-coins-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+"@commercetools-docs/gatsby-theme-docs": patch
```
emmenko, line 2: In theory we don't need a changeset here, nothing in the code changed that would need a release.

emmenko (approved):

PR #744, opened 2020-11-02
Make "chapter" legal.

TohruKohrita:
In the new documentation website, we have not only sections, subsection and pages, but also chapters. Thus, we (docs-lectors) would like to stop Vale complaining about the usage of this word, chapter.


nkuehn (approved):

nkuehn:
@TohruKohrita  I assumed this is not intended to be in draft mode and approved it because I was requested.

TohruKohrita:
> @TohruKohrita I assumed this is not intended to be in draft mode and approved it because I was requested.

Thank you! 
@adinakleine

PR #745, opened 2020-11-04
fix(docs): to explicitly pass emotion theme when possible

### LINKED ISSUE #682: investigate "Cannot read property 'light' of undefined"  errors in sentry
Among a few that rather look like outliers here's one issue that seems to occur with a certain regularity: 

https://sentry.io/organizations/commercetools/issues/1452889339/ 

"light" could relate to something that has two themes, maybe a code example block?

emmenko:
Fixes #682 

It appeared that under some unknown circumstances the `props.theme` accessed from within a styled component results in an empty object when building the website in production mode. To fix that, we explicitly pass wherever possible the `theme` object after reading it from the context with `useTheme`.

### DIFF
--- a/.changeset/neat-glasses-argue.md
+++ b/.changeset/neat-glasses-argue.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+'@commercetools-docs/ui-kit': patch
+---
+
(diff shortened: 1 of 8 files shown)


davifantasia (changes requested):
Title colour is not consistently right, please check the Percy snapshots.

emmenko:
Good that we have VRTs 🥳

@davifantasia please take another look, thanks

davifantasia (approved):
LGTM

PR #755, opened 2020-11-18
refactor: use fontsource instead of typeface package

emmenko:
The typeface package is deprecated

image


nkuehn (approved):
Looks good to me - thanks for the maintenance work. 

Side note:  It's not better or worse then the previous lib - Safari still has a short burst of unstyled typesetting when initially loading or reloading (other browsers just on initial load).

PR #758, opened 2020-11-24
refactor: generate icons using svgr cli

emmenko:
Generate SVG icon components using `@svgr/cli`, to avoid importing SVG files via Babel/Rollup/Webpack.


davifantasia (approved):
I don't see what the current advantage is by moving to svgr cli now but I see that it's more flexible and extensible than rollup,. So it has something beneficial for the future.

emmenko:
One thing is that it allows us to remove the webpack/rollup loaders, which essentially do the same thing.
And this is also required to be able to do #759

TimonRey (approved):
💯 🙌

PR #759, opened 2020-11-24
refactor: compile and bundle monorepo using preconstruct cli

emmenko:
Simplifies the tooling setup (for example no more Rollup, as it comes built into Preconstruct).

https://preconstruct.tools


davifantasia (approved):
LGTM

PR #760, opened 2020-11-24
Introduce new "rmf-codegen" package, replacing "ramldoc-generator"

nkuehn:
With the next steps being done in closer integrating the RAML modeling and documentation workflow it became clear that our existing "ramldoc-generator" package which downloads and provides rmf-codegen is too limited and cannot keep up with the development of rmf-codegen (beyond bumping the version). 

 * we need the `--watch` parameter to live update when the source RAML changed
 * we want to be able to use the full CLI and help output of rmf-codegen
 * we need the "validate" command in addition to the "generate" command
 * we want to already generate the postman collection during docs builds

This PR 
 1. removes the previous-previous generation legacy RAML package that was never uses in production (to not create even more confustion)
 1. adds a slimmer "rmf-codegen" package that only cares about providing the rmf-codegen JAR and the CLI command.  It directly connects the terminals instead of wrapping using shelljs to get whatever coloring and interactivity rmf-codegen itself provides.
 1. migrates related READMEs, dependencies and commands
 1. marks the "ramldoc-generator" package deprecated in the README (To be discussed - we could also choose to completely remove it and mark it deprecated in NPM)


davifantasia (changes requested):
Editing

davifantasia (approved):
LGTM

packages/ramldoc-generator/README.md
```diff
@@ -1,5 +1,7 @@
 # commercetools Docs Kit "RAMLdoc" Generator
 
+> **DEPRECATED** in favor of the new **@commercetools-docs/rmf-codegen** package which provides the complete functionality of rmf-codegen without a restrictive wrapper CLI.
+
```
emmenko, line 4: Suggestion: we can remove the rest of the README and document how to migrate to the new package.

packages/rmf-codegen/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# @commercetools-docs/rmf-codegen
```
emmenko, line 1: Remove this file, as it's autogenerated by changeset.

nkuehn:
Thanks for the review !

I've set up proper releases and better documentation of rmf-codegen with @jenschude  in the meanwhile - I will update this to base off the github releases instead of the intransparent Bintray ones, so let's not merge yet. 

https://github.com/commercetools/rmf-codegen/releases

.changeset/proud-carrots-press.md
```diff
@@ -0,0 +1,8 @@
+---
+'@commercetools-docs/gatsby-theme-api-docs': patch
+'@commercetools-docs/gatsby-transformer-raml': patch
+'@commercetools-docs/ramldoc-generator': patch
+'@commercetools-docs/rmf-codegen': major
```
emmenko, line 5:
```suggestion
'@commercetools-docs/rmf-codegen': minor
```

As otherwise the theme packages are also bumped as major. Unless you want to do that of course.

nkuehn: OK, looks like the better compromise.  I intended to have the new package start with a  [major].0.0  but it's just a formality so it'll start at 1.1.0 now.

PR #769, opened 2020-11-26
fix: regression in Firefox about autoprefixer for height: fit-content

emmenko:
See https://github.com/thysultan/stylis.js/issues/245


TimonRey (approved):
Looks good! 💯 
Good that someone of us uses firefox as well :D

davifantasia (approved):
LGTM

Issue #770, opened 2020-11-26
Backslash in mdx is rendered instead of being transformed to a new line


davifantasia:
Before:

https://commercetools-docs-git-change-renovate-config.commercetools.vercel.app/api/projects/shoppingLists#remove-lineitem

Curent:

https://docs.commercetools.com/api/projects/shoppingLists#remove-lineitem

davifantasia:
Apparently this was a problem solved back in April - https://github.com/commercetools/commercetools-docs-kit/pull/388

I tried to use the same solution but it didn't work so I'm investigating further.

Related issue - https://github.com/mdx-js/mdx/issues/421

PR #771, opened 2020-11-26
fix: include commonmark in gatsby-plugin-mdx options

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/770

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -124,6 +124,8 @@ module.exports = (themeOptions = {}) => {
             '.mdx',
             // ".md"
           ],
+          // implement commonmark for stricter compatibility, e.g. backslash transformed to newlines
+          commonmark: true,
           // List of remark plugins, that transform the markdown AST.
           remarkPlugins: [require('remark-emoji')],
           // List of rehype plugins, that transform the HTML AST.
--- a/.changeset/six-peaches-hope.md
+++ b/.changeset/six-peaches-hope.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-docs/gatsby-theme-docs": patch
+---
+
+fix: include commonmark in gatsby-plugin-mdx options, the ensures backslashes in mdx texts are transformed to new lines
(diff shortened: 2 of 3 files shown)


packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -124,6 +124,8 @@ module.exports = (themeOptions = {}) => {
             '.mdx',
             // ".md"
           ],
+          // implement commonmark for stricter compatibility, e.g. backslash transformed to newlines
+          commonmark: true,
```
emmenko, line 128: I'm surprised that this works to be honest, as this option is not documented.

https://github.com/gatsbyjs/gatsby/blob/c993d52bff1b2255ba8a2c0d04726bfdb71b070a/packages/gatsby-plugin-mdx/gatsby-node.js#L71-L126

emmenko: image

emmenko: I don't think this is the correct approach.

nkuehn: Outch, this is a regression of a regression we already had during migration.

Functionally, the fix here is correct, the behavior is one of the relevant differences between "tolerant" and commonmark MD parsing. 

The Gatsby devs have just recently introduced their option validation and their issue tracker shows quite some people finding options that are implemented but not part of their validation.   Their plugins often feed options through to sub-plugins but do not document all of them again - a practice that is annoying in the image related plugins, too. 

PS: The MDX / commonmark topic sounds eeerily similar to the remark upgrade to a pure commonmark parser that we discussed recently @emmenko

nkuehn (approved):
Functionally correct. Please file an issue at gatsbyjs that the options validation does not allow setting the option (or something is wrong with their dependency hell between MDX / remark parser versions that are be default commonmark or not)

packages/gatsby-theme-docs/gatsby-config.js
emmenko: Yeah, I tried to follow the trail a bit and it seems indeed that the plugin options are passed down to some other libraries, like `remark-parse`.

Alright then, let's keep it like this for now!

emmenko (approved):
Let's get this in, thanks for looking into it!

I think we should keep an eye on this though.

PR #783, opened 2020-12-02
fix: remove width overrides in right column

### LINKED ISSUE #777: Remove unncessary width overrides in the right hand column
The right hand column's with is defined in the grid to flow into 200px -> 224px (minmax function). 

But some content elements inside the column are setting the 224 px again or have own definitions via breakpoints. This leads to the nasty effects that we have with scrollbars etc. 

No content inside the grid column should redefine the width, they must take "auto" or just have no width defined.

TimonRey:
closes #777

### DIFF
--- a/.changeset/perfect-trainers-thank.md
+++ b/.changeset/perfect-trainers-thank.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+A bug related to the width of the right column was fixed.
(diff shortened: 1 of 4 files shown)


nkuehn (commented):
Great, very happy to see this old longstanding annoyance resolved. 

*Comment 1*:  In Safari, the spacing to the scrollbar is tight but acceptable to me. Our next spacings size would likely be too much. 

image

*Comment 2*: While testing I found an inconsistency in the widths - I think the right column would better be smaller here: 

image

Current CSS:
image

Proposal:  Switch right column size to 200px instead of 224 in this breakpoint to give more space to the actual content and be consistent with the next higher breakpoint (currently, when resizing the window, the right column is getting smaller, then bigger, then smaller, then bigger. 

(I know, this is beyond the original issue scope).

nkuehn (approved):
this is so nice now!  Much less wiggly when resizing and we found the root cause!

TimonRey:
🥳 🙌

Issue #643, opened 2020-09-10
Implement Filtering over ChildSectionNavigator  (phase 1)


nkuehn:
we have now received the final UX design for the ChildSectionNavigator component.  It defines the spacing more precisely and specifies a filter search box at the top of the component. 

This ticket covers implementing the basic functionality,  advanced / nice to have functionality is to be implemented in a separate ticket. 

Design link - https://commercetools.invisionapp.com/console/commercetools-API-DOCS---Component-V2-ckem9zrer39id01z90qsjc8iy/cketv6lbk02te01xk6zll8jwo/play

 * implement spacings and minimum height as specified.
 * filtering the contents as-you-type 
 * active / inactive state of the search box

Not immediately doing here:
 * display sentence beginnings like "set" "delete", "add a" etc.  in lighter font
 * fixing the box height when filtering (to be evaluated by UX first after basic behavior is implemented)
 * highlighting the matching substrings in color

davifantasia:
Design link - https://commercetools.invisionapp.com/console/commercetools-API-DOCS---Component-V2-ckem9zrer39id01z90qsjc8iy/cketv6lbk02te01xk6zll8jwo/play

nkuehn:
done already

PR #786, opened 2020-12-03
Feat: Implement filtering over child section nav

TimonRey:
is part of #643

### DIFF
--- a/.changeset/purple-kiwis-sit.md
+++ b/.changeset/purple-kiwis-sit.md
@@ -1,0 +1,7 @@
+---
+'@commercetools-docs/gatsby-theme-docs': major
+'@commercetools-docs/ui-kit': patch
+'@commercetools-website/api-docs-smoke-test': patch
+---
+
+A new filter for the child section navigator was implemented.
--- a/.changeset/purple-kiwis-sit.md
+++ b/.changeset/purple-kiwis-sit.md
@@ -1,5 +1,5 @@
 ---
-'@commercetools-docs/gatsby-theme-docs': major
+'@commercetools-docs/gatsby-theme-docs': minor
 '@commercetools-docs/ui-kit': patch
 '@commercetools-website/api-docs-smoke-test': patch
 ---
--- a/.changeset/purple-kiwis-sit.md
+++ b/.changeset/purple-kiwis-sit.md
@@ -1,7 +1,7 @@
 ---
 '@commercetools-docs/gatsby-theme-docs': minor
 '@commercetools-docs/ui-kit': patch
-'@commercetools-website/api-docs-smoke-test': patch
+'@commercetools-website/docs-smoke-test': patch
 ---
 
 A new filter for the child section navigator was implemented.
--- a/packages/gatsby-theme-docs/src/components/child-sections-nav.js
+++ b/packages/gatsby-theme-docs/src/components/child-sections-nav.js
@@ -1,33 +1,114 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
-import { designSystem, ContentNotifications } from '@commercetools-docs/ui-kit';
+import {
+  designSystem,
+  ContentNotifications,
+  createStyledIcon,
+} from '@commercetools-docs/ui-kit';
+import SecondaryIconButton from '@commercetools-uikit/secondary-icon-button';
 import SpacingsStack from '@commercetools-uikit/spacings-stack';
+import { CloseIcon } from '@commercetools-uikit/icons';
 import { usePageData } from '../hooks/use-page-data';
+import { SearchSvgIcon } from '../icons';
+
+const SearchIcon = createStyledIcon(SearchSvgIcon);
+
+const SearchBoxContainer = styled.div`
+  border: 1px solid ${designSystem.colors.light.borderSecondary};
+  border-bottom: none;
+  padding: ${designSystem.dimensions.spacings.m};
+  border-radius: ${designSystem.tokens.borderRadiusForChildSectionNav}
+    ${designSystem.tokens.borderRadiusForChildSectionNav} 0 0;
+`;
+
+const SearchInput = styled.input`
+  appearance: none;
+  background-color: ${designSystem.colors.light.surfacePrimary};
+  border: 1px solid ${designSystem.colors.light.borderInput};
+  border-radius: ${designSystem.tokens.borderRadiusForSearchInput};
+  box-shadow: none;
+  box-sizing: border-box;
+  color: ${designSystem.colors.light.textPrimary};
+  display: flex;
+  flex: 1;
+  font-family: inherit;
+  font-size: ${designSystem.typography.fontSizes.body};
+  height: ${designSystem.dimensions.heights.inputSearch};
+  min-height: ${designSystem.dimensions.heights.inputSearch};
+  outline: none;
+  overflow: hidden;
+  padding: 0
+    calc(
+      ${designSystem.dimensions.spacings.l} +
+        ${designSystem.dimensions.spacings.xs}
+    );
+  width: ${designSystem.dimensions.widths.searchBar};
+  &::placeholder {
+    color: ${designSystem.colors.light.textFaded};
+  }
+  &:active,
+  &:focus:not(:disabled) {
+    border-color: ${designSystem.colors.light.borderHighlight};
+    padding-right: ${designSystem.dimensions.spacings.xs};
+  }
+  &:disabled {
+    background-color: ${designSystem.colors.light
+      .surfaceForSearchInputWhenDisabled};
+  }
+`;
+
+const SearchInputIcon = styled.span`
+  position: absolute;
+  z-index: 1;
+  width: ${designSystem.dimensions.spacings.l};
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  padding-top: ${designSystem.dimensions.spacings.s};
+  padding-left: ${designSystem.dimensions.spacings.xs};
+  ${(props) => `${props.position}: ${designSystem.dimensions.spacings.xs};`}
+`;
+
+const SearchBox = styled.div`
+  position: relative;
+  width: ${designSystem.dimensions.widths.searchBar};
+`;
 
 const Container = styled.div`
+  min-height: calc(90px - ${designSystem.dimensions.spacings.m} * 2);
   border: 1px solid ${designSystem.colors.light.borderSecondary};
+  @media screen and (${designSystem.dimensions.viewports.mobile}) {
+    padding: ${designSystem.dimensions.spacings.s};
+  }
   padding: ${designSystem.dimensions.spacings.m};
-  border-radius: ${designSystem.tokens.borderRadiusForChildSectionNav};
+  border-radius: 0 0 ${designSystem.tokens.borderRadiusForChildSectionNav}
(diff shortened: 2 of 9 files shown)


websites/api-docs-smoke-test/src/content/layouts/type-large-desktop.mdx
```diff
@@ -2,14 +2,43 @@
...
+### Arnold Schwarzenegger
+
+### James Bond
+
+### Diego Maradona
+
+### SectionTitleWithLongText
+
+### SectionTitleWithSimilarText
+
```
nkuehn, line 41: The child section navigator component is provided by the base theme, not by the api docs theme so this test content should live in that smoke test website, too and not in the api docs smoke test.

websites/api-docs-smoke-test/src/content/layouts/type-regular-desktop.mdx
```diff
@@ -2,14 +2,42 @@
...
+### Captain America
+
+### Arnold Schwarzenegger
+
+### James Bond
+
+### Diego Maradona
+
+### SectionTitleWithLongText
+
```
nkuehn, line 39: same here, wrong smoke test site

packages/gatsby-theme-docs/src/components/child-sections-nav.js
```diff
@@ -71,27 +152,61 @@ SectionNavigation.propTypes = {
...
 const ChildSectionsNav = (props) => {
+  const [searchValue, setSearchValue] = useState('');
   const pageData = usePageData();
   const sectionToC = skipEmptyLevels(
     findCurrentSection(props.parent, pageData.tableOfContents)
   );
+  const filteredSectionToCList = sectionToC.items.filter(
+    (item) =>
+      !searchValue ||
+      item.title.toLowerCase().includes(searchValue.toLowerCase())
```
nkuehn, line 163: This line is the actual search behavior.  I would like to consider supporting multi word searches in a minimal way (i.e. not adding a full text library here).  E.g. when typing "cart    update"  matching the "UpdateCart" heading would be my expected behavior.  

`const words = searchValue.toLowerCase().split(/\W+/)`  (split into array by non-word characters)

Then, the lowercase heading title would have to match *all* of the words in any order (i.e. individual checks, many ways to code that available, e.g. a reduce function on the words array). 

An advanced feature would be to allow typos via allowing a certain levenshtein distance, but this would have to be an alternative match to the substring search we have here.  I would rather wait for feedback before going into actual NLP.

nkuehn (changes requested):
See code comments for code and behavior topics. 

UX feedback @Shecki :  I am not sure that the large typography with the high spacing between lines was intended.   I do remember the goal being to stay recognizably similar to the index nav because that is the same content with the same functionality in a different place (child section navigator is an alternative to the third heading level in the index nav).  The result does not indicate a notable similarity between index nav and child section navigator any more to me: 

image

I have not checked against the screens, they do not have a screen that has both the index nav and this element next to each other.   I do remember we reduced the font size for the index nav at some point but did not update the designs.

.changeset/purple-kiwis-sit.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-docs/gatsby-theme-docs': major
```
nkuehn, line 2: PS: why major?  I can't see a breaking change here.

.changeset/purple-kiwis-sit.md
```diff
@@ -0,0 +1,7 @@
...
+'@commercetools-docs/ui-kit': patch
+'@commercetools-website/api-docs-smoke-test': patch
```
nkuehn, line 4: No releases for smoke test sites please, they are not published to NPM

emmenko: That's ok, changeset includes all packages in the workspace.
Only packages marked as public will be released then.

.changeset/purple-kiwis-sit.md
emmenko: Agree, it should be minor

```suggestion
'@commercetools-docs/gatsby-theme-docs': minor
```

packages/gatsby-theme-docs/src/components/child-sections-nav.js
```diff
@@ -1,33 +1,114 @@
...
+
+const SearchBoxContainer = styled.div`
+  border: 1px solid ${designSystem.colors.light.borderSecondary};
+  border-bottom: none;
+  padding: ${designSystem.dimensions.spacings.m};
+  border-radius: ${designSystem.tokens.borderRadiusForChildSectionNav}
+    ${designSystem.tokens.borderRadiusForChildSectionNav} 0 0;
+`;
+
+const SearchInput = styled.input`
```
emmenko, line 25:
```suggestion
// TODO: this component is visually similar to the `SearchInput` component used
// for the docs search. As a follow up, we should look into extracting the common logic
// into one input component.
const SearchInput = styled.input`
```

packages/gatsby-theme-docs/src/components/child-sections-nav.js
emmenko: In the MC we use a library called `fuse.js` for advanced search capabilities.

https://fusejs.io

I would suggest to focus on this as a follow up, which it also helps with "highlighting" the matching search text.

packages/gatsby-theme-docs/src/components/child-sections-nav.js
```diff
@@ -71,27 +152,61 @@ SectionNavigation.propTypes = {
...
+    <div>
+      <SearchBoxContainer>
+        <SearchBox>
+          <SearchInputIcon position="left">
+            <SearchIcon size="medium" />
+          </SearchInputIcon>
+          {searchValue !== '' ? (
+            <SearchInputIcon position="right">
+              <SecondaryIconButton
+                label="Close search dialog"
```
emmenko, line 177:
```suggestion
                label="Reset input"
```

packages/ui-kit/src/design-system.js
```diff
@@ -51,6 +51,7 @@ export const colors = {
     borderForReleaseNoteTypeFix: '#b866ff',
     // Links
     link: customProperties.colorPrimary25,
+    linkChildSectionNavigationBackground: '#f2f2f2',
```
emmenko, line 54: Please use the correct design token / property

.changeset/purple-kiwis-sit.md
nkuehn: I remember we removed them in the past, but you're right that it does not matter in the end.

packages/gatsby-theme-docs/src/components/child-sections-nav.js
nkuehn: I don't think a "smart" NLP lib like fuse or the many others out there is the right behavior here since the primary use case are data type names which an NLP algorithm fails on anyways because ManyCamelCasedWords are just one word for them anyways unless we start throwing an additional custom decompounder at it that we fill with a custom dictionary of our terms (there are no good english decompounders because english does not use compound words).  

A rabbithole I would not want to couple the suggested simple but effective improvement to.

Shecki:
https://commercetools.invisionapp.com/console/commercetools-API-DOCS---Component-V2-ckem9zrer39id01z90qsjc8iy/cketv6lbk02te01xk6zll8jwo/play

@nkuehn the design is still on invision.
i need to check during this week, if everything is correct

nkuehn:
> @nkuehn the design is still on invision.
> i need to check during this week, if everything is correct

@Shecki  I know, but all the designs in invision have 16px / 14px font size for the "index nav" font styling but the implementation is 14px / 12px as agreed with you during the review process.    

**Here's the preview deeplink to see it together**: https://commercetools-docs-kit-git-filtering-over-childsectionnavigator.commercetools.now.sh/api-docs-smoke-test/layouts/type-regular-desktop

Even newer designs of the index nav use the "Old" bigger fonts but are implemented in 14/12px  - e.g. https://commercetools.invisionapp.com/console/commercetools-API-DOCS---Component-V2-ckem9zrer39id01z90qsjc8iy/ckfxpg9dz008901yu9wsvbc40/inspect 

Consequentially, I did the initial implementation of the ChildSectionNavigator using the font sizes and spacings that were chosen later during the review process, but this PR is following your design files  by the word and hence goes back to 16/14.

packages/ui-kit/src/design-system.js
TimonRey: Right, thanks 👍

.changeset/purple-kiwis-sit.md
TimonRey: So I changed now the theme to minor and kept the other two as before 👍

websites/api-docs-smoke-test/src/content/layouts/type-large-desktop.mdx
TimonRey: I moved it now into a new page in the smoke test under components called `Child section navigation`
90d6936

Shecki:
@nkuehn ok, sorry now it get your point.
and yes, correct it should be  14/12px.

i have to check all my designs and see where I still use the old sizes.


But I can´t find the update actions. within the link you shared

.changeset/purple-kiwis-sit.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-docs/gatsby-theme-docs': minor
+'@commercetools-docs/ui-kit': patch
+'@commercetools-website/api-docs-smoke-test': patch
```
emmenko, line 4:
```suggestion
'@commercetools-website/docs-smoke-test': patch
```

websites/docs-smoke-test/src/content/components/child-section-navigation.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+### Arnold Schwarzenegger
+
+### James Bond
+
+### Diego Maradona
+
+### SectionTitleWithLongText
+
+### SectionTitleWithSimilarText
```
emmenko, line 32: Let's use consistent naming here, not a mix of different things 😅

packages/gatsby-theme-docs/src/components/child-sections-nav.js
TimonRey: This is now the current search behavior...let me know what you think of it and if we covered every use case with it
694dcc1

packages/gatsby-theme-docs/src/components/child-sections-nav.js
```diff
@@ -1,33 +1,114 @@
...
+import { CloseIcon } from '@commercetools-uikit/icons';
 import { usePageData } from '../hooks/use-page-data';
+import { SearchSvgIcon } from '../icons';
+
+const SearchIcon = createStyledIcon(SearchSvgIcon);
+
+const SearchBoxContainer = styled.div`
+  border: 1px solid ${designSystem.colors.light.borderSecondary};
+  border-bottom: none;
+  padding: ${designSystem.dimensions.spacings.m};
```
nkuehn, line 20: Forgot the bottom spacing change: 

```suggestion
  padding: ${designSystem.dimensions.spacings.m} ${designSystem.dimensions.spacings.m} ${designSystem.dimensions.spacings.s} ${designSystem.dimensions.spacings.m};
```

nkuehn (changes requested):

emmenko (approved):
Code-wise looks good

nkuehn (approved):

nkuehn:
merged!  thanks a lot, what a really nice feature!

Issue #787, opened 2020-12-07
Query parameter "Array" type and it's description


davifantasia:
## Problem

* A query parameter that is of type array, is expected to be rendered as `Array of <type>`, instead it only renders `Array`
* Markdown included in description of query parameter is not parsed.

<img width="784" alt="Screenshot 2020-12-07 at 19 08 15" src="https://user-images.githubusercontent.com/8238495/101388009-a7335400-38bf-11eb-8c4c-d27b57c6fc36.png">

## Links

Change history API RAML spec: https://github.com/commercetools/commercetools-docs/pull/775
Change history API Swagger spec: https://history.europe-west1.gcp.commercetools.com/api-docs/

PR #788, opened 2020-12-08
feat: Query parameter "Array" type and it's description fix

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/787


websites/api-docs-smoke-test/source-raml/test/api.raml
```diff
@@ -72,29 +72,29 @@ uses:
...
           type: number
+          description: This `markdown` should be parsed correctly.
           required: false
-          example: 5
-        collection:
+          example: 50
+        objectParameter:
           type: objects.ObjectTestType
           required: false
+        arrayOfObject:
```
nkuehn, line 95: array of object is a very unrealistic, in fact technically impossible use case for a query parameter.  Since individual parameters can only be primitives,  sending the same HTTP query parameter multiple times can only be an array of String / Number / boolean,  And boolean does not really make sense to send multiple times, too. 

I suggest to rather test an Array of Stirngs.

nkuehn (approved):

PR #794, opened 2020-12-10
fix: code block tooltip to read from theme using hook

### LINKED ISSUE #790: Clipboard icon leads to "unable to load view" error
When hovering over the clipboard icon on any code snippet example, the page fails and shows an error "Unable to load view." 

https://docs.commercetools.com/api/authorization#examples

Screen Shot 2020-12-08 at 11 45 56 AM

Please fix. :-P

emmenko:
Fixes #790 (even though it's not reproducible in this repo).

TL;DR: recently we had seen issues related to styled components (emotion) not getting the theme context, even though the theme is correctly defined in the React context. As a workaround we manually read the theme using `useTheme`, and explicitly pass the `theme` as a prop.
This is a leftover of other related changes.

### DIFF
--- a/.changeset/tasty-crews-complain.md
+++ b/.changeset/tasty-crews-complain.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/ui-kit': patch
+---
+
+Fix code block tooltip to explicitly read from emotion theme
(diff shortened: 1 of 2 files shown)


nkuehn (approved):
Trust based approval since I cannot test it in a simple way.  

It's a bit discomforting that the type of issue is not observable in this repo but in the websites that use the theme.

PR #795, opened 2020-12-10
fix: do not restyle uikit icons

emmenko:


TimonRey (approved):
Thanks! 💯

emmenko:
The "baseline" alignment of the icon with the text does not really work, because the text can be wrapped into multiple lines.

image

We should keep it like this for now

image

---

A possible solution might be to have the vertical line of the icon shorter.

TimonRey:
I spoke to @Shecki and he will modify the icon for this use case. We can do that in another PR 👍

PR #800, opened 2020-12-11
fix: remove container with border in release-notes-subscribe-link component

### LINKED ISSUE #799: Fix release notes subscribe link component
We found a mistake on the release notes page in smoke-test where the release-notes-subscribe-link component still has an own border. After we modified the Layout of the page-header-side, this should ne covered by the layout and not the component itself.

Bildschirmfoto 2020-12-11 um 09 40 07

TimonRey:
closes #799


emmenko (commented):
Thanks. Can you add the changeset?

emmenko (approved):

PR #802, opened 2020-12-14
refactor: change datasource of rss feeds in smoke-test

### LINKED ISSUE #761: Don't refer to actual RSS feeds in the smoke test sites
Here's a regression that we had already fixed and that came back with the new multi-feed component:

The smoke test site is not using the local feeds of the smoke test itself any more but is referring to the current commercetools docs.  That leads to the visual regression tests randomly failing in some PR when commercetools releases something. 

Let's roll this back to the previous state that referenced a local RSS from the smoke test sites (will not work in dev, but in build)

TimonRey:
closes #761

### DIFF
--- a/websites/docs-smoke-test/src/content/components/rss-feeds.mdx
+++ b/websites/docs-smoke-test/src/content/components/rss-feeds.mdx
@@ -8,17 +8,6 @@ import { RssFeeds } from '@commercetools-docs/ui-kit'
 
 <Info>The RSS feed is only generated in production builds so this test fails in development mode.</Info>
 
-## Table with one feed
-
 <RssFeeds dataSources={[
-    "https://docs.commercetools.com/merchant-center/releases/feed.xml"
-]}/>
-
-## Table with multiple feeds
-
-<RssFeeds dataSources={[
-    "https://docs.commercetools.com/api/releases/feed.xml",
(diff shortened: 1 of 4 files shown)


websites/docs-smoke-test/src/content/components/rss-feeds.mdx
```diff
@@ -8,17 +8,6 @@ import { RssFeeds } from '@commercetools-docs/ui-kit'
...
-]}/>
-
-## Table with multiple feeds
-
 <RssFeeds dataSources={[
-    "https://docs.commercetools.com/api/releases/feed.xml",
-    "https://docs.commercetools.com/merchant-center/releases/feed.xml",
-    "https://docs.commercetools.com/custom-applications/releases/feed.xml",
-    "https://docs.commercetools.com/import-api/releases/feed.xml"
+    "/docs-smoke-test/releases/feed.xml"
```
nkuehn, line 12:
```suggestion
    "/docs-smoke-test/releases/feed.xml",
    "/api-docs-smoke-test/releases/feed.xml"
```
having just one feed in a test for multiple feeds is not really testing it.  If it's empty please create a dummy entry in the api docs site to test the aggregation logic.   The feeds are only available in full builds so in dev mode a single feed would not work too.

nkuehn (approved):

PR #815, opened 2021-01-07
fix: link check for smoke tests websites, add missing release list page for api smoke tests

emmenko:
I noticed that the link checker was only checking the template website, not the docs and api-docs smoke tests.

Apparently the skip rules for `/docs` and `/api` also include the smoke tests (because the path starts with the same value).

After fixing that, the link checker complained about the missing `releases.mdx` page for the api smoke tests.


emmenko:
I approved the snapshots, as all of them contain the new year 2021 in the footer. Also there is a new snapshot for the release notes overview page.

TimonRey (approved):
Looks good from my side 👍

PR #816, opened 2021-01-07
fix: patching lobotomized owl selector for emotion ssr

emmenko:


packages/gatsby-theme-docs/gatsby-ssr.js
```diff
@@ -32,23 +35,15 @@ export const replaceRenderer = ({
...
-    // I found the following modified Lobotomized Owl selector to work.
-    // It uses the "general sibling selector (~)" instead of the "adjacent sibling
-    // selector (+)" together with the `:not` psuedo-class to exclude the `style` elements.
-    // The "hack" is to replace all the normal Lobotomized Owl selector with
-    // the new one.
-    .replace(/> \* \+ \*/g, '> *:not(style) ~ *:not(style)');
-  replaceBodyHTMLString(patchedHtml);
+    // https://emmenko.medium.com/patching-lobotomized-owl-selector-for-emotion-ssr-5a582a3c424c
+    .replace(lobotomizedOwlSelectorRegex, patchedLobotomizedOwlSelector);
+  const patchedCss = css
```
emmenko, line 42: We're also patching the critical css styles now

packages/gatsby-theme-docs/gatsby-ssr.js
```diff
@@ -64,9 +59,9 @@ export const replaceRenderer = ({
     />,
     <style
       key="emotion-ssr"
-      data-emotion-css={ids.join(' ')}
+      data-emotion={`${docsCacheKey} ${ids.join(' ')}`}
```
emmenko, line 62: I saw this in the emotion repo examples/tests, so we're doing the same thing here.

packages/gatsby-theme-docs/gatsby-ssr.js
```diff
@@ -20,6 +20,9 @@ const iconLightDigest = createContentDigest(
   fs.readFileSync(require.resolve('./static/favicon-light-32x32.png'))
 );
 
+const patchedLobotomizedOwlSelector = '> *:not(style) ~ *:not(style)';
+const lobotomizedOwlSelectorRegex = />\s?\*\s?\+\s?\*/g;
```
nkuehn, line 24: since we're here to make this more robust the regex could also allow more than one whitespace-ish character. 
```suggestion
const lobotomizedOwlSelectorRegex = />\s*\*\s*\+\s*\*/g;
```

PR #821, opened 2021-01-11
Update Vale version, fix style rule and reformat YAML

nkuehn:
This updates vale and fixes a style rule that became incompatible with newer vale versions. 

Once this is merged we have a bit clearer image on why the vale GitHub action is currently failing in the docs content repo (rule out vale version differences and have an ini file that is known to work with latest vale). 

It also automates more YAML reformatting (also for .yml endings) to ease the update procedure for the standard Google styles.


lint-staged.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   '*.md': ['prettier --write --parser markdown'],
   '*.yaml': ['prettier --write --parser yaml'],
+  '*.yml': ['prettier --write --parser yaml'],
```
emmenko, line 4:
```suggestion
  '*.{yaml,yml}': ['prettier --write --parser yaml'],
```

packages/writing-style/package.json
```diff
@@ -20,8 +20,8 @@
   },
   "scripts": {
     "postinstall": "node scripts/download-vale.js",
-    "download:google": "node scripts/download-google-styles.js",
-    "download:writegood": "node scripts/download-writegood-styles.js"
+    "download:google": "node scripts/download-google-styles.js && prettier --write --parser yaml '**/*.{yaml,yml}'",
+    "download:writegood": "node scripts/download-writegood-styles.js && prettier --write --parser yaml '**/*.{yaml,yml}'"
```
emmenko, line 24: Hmm running prettier here is not necessary. The lint-staged hook should do that.

nkuehn: It turned out to be necessary to better be able to see what has really changed in the external styles so we can adjust our own overrides respectively before committing the whole update

emmenko: Sorry, I'm not sure I fully follow the point here 🤔

nkuehn: The script is pulling a large number of YAML files from an external repository to persist them as a copy in our repo (this is intended). Each line / object in them represents a certain writing style rule and some of them we override or modify in our own ruleset to adjust to our domain needs. 

When I update the external "base" rules I need a way to see what has really changed beyond formatting differences to be able to see which of our overrides I have to adjust / rename / add etc.  .   If I don't reformat the files It's all formatting noise and I don't see the actual differences.  If I reformat them I see the actual differences.    And I want to see that before committing because I don't want to commit an overall broken config.

emmenko: Thanks, I understand now. I'm fine to keep it then.

We can also avoid running prettier at all for those files, since they are downloaded anyway.

davifantasia (commented):
Looks good to me. It can be merged once the links test is fixed.

davifantasia (approved):

PR #827, opened 2021-01-13
chore: update vale again and fix changeset

nkuehn:
Followup PR to #821 , while we did this they released another new version that we'll better include right now. 

Also, this removes the unnecessary packages that were in the changeset so we can release.


davifantasia (approved):
LGTM

emmenko (approved):

PR #828, opened 2021-01-13
chore: enable gatsby dev flags to improve development performance

emmenko:
This PR proposes to enable the following Gatsby feature flags:

- `DEV_SSR: true`
- `FAST_REFRESH: true`
- `PRESERVE_WEBPACK_CACHE: true`

According to Gatsby, this should help improving the performance for local development.

Note that the `DEV_SSR` allows to ensure that a page is correctly rendered on the server (see some of the fixes in the PR as a result of that).

I suggest to checkout the branch and start the websites locally, to make sure that things work fine.

### DIFF
--- a/websites/api-docs-smoke-test/gatsby-config.js
+++ b/websites/api-docs-smoke-test/gatsby-config.js
@@ -5,6 +5,13 @@ const {
 const colorPresets = require('@commercetools-docs/gatsby-theme-docs/color-presets');
 
 module.exports = {
+  // https://www.gatsbyjs.com/docs/reference/release-notes/v2.28/#feature-flags-in-gatsby-configjs
+  // https://www.gatsbyjs.com/docs/reference/release-notes/v2.30
+  flags: {
+    DEV_SSR: true,
+    FAST_REFRESH: true,
+    PRESERVE_WEBPACK_CACHE: true,
+  },
   pathPrefix: '/api-docs-smoke-test',
   siteMetadata: {
     title: 'API Docs Smoke Test',
--- a/.changeset/twelve-shirts-help.md
+++ b/.changeset/twelve-shirts-help.md
@@ -1,0 +1,8 @@
+---
+"@commercetools-docs/gatsby-theme-docs": patch
+"@commercetools-website/api-docs-smoke-test": patch
+"@commercetools-website/docs-smoke-test": patch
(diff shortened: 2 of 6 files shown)


websites/api-docs-smoke-test/gatsby-config.js
```diff
@@ -5,6 +5,13 @@ const {
 const colorPresets = require('@commercetools-docs/gatsby-theme-docs/color-presets');
 
 module.exports = {
+  // https://www.gatsbyjs.com/docs/reference/release-notes/v2.28/#feature-flags-in-gatsby-configjs
+  // https://www.gatsbyjs.com/docs/reference/release-notes/v2.30
+  flags: {
+    DEV_SSR: true,
+    FAST_REFRESH: true,
+    PRESERVE_WEBPACK_CACHE: true,
```
davifantasia, line 13: Is this experimental flag (PRESERVE_WEBPACK_CACHE) intended for our production sites too?

emmenko: I'm not sure I understand the question 🤔

davifantasia: > I'm not sure I understand the question 🤔

From what I read on Gatsby's website, PRESERVE_WEBPACK_CACHE flag is experimental. So my questions is do we intend to use this flag on sites in our commercetools-docs repo?

Link - https://www.gatsbyjs.com/docs/reference/release-notes/v2.28/#experimental-new-cache-clearing-behaviors

emmenko: I think we should yes.

But still, the changes here are unrelated to that. You have to specify the flags for each website anyway.

TimonRey (approved):
I've tested it a couple of times, looks good 💯

emmenko:
@davifantasia anything else from your side?

davifantasia (approved):
LGTM

websites/api-docs-smoke-test/gatsby-config.js
davifantasia: > I think we should yes.

I would be wary of using anything experimental in production.

emmenko: I'm not sure where do you get this idea, but the flags are for development.

PR #831, opened 2021-01-18
chore(renovate): preserve semver ranges

emmenko:
* Moves renovate config into `.github` folder
* Uses `preserveSemverRanges` instead of `pinOnlyDevDependencies`

Just so that we keep things consistent with other repos.


TimonRey (approved):
Thanks :)

PR #832, opened 2021-01-19
feat: script for generating release note files

### LINKED ISSUE #723: Automate release note file generation with a script.
## Why

Creating a release note (looking for the right location, copying content from the template) can be cumbersome for authors and contributors. So this issue is to help fix the problem.

## Acceptance criteria

Users can create a release note by running a script inside the microsite folder, e.g.

```sh
yarn create-release-note
```

Ideally, a CLI would collect the date, type and title on the CLI interactively using a library like prompts ( https://www.npmjs.com/package/prompts  - there are many others but this is used by gatsby CLI so it's pulled anyways)

Interactive input:
 * "Type": (single select,  "enhancement" / "feature" / "fix" / "announcement")
 * "Release Date": (date, no time)
 * "Title": (String, single line, max 256 chars)
 * "Description in Feeds (title plus value proposition)": (String, single line, max 256 chars - can be prefilled with the title if that is easy to achieve)

File location:  `./src/releases/slugified-title.mdx`  (check what slugify lib we use anyways to avoid an additional dependency)

Script location:  Centrally provided as a bin by the docs theme package 

## Ouput template structure

(written in GitHub, indentation etc. may be wrong)

```js
fileContent = `
---
date: ${isoDate}
title: ${title}
description: | 
  ${description}
type: ${type}
topics:
# Please use the topics defined in the wiki and stay focused.
# https://wiki.commercetools.com/display/DOCS/Writing+release+notes#Writingreleasenotes-Typesandtopics
#  - Topic1
#  - Topic2
---

${description}

<!--more-->
```

TimonRey:
closes #723 

Note: pls check if the output has the correct format and if it matches the requirements. Thanks.

### DIFF
--- a/.changeset/slow-ants-tie.md
+++ b/.changeset/slow-ants-tie.md
@@ -6,5 +6,5 @@ A script for generating release note files was implemented.
 Run the following command in your console and follow the given instructions to create a new release note.
 
 ```
-yarn create-release-note
+yarn create-docs-release-note
 ```
--- a/.changeset/slow-ants-tie.md
+++ b/.changeset/slow-ants-tie.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+A script for generating release note files was implemented.
--- a/.changeset/slow-ants-tie.md
+++ b/.changeset/slow-ants-tie.md
@@ -3,3 +3,8 @@
 ---
 
 A script for generating release note files was implemented.
+Run the following command in your console and follow the given instructions to create a new release note.
+
+```
+yarn create-release-note
+```
--- a/packages/gatsby-theme-docs/bin/create-release-note
+++ b/packages/gatsby-theme-docs/bin/create-release-note
@@ -25,17 +25,18 @@ const questions = [
   {
     type: 'text',
     name: 'title',
-    message: 'Define a title',
+    message: 'The title of the release note. Max 256 character plain text.',
     validate: (title) =>
-      title.length > 265 ? 'Title is longer than 256 chars!' : true,
+      title.length > 256 ? 'Title is longer than 256 chars!' : true,
   },
   {
     type: 'text',
     name: 'description',
-    message: 'Write a description',
+    message:
+      'Key changes and their advantages. Max 256 character plain text. This text is tweet-able and used in RSS feeds.',
     format: (val, values) => `${values.title} ${val}`,
     validate: (description) =>
-      description.length > 265 ? 'Description is longer than 256 chars!' : true,
+      description.length > 256 ? 'Description is longer than 256 chars!' : true,
   },
 ];
 
--- a/packages/gatsby-theme-docs/bin/create-release-note
+++ b/packages/gatsby-theme-docs/bin/create-release-note
@@ -1,0 +1,70 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+const slugify = require('slugify');
+const prompts = require('prompts');
+
+const questions = [
+  {
+    type: 'multiselect',
+    name: 'type',
+    message: 'Pick the type of the release note',
+    choices: [
+      { title: 'enhancement', value: 'enhancement' },
+      { title: 'feature', value: 'feature' },
+      { title: 'fix', value: 'fix' },
+      { title: 'announcement', value: 'announcement' },
+    ],
+  },
+  {
+    type: 'date',
+    name: 'date',
+    message: 'Pick a release date',
+  },
+  {
+    type: 'text',
+    name: 'title',
+    message: 'Define a title',
+    validate: (title) =>
+      title.length > 265 ? 'Title is longer than 256 chars!' : true,
+  },
+  {
+    type: 'text',
+    name: 'description',
+    message: 'Write a description',
+    format: (val, values) => `${values.title} ${val}`,
+    validate: (description) =>
+      description.length > 265 ? 'Description is longer than 256 chars!' : true,
+  },
+];
+
+(async () => {
+  const response = await prompts(questions);
+
+  const template = `---
+date: ${response.date}
+title: ${response.title}
+description: |
+  ${response.description}
+type: ${response.type}
+topics:
+# Please use the topics defined in the wiki and stay focused.
+# https://wiki.commercetools.com/display/DOCS/Writing+release+notes#Writingreleasenotes-Typesandtopics
+#  - Topic1
+#  - Topic2
+---
+
+${response.description}
+
+<!--more-->
+`;
+
+  fs.writeFileSync(
+    path.join(
+      process.cwd(),
+      `src/releases/${slugify(response.title, '-')}.mdx`
+    ),
+    template
+  );
+})();
--- a/packages/gatsby-theme-docs/bin/create-release-note.js
+++ b/packages/gatsby-theme-docs/bin/create-release-note.js
@@ -11,10 +11,27 @@ const questions = [
     name: 'type',
     message: 'Pick the type of the release note',
     choices: [
-      { title: 'enhancement', value: 'enhancement' },
-      { title: 'feature', value: 'feature' },
-      { title: 'fix', value: 'fix' },
-      { title: 'announcement', value: 'announcement' },
+      {
+        title: 'enhancement',
+        value: 'enhancement',
+        description: 'for updates of existing features',
+      },
+      {
+        title: 'feature',
+        value: 'feature',
+        description: 'for the introduction of an entirely new feature',
+      },
+      {
+        title: 'fix',
+        value: 'fix',
+        description: 'for bug fixes',
+      },
+      {
+        title: 'announcement',
+        value: 'announcement',
+        description:
+          'for release notes with no product change nor immediate product change',
+      },
     ],
   },
   {
@@ -38,6 +55,13 @@ const questions = [
     validate: (description) =>
       description.length > 256 ? 'Description is longer than 256 chars!' : true,
   },
+  {
+    type: 'list',
+    name: 'topics',
+    message:
+      'Write down the topics for this release note. Seperate them with commas.',
+    separator: ',',
+  },
 ];
 
 (async () => {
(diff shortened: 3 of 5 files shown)


.changeset/slow-ants-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+A script for generating release note files was implemented.
```
emmenko, line 5: Please also document (add an example) how to use the script.

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env node
```
emmenko, line 1: Better to name this file `create-release-note.sh` (with the `.sh` extension).

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
...
+      { title: 'feature', value: 'feature' },
+      { title: 'fix', value: 'fix' },
+      { title: 'announcement', value: 'announcement' },
+    ],
+  },
+  {
+    type: 'date',
+    name: 'date',
+    message: 'Pick a release date',
+  },
```
emmenko, line 24: I think it would make sense to have a default value for the "today" date.

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
...
+const questions = [
+  {
+    type: 'multiselect',
+    name: 'type',
+    message: 'Pick the type of the release note',
+    choices: [
+      { title: 'enhancement', value: 'enhancement' },
+      { title: 'feature', value: 'feature' },
+      { title: 'fix', value: 'fix' },
+      { title: 'announcement', value: 'announcement' },
```
emmenko, line 17: It would be nice to add a short explanation of what/how these types are used for.

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
...
+  },
+  {
+    type: 'date',
+    name: 'date',
+    message: 'Pick a release date',
+  },
+  {
+    type: 'text',
+    name: 'title',
+    message: 'Define a title',
```
emmenko, line 28:
```suggestion
    message: 'The title of the release note. Max 256 character plain text.',
```

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
...
+    validate: (title) =>
+      title.length > 265 ? 'Title is longer than 256 chars!' : true,
```
emmenko, line 30:
```suggestion
      title.length > 256 ? 'Title is longer than 256 chars!' : true,
```

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
...
+  },
+  {
+    type: 'text',
+    name: 'description',
+    message: 'Write a description',
+    format: (val, values) => `${values.title} ${val}`,
+    validate: (description) =>
+      description.length > 265 ? 'Description is longer than 256 chars!' : true,
```
emmenko, line 38:
```suggestion
      description.length > 256 ? 'Description is longer than 256 chars!' : true,
```

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
...
+    type: 'text',
+    name: 'title',
+    message: 'Define a title',
+    validate: (title) =>
+      title.length > 265 ? 'Title is longer than 256 chars!' : true,
+  },
+  {
+    type: 'text',
+    name: 'description',
+    message: 'Write a description',
```
emmenko, line 35:
```suggestion
    message: 'Key changes and their advantages. Max 256 character plain text. This text is tweet-able and used in RSS feeds.',
```

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
...
+(async () => {
+  const response = await prompts(questions);
+
+  const template = `---
+date: ${response.date}
+title: ${response.title}
+description: |
+  ${response.description}
+type: ${response.type}
+topics:
```
emmenko, line 51: We should also add a prompt for the topics. The user can enter a comma separated list of values, which you then transform into a list.

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
...
+# Please use the topics defined in the wiki and stay focused.
+# https://wiki.commercetools.com/display/DOCS/Writing+release+notes#Writingreleasenotes-Typesandtopics
```
emmenko, line 53: I would try to avoid linking to internal pages. I think we can provide all the necessary information here.

packages/gatsby-theme-docs/bin/create-release-note
```diff
@@ -0,0 +1,70 @@
...
+#  - Topic1
+#  - Topic2
+---
+
+${response.description}
```
emmenko, line 58: We can maybe put a "lorem ipsum" text here.

emmenko (changes requested):
Thanks, I left some comments to improve the prompts infos.

I think it would also be nice to have an integration test.

packages/gatsby-theme-docs/bin/create-release-note
TimonRey: Do you mean something like a description where to find it in the wiki? 🤔

packages/gatsby-theme-docs/bin/create-release-note
TimonRey: 🤔 okay

emmenko: This is just a suggestion btw. The point is that the content of the release note has to be written anyway.

packages/gatsby-theme-docs/bin/create-release-note
emmenko: I was thinking as comments.

Maybe check with @nkuehn if he's fine having the private link here.

packages/gatsby-theme-docs/bin/create-release-note
nkuehn: Really?  To me, `.sh` is for bash script and not JS.  Editors would pick up the wrong syntax highlighting, too.  I would rather consider to keep it "js" and  define an extensionless binary name in package.json that will then be the actual command.

packages/gatsby-theme-docs/bin/create-release-note
nkuehn: there is a wiki page with the usage instructions in the "let's write the docs" section.

packages/gatsby-theme-docs/bin/create-release-note
nkuehn: Better than nothing I guess if the alternative is to have to start selecting even the year from scratch.

packages/gatsby-theme-docs/bin/create-release-note
nkuehn: +1

packages/gatsby-theme-docs/bin/create-release-note
nkuehn: Yes, let's avoid any links to our closed wiki etc.   
It's as it is and our people know where the "Let's write the docs" wiki is.

packages/gatsby-theme-docs/bin/create-release-note
emmenko: Sorry you're right, I meant to say to add the file extension (which is obviously `.js` in this case) 😅

packages/gatsby-theme-docs/bin/create-release-note
emmenko: Yes I meant to add this information in the prompt, so that people don't have to look it up from the wiki

packages/gatsby-theme-docs/bin/create-release-note
TimonRey: Ahh, okay thanks 😄

packages/gatsby-theme-docs/bin/create-release-note
TimonRey: So at the moment the prompt has the "todays" date as default. Don't need to add an initial value here for that 🤔

packages/gatsby-theme-docs/bin/create-release-note
TimonRey: I added the explanation texts from the wiki page to the descriptions of the types now

packages/gatsby-theme-docs/package.json
```diff
@@ -13,6 +13,9 @@
     "url": "https://github.com/commercetools/commercetools-docs-kit.git",
     "directory": "packages/gatsby-theme-docs"
   },
+  "bin": {
+    "create-release-note": "bin/create-release-note.js"
```
emmenko, line 17: Nit: maybe have a more specific script name

```suggestion
    "create-docs-release-note": "bin/create-release-note.js"
```

packages/gatsby-theme-docs/bin/create-release-note.js
```diff
@@ -0,0 +1,100 @@
...
+  ${response.description}
+type: ${response.type}
+topics:
+${response.topics
+  .map((topic) => {
+    return `- ${topic}
+`;
+  })
+  .toString()
+  .replace(/,/g, '')}
```
emmenko, line 83: I think this should work as well (but please double check):

```suggestion
${response.topics.map((topic) => `- ${topic}`).join('\n')}
```

emmenko (commented):
Thanks, looking better now! 👌

packages/gatsby-theme-docs/bin/create-release-note.js
TimonRey: Thanks! Yes this is much better 💯

packages/gatsby-theme-docs/package.json
TimonRey: Ok, just don't want the command to be too long :)

emmenko (approved):
💯

Issue #836, opened 2021-01-21
During the pre-build phase, generate API specs using rmf-codegen package


davifantasia:
Requires 2 steps:

1) Setup Java on Vercel
2) Refactor to generate specs pre-build

Related issue and PR:

* https://github.com/commercetools/commercetools-docs-kit/issues/784
* https://github.com/commercetools/commercetools-docs-kit/pull/785

PR #838, opened 2021-01-21
feat: generate api specs during prebuild

davifantasia:
Closes https://github.com/commercetools/commercetools-docs-kit/issues/836

Most of the files changes are deletions and moving RAML files to present the structure we expect in commercetools-docs repo. So here are the relevant files to review:

* `packages/rmf-codegen/scripts/install-vercel-java.js`
* `packages/rmf-codegen/scripts/helpers.js`
* `packages/rmf-codegen/package.json`
* `packages/rmf-codegen/scripts/download-rmf-codegen.js`
* `websites/api-docs-smoke-test/package.json`


.changeset/lovely-jeans-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@commercetools-docs/rmf-codegen": patch
```
emmenko, line 2: Is it a patch or a feature?

api-specs/test/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
emmenko, line 1: You have to add the `api-specs` folder to the list of workspaces.

https://github.com/commercetools/commercetools-docs-kit/blob/5336b9977734fdda6f4a1f3359cfd52ea4a68c06/package.json#L7

api-specs/test/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "name": "@commercetools-api-specs/test",
```
emmenko, line 2: Is there going to be more than one "api-spec" package? Otherwise you can just make it a single package:

```suggestion
  "name": "@commercetools-local/api-specs",
```

packages/rmf-codegen/scripts/install-vercel-java.js
```diff
@@ -0,0 +1,10 @@
+const shelljs = require('shelljs');
+const { javaIsInstalled, abortIfError } = require('./helpers');
+
+if (!javaIsInstalled() && shelljs.which('amazon-linux-extras')) {
+  console.info('Installing Java on Vercel.');
```
emmenko, line 5: Where do you check that this script only runs in Vercel? 🧐

websites/api-docs-smoke-test/package.json
```diff
@@ -4,12 +4,13 @@
...
   "scripts": {
     "start": "yarn develop",
+    "predevelop": "cd ../../api-specs/test && yarn generate-ramldoc",
     "develop": "npx gatsby develop",
     "serve": "npx gatsby serve --prefix-paths -o",
     "clean": "npx gatsby clean",
     "clean:develop": "yarn clean && yarn develop",
     "build": "npx gatsby build --prefix-paths",
-    "prebuild": "yarn clean",
+    "prebuild": "yarn clean && cd ../../api-specs/test && yarn generate-ramldoc",
```
emmenko, line 13:
```suggestion
    "prebuild": "yarn clean && yarn predevelop",
```

emmenko (changes requested):

.changeset/lovely-jeans-repeat.md
davifantasia: I guess it can be a minor.

api-specs/test/package.json
davifantasia: I'm not sure what you mean but we have multiple api specs in commercetools.

packages/rmf-codegen/scripts/install-vercel-java.js
davifantasia: I don't know how to check that, but the log is an indicator that this is expected to run on Vercel.

emmenko: Then mentioning Vercel in the log does not make any sense, sorry.

api-specs/test/package.json
emmenko: You are adding here a package `test` under the `api-specs` folder. My question is if you plan to add other packages besides `test` under `api-specs` or not.

davifantasia: > You are adding here a package `test` under the `api-specs` folder. My question is if you plan to add other packages besides `test` under `api-specs` or not.

What is important to me is the possibility of adding other packages, so if I plan to or not does not matter to me. That is what I explained with my previous response.

packages/rmf-codegen/scripts/install-vercel-java.js
davifantasia: I disagree.

Do you want the log removed?

emmenko: No, simply don't mention Vercel.

emmenko: If you want to keep this specific to Vercel, we should also additionally check for the `VERCEL` environment variable.

Then having the script running only on Vercel does make sense to have.

packages/rmf-codegen/scripts/install-vercel-java.js
```diff
@@ -0,0 +1,10 @@
+const shelljs = require('shelljs');
+const { javaIsInstalled, abortIfError } = require('./helpers');
+
+if (!javaIsInstalled() && shelljs.which('amazon-linux-extras')) {
+  console.info('Installing Java with amazon-linux-extras.');
```
emmenko, line 5: Let's check for the `VERCEL` env variable, so that this script only runs in a Vercel environment. See https://github.com/commercetools/commercetools-docs-kit/pull/838#discussion_r564347300

packages/rmf-codegen/scripts/helpers.js
```diff
@@ -0,0 +1,18 @@
+const shelljs = require('shelljs');
```
nkuehn, line 1: shelljs is overkill here and not a dependency of the package.  Iit's intended to normalize common commands like copying / moving / listing / finding files across operating systems and distros, but for just calling a command it's not needed).

packages/rmf-codegen/scripts/install-vercel-java.js
nkuehn: Yes, this should not happen generally on an amazon linux.  I would consider checking for the `CI=true` env variable instead, it's also set by Vercel and seems to be a consensus among CI vendors (circleCI etc).  

you can then check  for the `VERCEL` variable instead of the amazon linux executable check that requires shelljs.  

--- 

I general, I'm not even really a fan of doing automatic installation on any environment directly from the package install.  Would a more conservative option where we add it to the site specific build scripts suffice?  Most CI systems have better built-in or "canonical" ways to provide java.   E.g. in a GitHub action I would use the `setup-java` action.

nkuehn (commented):
summary: I am not convinced the package should install stuff in the operating system on install.  It's not its role for me.

packages/rmf-codegen/scripts/install-vercel-java.js
emmenko: > I general, I'm not even really a fan of doing automatic installation on any environment directly from the package install.

Me neither. I thought this requirement came from you.

Then I would suggest to drop this and configure CI to install Java.

nkuehn: I do like that we fully "simulate" the required CI setup and typical filesystem layout here in the kit repository to have better testing of the real use case, so a lot of this here is not useless and should actually be done IMHO.  

Also, checking for java being installed and warning about it at install time is good.  But the actual install better belongs into  a top level `vercel-build` NPM script that would then be something like  `amazon-linux-extras install java-openjdk11 && yarn build`.  Vercel automatically picks an NPM script that is called "vercel-build" over the default "build" if such is present.

davifantasia:
Replaced b y https://github.com/commercetools/commercetools-docs-kit/pull/849

PR #842, opened 2021-01-26
fix: size of grey hovering area in child section nav

### LINKED ISSUE #825: ChildSectionNavigator: grey hovering area not properly sized
Here's a situation where the hover effect in the child section navigator is not properly sized.  The grey area is not large enough.  My impression is that the whole layout is not fit for multiline entries, let's add a test for that to the kit, too. 

image

TimonRey:
closes #825

### DIFF
--- a/websites/docs-smoke-test/src/content/components/child-section-navigation.mdx
+++ b/websites/docs-smoke-test/src/content/components/child-section-navigation.mdx
@@ -15,7 +15,7 @@ navLevels: 2
 
 ### ProductVariant
 
-### ProductVariantAvailability
+### Product Variant and Price Availability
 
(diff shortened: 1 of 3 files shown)


websites/docs-smoke-test/src/content/components/child-section-navigation.mdx
```diff
@@ -15,7 +15,7 @@ navLevels: 2
 
 ### ProductVariant
 
-### ProductVariantAvailability
+### Product Variant and Price Availability
```
TimonRey, line 18: I also added two multiline examples for the smoke test

emmenko (approved):

PR #844, opened 2021-01-27
fix: javascript bundle fails on development

TimonRey:
We've noticed that the javascript bundle fails on development at the moment. Looks like gatsby has a bug in the new `2.31.1` version. So I downgraded it to `2.30.3` for now.

More information to the bug:
https://github.com/gatsbyjs/gatsby/issues/29213


emmenko:
Thanks. The changeset is missing.

emmenko (approved):

PR #851, opened 2021-01-28
align arrowUpIcon to page title

### LINKED ISSUE #846: Index nav: align back to top icon to the page title
The simplest implementation is to add 1 pixel top margin to the icon or page title.
Removing the spacing from the icon because it's reused from the ui-kit

TimonRey:
This was already discussed with @Shecki 
closes #846

### DIFF
--- a/packages/gatsby-theme-docs/src/layouts/internals/layout-page-navigation.js
+++ b/packages/gatsby-theme-docs/src/layouts/internals/layout-page-navigation.js
@@ -135,16 +135,17 @@ const LayoutPageNavigation = (props) => {
       <SpacingsStack scale="m">
         <PageTitleLink href="#top">
           <SpacingsInline scale="s">
-            <div
-              css={css`
-                margin-top: ${designSystem.dimensions.heights
-                  .arrowUpIconMarginTop};
-              `}
-            >
+            <div>
               <ArrowUpIcon size="medium" color="neutral60" />
             </div>
             <SpacingsStack scale="xs">
-              <div>{props.pageTitle}</div>
+              <div
+                css={css`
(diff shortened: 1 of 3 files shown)


Shecki:
@TimonRey from my side its fine

packages/gatsby-theme-docs/src/layouts/internals/layout-page-navigation.js
```diff
@@ -135,7 +135,12 @@ const LayoutPageNavigation = (props) => {
       <SpacingsStack scale="m">
         <PageTitleLink href="#top">
           <SpacingsInline scale="s">
-            <div>
+            <div
+              css={css`
+                margin-top: ${designSystem.dimensions.heights
+                  .arrowUpIconMarginTop};
+              `}
+            >
```
emmenko, line 143: Instead of using negative margings, I'd rather add some padding to the element on the right. For example 1 or 2 pixels.

image

emmenko (changes requested):

packages/ui-kit/src/design-system.js
```diff
@@ -124,6 +124,7 @@ export const tokens = {
 
 export const dimensions = {
   heights: {
+    onePixel: '1px',
```
emmenko, line 127: Let's add a comment mentioning the reasons for this special case.

PS: this value is not really used as a height. I think I would prefer to inline it in the code instead of using a variable, because we don't really want to encourage the usage of 1px spacing.

TimonRey: Allright 👍

emmenko (approved):

PR #852, opened 2021-01-28
chore: add outline to string literals

### LINKED ISSUE #797: Design: Striped table has same background as string literals
String literals have a light grey background that is the same as the striped table background 

@Shecki  is this tolerable or do we need to adjust? :
image

TimonRey:
closes #797 

@Shecki link to an example: https://commercetools-docs-kit-git-add-outline-to-string-literals.commercetools.vercel.app/docs-smoke-test/views/links

### DIFF
--- a/packages/ui-kit/src/components/markdown.js
+++ b/packages/ui-kit/src/components/markdown.js
@@ -220,9 +220,9 @@ const TableHeader = styled.th`
 `;
 
 const InlineCode = styled.code`
-  outline: 1px solid ${colors.light.borderSecondary};
   background-color: ${colors.light.surfaceInlineCode};
   border: 1px solid ${colors.light.surfaceInfo};
+  border-color: ${colors.light.borderSecondary};
   border-radius: ${dimensions.spacings.xs};
   color: ${colors.light.textCode};
   font-family: ${typography.fontFamilies.code};
--- a/packages/ui-kit/src/components/markdown.js
+++ b/packages/ui-kit/src/components/markdown.js
@@ -220,6 +220,7 @@ const TableHeader = styled.th`
 `;
 
 const InlineCode = styled.code`
+  outline: 1px solid ${colors.light.borderSecondary};
   background-color: ${colors.light.surfaceInlineCode};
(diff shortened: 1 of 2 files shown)


Shecki (changes requested):
@TimonRey  it now looks like it has two outlines?
The old one should be replaced with the new color.

Also the corner-radius should be used from the old outline.

Thanks

emmenko:
@Shecki is this only relevant for inline code elements inside tables? Or in general?

Shecki:
@emmenko in gerneral

emmenko:
Ok

packages/ui-kit/src/components/markdown.js
```diff
@@ -220,6 +220,7 @@ const TableHeader = styled.th`
 `;
 
 const InlineCode = styled.code`
+  outline: 1px solid ${colors.light.borderSecondary};
```
emmenko, line 223: You shouldn't change the `outline` style, but the border color instead.

emmenko (changes requested):

packages/ui-kit/src/components/markdown.js
```diff
@@ -222,6 +222,7 @@ const TableHeader = styled.th`
 const InlineCode = styled.code`
   background-color: ${colors.light.surfaceInlineCode};
   border: 1px solid ${colors.light.surfaceInfo};
+  border-color: ${colors.light.borderSecondary};
```
emmenko, line 225:
```suggestion
  border: 1px solid ${colors.light.borderSecondary};
```

emmenko (changes requested):

emmenko (approved):

emmenko:
VRTs approved ✅

Shecki:
@TimonRey  can we do something about this view?
image

That it looks like A or B?
image

Shecki:
**Please don't look at the colors here**..
It's just about the form of the background

emmenko:
Option B is not possible.

Option A also does not really make sense because it's essentially a sentence in a new line.

PR #853, opened 2021-01-28
fix: remove top margin on first content element

### LINKED ISSUE #776: Any first content element should have *no* top margin
The Designs assumed that every page has a subtitle and the implementation was approved on that assumption too. 

But some pages do not have a subtitle and directly start with a heading, image etc.   

Esp. in the case of a heading the heading's extra top space occurs at the top of the page, which is not intended (content should start aligned with the index navigation)

So the first element in the first section of the page content needs to have no margin top.  (will be a fancy CSS expression and we may run into stupid specificity issues)

TimonRey:
closes #776

### DIFF
--- a/packages/ui-kit/src/components/markdown.js
+++ b/packages/ui-kit/src/components/markdown.js
@@ -245,6 +245,10 @@ const TypographyPage = styled.div`
     margin-top: ${dimensions.spacings.m};
   }
 
+  section:first-of-type > * {
+    margin-top: 0 !important;
+  }
+
   section > ${H2} + *,
   section > ${H3} + *,
   section > ${H4} + *,
--- a/packages/ui-kit/src/components/markdown.js
+++ b/packages/ui-kit/src/components/markdown.js
@@ -245,7 +245,7 @@ const TypographyPage = styled.div`
     margin-top: ${dimensions.spacings.m};
   }
 
-  section:first-of-type > * {
+  section:first-of-type > *:first-child {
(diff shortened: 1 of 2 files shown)


emmenko (approved):
Selector looks good.

I think @Shecki or @nkuehn should check the Percy snapshots and approve them if the outcome is ok.

Shecki:
from my side its fine, but i can´t approv

packages/ui-kit/src/components/markdown.js
```diff
@@ -245,6 +245,10 @@ const TypographyPage = styled.div`
     margin-top: ${dimensions.spacings.m};
   }
 
+  section:first-of-type > * {
```
nkuehn, line 248: I think the selector is matching too much.  The requirement is to only match the very first content element on a given page, not any direct descendant of the first section. 

```suggestion
  section:first-of-type > *:first-child {
```

nkuehn (changes requested):
Not as intended yet, you can see in the percy shots that it#s also removing a lot of the spacing on the body content.

nkuehn (approved):
Thanks! good to go for this step and the correct way to do it

Only one page - the beta page - has a visual regression that makes obvious that we have no explicit space any more now (in fact, never had).  This situation will be addressed with implementing the extra space for the search bar, which will give all pages a fixed extra spacing between header and body content area independently of the specific header or body content.  

image

emmenko (approved):

PR #860, opened 2021-02-09
feat: add opening animation to top menu

### LINKED ISSUE #564: Design review: top menu open/close animation
Open animation: smooth push down
Close animation: smooth push up

Recommended example https://www.zalando.de/

TimonRey:
closes #564 

Note: Pls let me know what you think about the duration of the animation. For the left and right bar we use 0.5sec at the moment.

### DIFF
--- a/packages/gatsby-theme-docs/src/components/overlay.js
+++ b/packages/gatsby-theme-docs/src/components/overlay.js
@@ -1,9 +1,16 @@
 import styled from '@emotion/styled';
+import { keyframes } from '@emotion/react';
 import { designSystem } from '@commercetools-docs/ui-kit';
 
+const backgroundAnimation = keyframes`
+  from { background-color: rgba(0, 0, 0, 0); }
+  to { background-color: rgba(0, 0, 0, 0.5); }
+`;
+
 const Overlay = styled.div`
-  background-color: rgba(0, 0, 0, 0.5);
-  z-index: ${designSystem.dimensions.stacks.overlay};
+  animation: ${backgroundAnimation} 0.3s;
+  animation-fill-mode: forwards;
+  z-index: ${(props) => props.zIndex || designSystem.dimensions.stacks.overlay};
   position: ${(props) => props.position || 'fixed'};
   top: ${(props) => props.top || '0'};
   left: 0;
--- a/packages/gatsby-theme-docs/src/components/overlay.js
+++ b/packages/gatsby-theme-docs/src/components/overlay.js
@@ -8,7 +8,7 @@ const backgroundAnimation = keyframes`
 `;
 
 const Overlay = styled.div`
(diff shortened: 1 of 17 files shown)


nkuehn (commented):
Speed is nice, fast but clearly an animation.  
But the movement itself is a bit strange, the top nav bar seems to first disappear and then appear as a whole. This moves more elements than necessary.

emmenko (commented):
Hmm the top menu container hovers over the top navigation when "opening", causing a weird effect.

Maybe we should also add some opacity in the animation to try smoothing the effect?

nkuehn:
> Maybe we should also add some opacity in the animation to try smoothing the effect?

I wonder whether the root cause is rather that this animates the wrong DOM element.  I can imaging that the animation we are looking for is letting the  the "Content" element inside the "Container" appear and not the whole "Container"

emmenko (approved):

emmenko:
Code-wise should be fine now from my side.

nkuehn:
Getting there, but visually still not smooth to me. 

It has a hard brightness "blink" at the beginning due to the 50% opacity overlay appearing immediately while the menu appears only gradually.  

My guess would be that the overlay's opacity would also have to ease in with the same timing?

packages/gatsby-theme-docs/src/components/overlay.js
```diff
@@ -1,8 +1,16 @@
...
+
+const backgroundAnimation = keyframes`
+  from { background-color: rgba(0, 0, 0, 0); }
+  to { background-color: rgba(0, 0, 0, 0.5); }
+`;
 
 const Overlay = styled.div`
-  background-color: rgba(0, 0, 0, 0.5);
-  z-index: 20;
+  animation: ${backgroundAnimation} 0.3s;
```
nkuehn, line 11: Nitpick: Please use the same transition curve like the menu, currently the background seems to still change while the menu is already perceived to be fully opened. 

```suggestion
  animation: ${backgroundAnimation} 0.3s ease-out;
```

nkuehn (approved):
Nice, thank you! 

Just one nitpick as a comment.

davifantasia:
@TimonRey The remaining Percy test is not looking right - the snapshot of the left menu is missing the overlay. Please look into that.

nkuehn:
> he remaining Percy test is not looking right

Whether the snapshot has the overlay visible already or not is not an issue (more a testing artifact with animation timing), but one of the shots has actual layout changes.  The viewport size with collapsed left hand menu has a different layout now, the sidebar does not show the microsite title any more but it keeps showing the hamburger menu. I assume it's an artifact of the z-index and DOM refactors. 

I could live with the hamburger menu staying visible but we can't just remove the microsite title (or overlay it like it's here)

PR #866, opened 2021-02-12
chore: simplify tooling deps

emmenko:
* use babel preset from appkit (which was essentially the same as we have here)
* use eslint preset from appkit
* use `@manypkg/cli`
  * `manypkg check` makes sure that packages in the monorepo have the correct deps and versions defined
  * `manypkg upgrade` helps upgrading packages in bulk (replaces `bulk-update-versions`)
* use node 14


TimonRey (approved):
Thanks!

PR #869, opened 2021-02-18
Implement icons for microsite title

### LINKED ISSUE #845: Add icon next to microsite title
- placement and spacing: todo @Shecki pls add the link to invision here
- icons assets: https://drive.google.com/drive/folders/1ZXtyvOtAiysIg9Ky_4Ew6NKKwnmm39cx?usp=sharing
- Icon has to be configured separately from the theme because the icons have to be unique for each site. That's why they should live in the microsite and not in the theme
- we suggest to not add a configuration but to add the icons to a conventional path
 - `src/site-icon.svg`
 - `src/site-icon-small.svg`

open questions:
- What's the initial state when no icon is configured yet? Default icon or no icon? @Shecki 
- Which icon should be used next to the title? Small (25x25) or the big one? @Shecki

TimonRey:
closes #845 

Note: The default icon still has to be made @Shecki


packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -8,6 +8,7 @@ import { BackIcon } from '@commercetools-uikit/icons';
 import SpacingsInline from '@commercetools-uikit/spacings-inline';
 import SpacingsStack from '@commercetools-uikit/spacings-stack';
 import { designSystem, createStyledIcon } from '@commercetools-docs/ui-kit';
+import PlaceholderSiteIcon from '../../overrides/site-icon';
```
nkuehn, line 11: not really a "Placeholder" in this context.

packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -354,9 +355,12 @@ const Sidebar = (props) => {
         <WebsiteTitle theme={theme}>
           <SpacingsStack scale="xs">
             <div>{props.isGlobalBeta && <BetaFlag />}</div>
-            <WebsiteTitleLink as={Link} id="site-title" to="/" theme={theme}>
-              {props.siteTitle}
-            </WebsiteTitleLink>
+            <SpacingsInline scale="s">
+              <PlaceholderSiteIcon />
```
nkuehn, line 359: you could try if it works too when being part of the clickable link (should ideally behave like the top site level icon+text).  But don't invest a lot.

packages/gatsby-theme-docs/src/overrides/site-icon.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import SiteIcon from '../icons/generated/OpenSourceSmall';
```
nkuehn, line 2:
```suggestion
import DefaultSiteIcon from '../icons/generated/OpenSourceSmall';
```

packages/gatsby-theme-docs/src/overrides/site-icon.js
```diff
@@ -0,0 +1,12 @@
...
+
+// A React component to be rendered next to the microsite title
+// eslint-disable-next-line react/display-name
+export default () => {
+  return (
+    <div>
```
nkuehn, line 8: unncessary div wrapper.  If a specific icon needs it it can do that in its override

websites/api-docs-smoke-test/src/@commercetools-docs/gatsby-theme-docs/overrides/site-icon.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import SiteIcon from '../../../icons/HttpApiSmall';
```
nkuehn, line 2: The standard icons need to be provided by the ui kit so we can centrally change the icon, but the mapping to the microsite is still local.

websites/api-docs-smoke-test/src/@commercetools-docs/gatsby-theme-docs/overrides/site-icon.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import SiteIcon from '@commercetools-docs/gatsby-theme-docs/src/icons/generated/HttpApiSmall';
```
nkuehn, line 2: icons folder is exporting the icons, no need to look into generated:
```suggestion
import HttpApiSmall from '@commercetools-docs/gatsby-theme-docs/src/icons/HttpApiSmall';
```
But the pattern is still bad if the icons are intended for consumption from other packages.  In that case they must be exported from the root of the gatsby-theme-docs package so that you can do something like this here:
```suggestion
import HttpApiSmall from '@commercetools-docs/gatsby-theme-docs';
// or
import Icons from '@commercetools-docs/gatsby-theme-docs';

...
<Icons.HttpApiSmall>
```

websites/api-docs-smoke-test/src/@commercetools-docs/gatsby-theme-docs/overrides/site-icon.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Icons } from '@commercetools-docs/gatsby-theme-docs';
```
emmenko, line 2: This looks weird. The icons are shipped with the docs theme but then we override the theme using the icons exported from the theme.

There are 2 ways of doing this:
* keep the icons in the theme, do not expose them, use a config option to tell the theme which icon type to render.
* do not keep the icons in the theme, just use the override placeholder. Consumers of the theme should then provide the icon.

For me it seems like option 1 is the preferred approach.

nkuehn: hi @emmenko, I did go through the requirements with timon before we chose this approach and we had the other two you mentioned on the list, too. 
 
 * theme or ui kit should provide the typically used icons to sites centrally, so the frequently used "main" ones can be updated centrally
* a site needs to use an icon as its home icon but also anywhere else in the content (not only the "own" icon, also for referring to other sites). So some package needs to export them; alternatively they would have to be available in any MDX by default. 
 * the icons are not a 1:1 mapping to the color presets, so binding it to the site color does not work
 * avoid introducing a third way to customize.  We have overrides and we have config flags in the theme config already. 
 * a site should be able to provide an own, custom icon without having to have it accepted into the docs-kit. It's not a use case already now but our goal is to have many more and less prominent pieces of commercetools docs use the kit and it would be slow and inflexible to force all icons into the kit.  (sure, this part is a bit opinionated)

What I would consider improving here is to rather make the icons a part of the docs/ui-kit  and not the theme.

nkuehn (commented):
OK for me, but I suggest to move the icons to the UI kit

nkuehn (approved):

PR #878, opened 2021-03-01
Move searchbox over index nav

### LINKED ISSUE #848: Search box #1: In top nav bar: smaller and in left-hand column space
Design: https://commercetools.invisionapp.com/console/share/5M1WVB136H 

Search box gets a smaller design that fits the right hand index column, too.    It's always right-aligned in the top bar and the width matches the right-hand column in the page designs (200-225px i think). 

Follow-up ticket: #847 

Open questions:
 * confirm where it is in the wide layout

TimonRey:
closes #848

### DIFF
--- a/packages/gatsby-theme-docs/src/components/search-input.js
+++ b/packages/gatsby-theme-docs/src/components/search-input.js
@@ -7,6 +7,7 @@ import { designSystem, createStyledIcon } from '@commercetools-docs/ui-kit';
 import { SearchSvgIcon, SlashSvgIcon } from '../icons';
 
 const SearchIcon = createStyledIcon(SearchSvgIcon);
+const iconHeight = '16px';
 
 const Container = styled.div`
   position: relative;
@@ -24,8 +25,8 @@ const Input = styled.input`
   flex: 1;
   font-family: inherit;
   font-size: ${designSystem.typography.fontSizes.small};
-  height: ${designSystem.dimensions.heights.inputSearchSmall};
-  min-height: ${designSystem.dimensions.heights.inputSearchSmall};
+  height: ${designSystem.dimensions.heights.inputSearchSecondary};
+  min-height: ${designSystem.dimensions.heights.inputSearchSecondary};
   outline: none;
   overflow: hidden;
   padding: 1px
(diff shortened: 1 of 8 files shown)


nkuehn (commented):
Hi, 
to cover all responsive and fluid situations the top header needs to get a lot of the same grid setup that the article body part has.   (Except for the smaller breakpoints where the index nav collapses but in the header the search box is still there).

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -63,27 +63,48 @@ const Container = styled.header`
...
-      [row1-start] 'header-content header-blank' ${designSystem.dimensions
+      [row1-start] 'header-top-menu header-searchbox' ${designSystem.dimensions
         .heights.header} [row1-end]
       / ${designSystem.dimensions.widths.pageContentWithMargins}
       minmax(${designSystem.dimensions.widths.pageNavigation}, 1fr);
   }
+
+  ${(props) =>
+    props.allowWideContentLayout
+      ? `@media screen and (${designSystem.dimensions.viewports.largeDesktop}) {
```
TimonRey, line 74: I've implemented now the viewport for the wide layout. Problem is i'm not sure why the index nav isn't moving with the page size anymore when I make the window smaller. I guess this has something to do with my grid config here which probably has to be adjusted. Suggestions are welcome :)

packages/gatsby-theme-docs/src/components/search-input.js
```diff
@@ -53,7 +55,7 @@ const Input = styled.input`
 const SearchInputIcon = styled.span`
   position: absolute;
   z-index: ${designSystem.dimensions.stacks.base};
-  top: ${designSystem.dimensions.spacings.s};
+  top: ${designSystem.dimensions.specialSpacings.searchIcon};
```
emmenko, line 58: What's the need for this change?

packages/ui-kit/src/design-system.js
```diff
@@ -126,6 +126,7 @@ export const dimensions = {
   heights: {
     header: '48px',
     inputSearch: '32px',
+    inputSearchSmall: '26px',
```
emmenko, line 129: Why do we have two heights here?

TimonRey: The first one is for the searchbox in the childsection navigator. Maybe I could rename it to make it clearer

packages/gatsby-theme-docs/src/components/search-input.js
TimonRey: You need 5px to have the icon perfectly aligned with the input field. And there is no 5px value in the spacings 🤔

emmenko:
@TimonRey I pushed a commit and renamed some of the design tokens

emmenko (approved):

nkuehn (changes requested):
Generally nice, I like the concept. 

in Detail here's one breakpoint situation where the wide layout is fluid, but the search box does not seem to be fluid in width, too:

image

nkuehn (approved):
nice!

PR #879, opened 2021-03-02
Format union type members as heading and make them linkable

nkuehn:
@TohruKohrita  this is a small first step to improve the visual and linking experience of the union type listing

It does:
 * give the child types an anchor to be able to manually link to them
 * format them like the other low-level headings

It does not (will be separate changes)
 * add the child section navigator
 * automatically links _to_ the child types in other locations


TimonRey (approved):
Looks good from tech side 👍

PR #880, opened 2021-03-02
ISO 31 (international) number formating for constants and API

nkuehn:
This is a proposal to format Numbers following ISO-31-0 (now ISO-80000-1), using a "narrow non-breaking space" as the thousand separator.  The choice of scientific / engineering standards over our otherwise used Google Writing Style is in line with our choice of dat formatting which is also not using the US convention. 

Changing another separator is easy, the actual work was in the infrastructure. 

It's applied to constants (aka limits) here and a less frequently used feature of the API docs (constraints on type properties). If you have more ideas where it could be applied to places that are not handwritten MD(X) please comment here. 

Applying the same formatting in MDX requires putting the `&#8239;` escape sequence into numbers as a thousands separator manually.  The Unicode character can be inserted as an actual unicode character, too and would be preserved in the files but I doubt this is useful because there's no way to differentiate it from a normal space.


TimonRey (approved):
Looks good to me 👍

PR #885, opened 2021-03-03
Match package.json engine to .nvmrc for vercel ( Node v14 LTS)

nkuehn:
This tells vercel to also use node 14 for builds and matches the entry in .nvmrc.


TimonRey (approved):
🥳

PR #888, opened 2021-03-04
Gatsby 3 migration

nkuehn:
This is a branch off the [renovate branch for gatsby v3 ](https://github.com/commercetools/commercetools-docs-kit/compare/renovate/major-gatsby-monorepo...gatsby-3-migration?expand=1 )

The smoke test site is running in develop mode, but there's issue(s) remaining with the actual build.  All topics I ran into to get it running again at all were related to webpack 5 not having node.js polyfills by default and seeming to be stricter in general. 

Known issues:
 * (done, ignoring the lib in SSR html render phase) build error with the `tmp` library doing something trying to address node.js APIs (fs.constants, os.constants) (see comment)
 * (done) RSS feed does not load any more (the rss-parser lib needs nodejs polyfills for http).  There do not seem to be attractive alternatives, but the rss-parser also provides a `ParseString` api instead of `ParseUrl`, so we could resort to not polyfilling the nodejs http(s) apis but doing the HTTP fetch ourselves. 
 * (done) for incremental builds to work we need to get rid of any "fs" usage in the SSR build step ( https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v2-to-v3/#using-fs-in-ssr ).  I found one place in a gatsby-ssr.js where we build two icon digests by loading them.  The pattern using raw imports requires us to change the ESLint rules though. 

**please directly contribute here - I just kicked it off for curiosity**


emmenko:
Thanks for kicking it off!

I would suggest to extract the improvements not strictly related to the v3 migration into separate PRs, for example the RSS parsing part.

nkuehn:
Agree, we should use this to get it running and build a list of things to be cleaned up afterwards. 

RSS parsing needs to be looked at nevertheless because the polyfill based approach used here at the moment does build but does not actually work at runtime.

nkuehn:
Stuck on this ugly one in the `tmp`library.  It looks like a transitive dependency. 

```
14 | const path = require('path');
  15 | const crypto = require('crypto');
> 16 | const _c = { fs: fs.constants, os: os.constants };
     |             ^
  17 | const rimraf = require('rimraf');
  18 |
  19 | /*


  WebpackError: TypeError: 'get' on proxy: property 'constants' is a read-only and non-c
  onfigurable data property on the proxy target but the proxy did not return its actual 
  value (expected '[object Object]' but got '[object Object]')
  
  - tmp.js:16 
    [@commercetools-website]/[tmp]/lib/tmp.js:16:13

```

nkuehn:
Status: It builds and seems to run but something with Workflows in the PR and the status checks is broken (they are not even starting to run).

packages/ui-kit/src/components/rss-feeds.js
```diff
@@ -9,7 +9,8 @@ import RssFeedTable from './rss-feed-table';
 async function fetcher(...args) {
   const rssParser = new Parser();
   const promises = args.map(async (url) => {
-    const feedData = await rssParser.parseURL(url);
+    const feedString = await (await fetch(url)).text();
+    const feedData = await rssParser.parseString(feedString);
```
emmenko, line 13: Hmm, we're still importing the library here which is still not ideal.

I think we should drop this library and use a client-only solution that works in browsers.

For example: https://css-tricks.com/how-to-fetch-and-parse-rss-feeds-in-javascript/

Or find a library that does that but that works in browsers without using nodejs APIs.

websites/docs-smoke-test/gatsby-config.js
```diff
@@ -9,8 +9,9 @@ module.exports = {
   // https://www.gatsbyjs.com/docs/reference/release-notes/v2.30
   flags: {
     DEV_SSR: true,
-    FAST_REFRESH: true,
     PRESERVE_WEBPACK_CACHE: true,
+    FAST_DEV: true,
+    PRESERVE_FILE_DOWNLOAD_CACHE: true,
```
emmenko, line 14: Last time I tried using these there were some issues. Have you encountered any problems with these flags being active?

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -499,10 +490,25 @@ exports.onCreateWebpackConfig = ({ actions, getConfig }, themeOptions) => {
...
+  }
   config.resolve = {
     ...config.resolve,
     // Add support for absolute imports
     modules: [path.resolve(__dirname, 'src'), 'node_modules'],
+    fallback: {
+      ...config.resolve.fallback,
+      http: false, // rss-parser related
+      https: false, // rss-parser related
+      timers: false, // rss-parser related
```
emmenko, line 508: These should go away if we drop the `rss-parser` library

emmenko (commented):
Looks promising!

Have you looked into the `gatsby-plugin-image`? I think in our case it should be pretty straight forward as we don't use the image component directly. Still, we should probably use the new package.

emmenko:
I pushed the changes to the renovate branch, we can continue from there.

PR #893, opened 2021-03-08
refactor: replace rss-parser with own custom script, to make it more browser compatible

emmenko:
The `rss-parser` library uses nodejs APIs, which causes some problems with Webpack 5.

Given that we want to fetch and parse RSS feeds on browser runtime, we should use an implementation that uses browser compatible APIs.


package.json
```diff
@@ -40,6 +40,7 @@
     "versions:uikit": "manypkg upgrade @commercetools-uikit"
   },
   "dependencies": {
+    "@babel/helper-define-map": "^7.12.13",
```
emmenko, line 43: No idea why, but apparently this dep somehow does not get installed with Yarn.

packages/ui-kit/src/components/rss-feeds.js
```diff
@@ -1,23 +1,50 @@
 import React from 'react';
 import useSWR from 'swr';
-import Parser from 'rss-parser';
 import PropTypes from 'prop-types';
 import LoadingSpinner from '@commercetools-uikit/loading-spinner';
 import ContentNotifications from './content-notifications';
 import RssFeedTable from './rss-feed-table';
 
+const parseFieldValue = (element) =>
+  element.innerHTML.replace('<![CDATA[', '').replace(']]>', '');
```
nkuehn, line 9:
```suggestion
[ ...element.childNodes ].map(textOrCdataNode => textOrCdataNode.data).join()
```
(beauty contest in sticking to DOM apis  ;-)

package.json
nkuehn: Can you explain what it has to do with the context of this PR?  I can't even find documentation.

emmenko: Nothing to do with this PR. On my machine this dep simply didn't get installed, which prevents the website to start.

packages/ui-kit/src/components/rss-feeds.js
```diff
@@ -1,23 +1,46 @@
...
 import ContentNotifications from './content-notifications';
 import RssFeedTable from './rss-feed-table';
 
+// Inspired by https://css-tricks.com/how-to-fetch-and-parse-rss-feeds-in-javascript/
+async function parseRssFeed(url) {
+  const response = await fetch(url);
+  const rawBody = await response.text();
+  const data = new window.DOMParser().parseFromString(rawBody, 'text/xml');
+
+  const feedTitle = data.querySelector('title').firstChild.data;
```
emmenko, line 14: I went with this simpler approach `firstChild.data` (thanks for the suggestion) for accessing the text value.
We only have single children, so we don't need to use `childNodes`.

davifantasia (approved):
LGTM

PR #895, opened 2021-03-09
chore: upgrade prism-react-renderer to 1.2.0

emmenko:


emmenko:
Hmm the production build still has some issues when running on Cypress, but only the first time the test runs. It seems like some kind of cache issue or something.

If we use Cypress v6 we can then use the retry feature. We should look again into #779 I guess...

emmenko:
No more errors now but some of the theme colors are different now 🤔

PR #903, opened 2021-03-11
refactor: expose hook to format ISO 31-0 number using react-intl

emmenko:
Follow up of #880 

Using the polyfill apparently causes issues in appkit
image

Since we are already using `react-intl`, we can simply use their API to achieve the same result.


packages/gatsby-theme-docs/src/templates/homepage.js
```diff
@@ -10,32 +11,34 @@ import { SEO, ThemeProvider } from '../components';
...
-            {/* This wrapper div is important to ensure the vertical space */}
-            <div>
-              <MDXRenderer>{props.data.contentPage.body}</MDXRenderer>
-            </div>
-          </Markdown.TypographyPage>
-        </MDXProvider>
-      </LayoutContentHomepage>
-    </PageDataContext.Provider>
-  </ThemeProvider>
+  <IntlProvider locale="en">
```
nkuehn, line 14: OK, thanks!  Understood it now

packages/ui-kit/src/hooks/use-iso310-number-formatter.js
```diff
@@ -0,0 +1,24 @@
+import { useCallback } from 'react';
+import { useIntl } from 'react-intl';
+
+// https://en.wikipedia.org/wiki/ISO_31-0#Numbers (now ISO 80000-1, but its text is not public)
+const narrowNonBreakingSpace = '\u202F';
+
+const useISO310NumberFormatter = () => {
```
nkuehn, line 7: @emmenko  To be able to use the function in node.js preprocessing like a remark plugin, too it would be useful to not only expose the hook but also a plain javascript function that takes a number and returns the string.  What do you think?  

I consider adding a remark plugin that does the number formatting automatically in content.

emmenko: Let's do this as a follow up.

PR #904, opened 2021-03-11
Incremental builds on Vercel

nkuehn:
https://www.gatsbyjs.com/docs/reference/release-notes/v3.0/#incremental-builds-in-oss

It works now!  The impact in this repository is not huge (1-2 minutes maybe) because it's not a lot of content and has extra preprocessing steps that are not changing.  See comments for explanation.


nkuehn:
* First run that fully builds but then hopefully caches:  https://vercel.com/commercetools/commercetools-docs-kit/3KZ229MYSAPYsDwn393LnHN2S77o   (7 minutes)
 * Next build with one incremental change:  https://vercel.com/commercetools/commercetools-docs-kit/9tVBf4ysrJsSZ2QACSBdWsDZMSwj 

Learning:
 * it does not work yet - the build cache size is exactly identical despite the fact that it should have grown by adding stuff. 
 * Guesses:
   * framework detection not working? 
   * stuff not actually copied into the cache folders?

nkuehn:
Update: locally incremental and cached **do** work and result in a four-fold increase in build time:
image

The gain is ca. three-fold in the docs-smoke-test site. The gain distributes relatively evenly between initialization, graphQL node sourcing, running the actual page queries and building the pages. 

Signs that are cached build has taken place are that you see lines like "success Cleaning up stale page-data" or "info There are no new or changed html files to build."  which otherwise are not present at all.

package.json
```diff
@@ -9,13 +9,14 @@
     "packages": ["packages/ui-kit"]
   },
   "scripts": {
+    "preinstall": "mkdir -p public && mkdir -p .cache && ls -Rl public > public/files-public-before.txt && ls -Rl .cache > public/files-cache-before.txt",
```
nkuehn, line 12: this could be removed again if we want - it's a debug helper that makes the cache contents before and after the build available in the site URL.  Here in the kit I tend to keep it since debugging incremental builds will likely stay an issue.

package.json
```diff
@@ -9,13 +9,14 @@
...
     "postinstall": "preconstruct dev && manypkg check",
     "clean": "lerna run clean --scope '@commercetools-website/*'",
     "start": "yarn --cwd websites/docs-smoke-test start",
     "build-packages": "preconstruct build",
     "build": "cross-env NODE_ENV=production lerna run build --scope '@commercetools-website/*'",
     "prebuild": "yarn build-packages && yarn generate-icons",
-    "postbuild": "cp -f ./websites/index.html ./public && yarn check-links",
+    "postbuild": "cp -f ./websites/index.html ./public && yarn check-links && ls -Rl .cache > public/files-cache-after.txt && ls -Rl public > public/files-public-after.txt",
```
nkuehn, line 19: Same here, technicall spoken not necessary in production

package.json
```diff
@@ -68,6 +69,7 @@
     "eslint": "7.21.0",
     "eslint-formatter-pretty": "4.0.0",
     "eslint-import-resolver-webpack": "0.13.0",
+    "gatsby": "3.0.4",
```
nkuehn, line 72: Vercel's framework detection looks into package.json and wants to see gatsby there.  It's a redundant entry here because it actually belongs into the workspace folders only.  But it is not bloating beause we automate using the same version everywhere (renovate bot) so it's effectively symbolic.

websites/api-docs-smoke-test/package.json
```diff
@@ -10,8 +10,8 @@
     "clean": "npx gatsby clean",
     "clean:develop": "yarn clean && yarn develop",
     "build": "npx gatsby build --prefix-paths",
-    "prebuild": "yarn clean && yarn predevelop",
-    "postbuild": "rm -rf ../../public/api-docs-smoke-test && mkdir -p ../../public && mv public ../../public/api-docs-smoke-test",
+    "prebuild": "mkdir -p ../../public/api-docs-smoke-test && rm -rf public && mv ../../public/api-docs-smoke-test ./public && mkdir -p ../../.cache/api-docs-smoke-test && rm -rf .cache && mv ../../.cache/api-docs-smoke-test ./.cache && yarn predevelop",
+    "postbuild": "mv public ../../public/api-docs-smoke-test && mv .cache ../../.cache/api-docs-smoke-test",
```
nkuehn, line 14: This is the other required step:  once Vercel has detected gatsyby, it assumes the top level `.cache` and `public` folders to be the cacheable places. 

So we move the site-specific caches into the global location (into a subfolder with the site name) and back.  For `public` this was done anyways after the build.  

It would be nicer if this job would be done centrally and more generic though - maintaining all those long script commands across many sites like in the content repos is error prone (but not super hard either because they don't change).  Maybe a little node script?  

If it were central, we could also set it up so that it only runs on vercel, not locally.

nkuehn:
Decision: We will first try the actual impact in the content repo and then decide whether it's worth it.

nkuehn:
closing this in favor of migrating the script we realized in the content repo to the kit and provide it from the package.

PR #907, opened 2021-03-14
Improve release note generator

### LINKED ISSUE #862: Release note generator needs correction
Currently, if you feed in 
Screenshot 2021-02-11 at 13 59 26

you get the output:
Screenshot 2021-02-11 at 14 02 39

Problems:
1. Even though a type was chosen, the `type` is not rendered. Since `type` is required, this MDX does not render a webpage. The browser sends back an error.
2. There should (probably even must) be 2 spaces before hyphens in the`topics` list. This is auto-corrected by the beautification program once you (insert the content and) save the file, so it will be auto-corrected but I think it's better to be right from the beginning.
3. After giving `type` manually and correcting the YAML list of `topics`, the webpage gets rendered. However, the `date` format is wrong, so we get "Invalid date":
Screenshot 2021-02-11 at 14 23 36
The correct format is `date: 2021-03-03`.

4. `description` gets a wrong value. Somehow, it is set to {value for title} + {value for description}. (Compare the first two screenshots above).
5. We got this wrong in wiki. Sorry! But the docs naming convention for release notes is to lower case the title and replace spaces with hyphens. So could you lower case the generated file names? Thank you!
6. (opinion) About the command line instruction (please see the first screenshot above), I think we better remove the time option because it won't be used. Simple YYYY-MM-DD suffices.
7. (opinion) Again about the command line instruction. This is a small thing and there is nothing wrong with "Title of the release note. Max. 256 character plain text." but max 256 is more important for "description" that come right after this. So I propose "Title of the release note. Write it in plain text."
8. (opinion) For the content of the generated MDX, it is perhaps better to replace the Latin with the following: Please write the content here. To use "Read More", insert `<!--more-->`.
9. Align the wording of the descriptions in the command line... suggestion (start with a verb):
- Insert the title of the release note. [...]
- Describe the key changes and their advantages briefly. [...]
- Fill in the topics for this release note. [...]

10. npm script in VSC after testing.
11. List of topics (selector)... for each microsite (Merchant Center, API, ... have different topics.). Maintenance of topics in the code (not wiki) would be okay.

Thank you!

TimonRey:
closes #862 

If you want to try it out, here's how it works :)
- Open the docs-kit repo in vs code and switch to this branch
- Go to the npm scrips window and open the scripts for either the `docs-smoke-test` or the `api-docs-smoke-test` microsite
- Run the `create-docs-release-note` script. Now you can follow the instructions and create a release note

**Note:**
For showing purposes, I created the config file for the topics only for the `docs-smoke-test`. If you want to see how the script behaves without a config file, run the script in the `api-docs-smoke-test`.

### DIFF
--- a/.changeset/itchy-timers-lick.md
+++ b/.changeset/itchy-timers-lick.md
@@ -2,4 +2,4 @@
 '@commercetools-docs/gatsby-theme-docs': patch
 ---
 
-The release-note file generator was refactored.
+Fix an issue with release note generator CLI.
--- a/.changeset/itchy-timers-lick.md
+++ b/.changeset/itchy-timers-lick.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+The release-note file generator was refactored.
--- a/.changeset/itchy-timers-lick.md
+++ b/.changeset/itchy-timers-lick.md
@@ -1,5 +1,24 @@
 ---
-'@commercetools-docs/gatsby-theme-docs': patch
+'@commercetools-docs/gatsby-theme-docs': minor
 ---
 
 Fix an issue with release note generator CLI.
+
+How to create a release note over your console:
+
+1. Navigate to the microsite where you want to create the release note
+2. Run the following command in your terminal
+
+```
+yarn create-docs-release-note
+```
+
+3. Follow the instructions in your console.
+4. After finishing all instructions in the console, you can find the generated release note in the `src/releases` folder of your microsite.
+
+How to create a release note over over Visual Studio Code:
+
+1. Go to the `npm scripts` section and open the `package.json` file from the microsite in which you want to create a release note.
+2. Run the script `create-docs-release-note`.
+3. Follow the instructions in the console.
+4. After finishing all instructions in the console, you can find the generated release note in the `src/releases` folder of your microsite.
--- a/.changeset/itchy-timers-lick.md
+++ b/.changeset/itchy-timers-lick.md
@@ -4,21 +4,17 @@
 
 Fix an issue with release note generator CLI.
 
-How to create a release note over your console:
+If you want to configure a topic list for a microsite, create a file called `docs-release-notes-config.yml` in the microsite folder. Then you can configure topics in the following format:
 
-1. Navigate to the microsite where you want to create the release note
-2. Run the following command in your terminal
-
+```yml
+topics:
+  Carts:
+    name: Carts
+  Limits:
+    name: Limits
+  Orders:
+    name: Orders
+    description: Describe the topic here
 ```
-yarn create-docs-release-note
-```
 
-3. Follow the instructions in your console.
-4. After finishing all instructions in the console, you can find the generated release note in the `src/releases` folder of your microsite.
-
-How to create a release note over over Visual Studio Code:
-
-1. Go to the `npm scripts` section and open the `package.json` file from the microsite in which you want to create a release note.
-2. Run the script `create-docs-release-note`.
-3. Follow the instructions in the console.
-4. After finishing all instructions in the console, you can find the generated release note in the `src/releases` folder of your microsite.
+The description is optional.
--- a/packages/gatsby-theme-docs/bin/create-release-note.js
+++ b/packages/gatsby-theme-docs/bin/create-release-note.js
@@ -38,7 +38,7 @@ const questions = [
   {
     type: 'text',
     name: 'date',
-    message: 'Insert the release date with the following format: 2021-01-21',
+    message: 'Insert the release date with the following format: YYYY-MM-DD',
     validate: (date) => {
       const checkDateFormat = moment(date, 'YYYY-MM-DD', true);
       return checkDateFormat.isValid() ? true : 'Please use a valid format!';
@@ -69,7 +69,8 @@ const questions = [
 ];
 
 (async () => {
-  if (!fs.existsSync(path.join(process.cwd(), `src/releases`))) {
+  const folderpath = path.join(process.cwd(), `src/releases`);
+  if (!fs.existsSync(folderpath)) {
     console.error(`No "releases" folder found in this directory`);
     process.exit();
   }
@@ -92,7 +93,7 @@ Please write the content here. To use "Read More", insert <!--more-->.
   const filemane = response.title.replace(/\s+/g, '-').toLowerCase();
 
   fs.writeFileSync(
-    path.join(process.cwd(), `src/releases/${slugify(filemane, '-')}.mdx`),
+    path.join(`${folderpath}/${slugify(filemane, '-')}.mdx`),
     template
   );
 })();
--- a/packages/gatsby-theme-docs/bin/create-release-note.js
+++ b/packages/gatsby-theme-docs/bin/create-release-note.js
@@ -69,6 +69,11 @@ const questions = [
 ];
 
 (async () => {
+  if (!fs.existsSync(path.join(process.cwd(), `src/releases`))) {
+    console.error(`No "releases" folder found in this directory`);
+    process.exit();
+  }
+
   const response = await prompts(questions);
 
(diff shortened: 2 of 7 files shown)


packages/gatsby-theme-docs/bin/create-release-note.js
```diff
@@ -35,36 +36,44 @@ const questions = [
...
     name: 'topics',
     message:
-      'Write down the topics for this release note. Seperate them with commas.',
+      'Fill in the topics for this release note. Seperate them with commas.',
     separator: ',',
   },
 ];
 
 (async () => {
+  if (!fs.existsSync(path.join(process.cwd(), `src/releases`))) {
```
emmenko, line 72: Nit: Declare the releases folder path as a variable since you also use it when writing the file.

emmenko (commented):
Left some suggestions. And then there is the config part that we discussed by using https://www.npmjs.com/package/cosmiconfig

.changeset/itchy-timers-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+The release-note file generator was refactored.
```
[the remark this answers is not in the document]
TimonRey: Thanks for the hint :)

websites/api-docs-smoke-test/package.json
```diff
@@ -9,6 +9,7 @@
     "serve": "npx gatsby serve --prefix-paths -o",
     "clean": "npx gatsby clean",
     "clean:develop": "yarn clean && yarn develop",
+    "create-docs-release-note": "PATH=$(npm bin):$PATH create-docs-release-note",
```
TimonRey, line 12: Note: I added this to the package.json scripts of each microsite so that the script can be run in the "npm scripts" window in vs code. Otherwise it's not visible

emmenko: Hmm this looks wrong. What are you trying to do here? 🤔

packages/gatsby-theme-docs/bin/create-release-note.js
```diff
@@ -4,68 +4,112 @@ const fs = require('fs');
...
+    console.error(`No "releases" folder found in this directory`);
+    process.exit();
+  }
+
+  const getQuestions = () => {
+    const moduleName = 'release-note-topic-config';
+    const explorerSync = cosmiconfigSync(moduleName, {
+      searchPlaces: [`${moduleName}.yml`],
+    });
+    const configFile = explorerSync.search();
```
emmenko, line 82: This can be defined statically in the file, no need to defined it within the function.

packages/gatsby-theme-docs/bin/create-release-note.js
```diff
@@ -4,68 +4,112 @@ const fs = require('fs');
...
 (async () => {
-  const response = await prompts(questions);
+  const folderpath = path.join(process.cwd(), `src/releases`);
+  if (!fs.existsSync(folderpath)) {
+    console.error(`No "releases" folder found in this directory`);
+    process.exit();
+  }
+
+  const getQuestions = () => {
+    const moduleName = 'release-note-topic-config';
```
emmenko, line 78: I would keep this config file "generic" to the script, and not specific to the topics. The topics is then an option within the config.

```suggestion
    const moduleName = 'docs-release-notes-config';
```

packages/gatsby-theme-docs/bin/create-release-note.js
```diff
@@ -4,68 +4,112 @@ const fs = require('fs');
...
+    const explorerSync = cosmiconfigSync(moduleName, {
+      searchPlaces: [`${moduleName}.yml`],
```
emmenko, line 80:
```suggestion
      searchPlaces: [`${moduleName}.yml`, `${moduleName}.yaml`],
```

websites/docs-smoke-test/release-note-topic-config.yml
```diff
@@ -0,0 +1,7 @@
+# Provide a list of topics
+Carts:
+  name: Carts
+Limits:
+  name: Limits
+Orders:
+  name: Orders
```
emmenko, line 7:
```suggestion
# Provide a list of topics
topics:
  - name: Carts
    description:
  - name: Limits
    description:
  - name: Orders
    description:
```

emmenko (changes requested):
Thanks for making things more configurable. I left some suggestions.

packages/gatsby-theme-docs/bin/create-release-note.js
TimonRey: Hmm okay. Yes I thought about that, too. Thought it's more clear that way because it suggests that you can only configure topics, nothing else. But that's fine for me.

.changeset/itchy-timers-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
```
emmenko, line 2: We are introducing a new feature now (the config), so this should be a minor bump.

```suggestion
'@commercetools-docs/gatsby-theme-docs': minor
```

.changeset/itchy-timers-lick.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fix an issue with release note generator CLI.
```
emmenko, line 5: Describe also how to use the config file.

packages/gatsby-theme-docs/bin/create-release-note.js
```diff
@@ -4,68 +4,112 @@ const fs = require('fs');
...
+  if (!fs.existsSync(folderpath)) {
+    console.error(`No "releases" folder found in this directory`);
+    process.exit();
+  }
+
+  const moduleName = 'docs-release-notes-config';
+  const explorerSync = cosmiconfigSync(moduleName, {
+    searchPlaces: [`${moduleName}.yml`, `${moduleName}.yaml`],
+  });
+  const configFile = explorerSync.search();
```
emmenko, line 81: Move this outside of the function, before `const selectQuestions`

packages/gatsby-theme-docs/bin/create-release-note.js
```diff
@@ -4,68 +4,112 @@ const fs = require('fs');
...
+
+  const getQuestions = () => {
+    if (configFile) {
+      const topics = configFile.config.topics;
+
+      const topicChoices = (topics) => {
+        return Object.keys(topics).map((topic) => {
```
emmenko, line 88: Topics should be a list, you don't need `Object.keys`

packages/gatsby-theme-docs/bin/create-release-note.js
```diff
@@ -4,68 +4,112 @@ const fs = require('fs');
...
+          const title = topics[topic].name;
+          const description = topics[topic].description;
+          return { title, value: title, description: description || '' };
```
emmenko, line 91:
```suggestion
          const title = topic.name;
          const description = topic.description || '';
          return { title, value: title, description };
```

.changeset/itchy-timers-lick.md
```diff
@@ -0,0 +1,24 @@
...
+
+3. Follow the instructions in your console.
+4. After finishing all instructions in the console, you can find the generated release note in the `src/releases` folder of your microsite.
+
+How to create a release note over over Visual Studio Code:
+
+1. Go to the `npm scripts` section and open the `package.json` file from the microsite in which you want to create a release note.
+2. Run the script `create-docs-release-note`.
+3. Follow the instructions in the console.
+4. After finishing all instructions in the console, you can find the generated release note in the `src/releases` folder of your microsite.
```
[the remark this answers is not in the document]
TimonRey: Right 😅

emmenko (approved):
Nice! 👌

adinakleine (approved):
Lgtm 👍

PR #911, opened 2021-03-15
Fix overlapping entries in child section nav

### LINKED ISSUE #899: Child navigation section visual regression
Link: https://docs.commercetools.com/api/projects/order-edits#detailed-list-of-stagedorderupdateactions

Image:
Screenshot 2021-03-10 at 15 05 57

TimonRey:
is part of #899 
This pull request implements code to prevent the entries from overlapping if the line is too long. There is still the discussion in the issue about whether the section names are correct.

This was already discussed and approved from design @Shecki

### DIFF
--- a/.changeset/silly-actors-remain.md
+++ b/.changeset/silly-actors-remain.md
@@ -3,4 +3,4 @@
 '@commercetools-website/docs-smoke-test': patch
 ---
 
-A bug was fixed in the child section navigator.
+Fix overflowing issue with long words in `<ChildSectionNav>`
--- a/.changeset/silly-actors-remain.md
+++ b/.changeset/silly-actors-remain.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+'@commercetools-website/docs-smoke-test': patch
+---
+
+A bug was fixed in the child section navigator.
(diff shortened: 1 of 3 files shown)


websites/docs-smoke-test/src/content/components/child-section-navigation.mdx
```diff
@@ -33,7 +33,7 @@ navLevels: 2
 
 ### Price Selection
 
-### Attribute
+### Set ShippingAddressAndCustomShippingMethod
```
emmenko, line 36: Nit 

```suggestion
### This is a VeryLongWordThatDoesNotHaveAnySpaceAndThereforeItOverflows
```

.changeset/silly-actors-remain.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+'@commercetools-website/docs-smoke-test': patch
+---
+
+A bug was fixed in the child section navigator.
```
emmenko, line 6:
```suggestion
Fix overflowing issue with long words in `<ChildSectionNav>`
```

packages/gatsby-theme-docs/src/components/child-sections-nav.js
```diff
@@ -107,7 +107,7 @@ const Link = styled.a`
   font-size: ${designSystem.typography.fontSizes.small};
   color: ${designSystem.colors.light.textSecondary};
   text-decoration: none;
-  overflow-wrap: break-word;
+  overflow-wrap: anywhere;
```
TimonRey, line 110: As an alternative, I've used `anywhere` here to break words if there are too long. Maybe you can give me some feedback here which solution you like more

emmenko: Good point, I think `anywhere` is the only value that works.

emmenko (approved):

PR #913, opened 2021-03-16
chore: update uikit to 11.2.0

emmenko:


packages/gatsby-theme-docs/src/components/content-pagination.js
```diff
@@ -72,7 +70,7 @@ const PaginationLink = (props) => (
     to={trimTrailingSlash(props.linkTo)}
     align={props.direction === 'left' ? 'right' : 'left'}
   >
-    <PaginationButton>
+    <PaginationButton insetScale="s">
```
emmenko, line 73: FYI: the `<Card>` component (from uikit) now accepts a `insetScale` option to control the padding, so need to override the style in this case.

PR #914, opened 2021-03-16
chore(deps): bump codegen

jenschude:


TohruKohrita:
@jens Is this the update that we talked about for the import api? If so, I would like to talk to you before merging this.

PR #925, opened 2021-03-24
Implement searchbox in index nav

### LINKED ISSUE #847: Search box #2: in right column when top bar is scrolled off screen
When the search box in the top bar is not visible any more a search box should appear above the "back to top" icon and page title in the right-hand column.  

Design: https://commercetools.invisionapp.com/console/share/5M1WVB136H 

The "box" itself has no functionality, it opens the search overlay.  It's rather a box-styled button that opens the search overlay, so it could be doable to have two of them. 

---- 

Problem with the current design:  The second "step" moves the index nav down compared to the first "step" where the search is not yet visible.  As discussed previously the space for the box must be free already to not move everything below around.  

----

FYI here's a situation with all the stuff crammed in.  Where should the box be? 
(I think the search box belongs above the start of the document text and below the title -> in between them ). 

image

TimonRey:
closes #847


nkuehn:
Hi @TimonRey  I just tried the implementation locally and it's not doing what was discussed and designed.   We defined that the layout has a full "line" of height reserved for the search box even if it not currently visible so that scrolling down does not lead to the janky movement of the index nav against the direction of scrolling (which it has now).   To prepare for that layout we had already temporarily removed the distance between the Title and the content. 

space for search bar

packages/gatsby-theme-docs/src/layouts/internals/layout-page-release-notes-filters.js
```diff
@@ -32,7 +32,6 @@ const SlidingContainer = styled.div`
 const GridContainer = styled.div`
   display: none;
   border-left: 1px solid ${designSystem.colors.light.borderPrimary};
-  margin-top: ${designSystem.dimensions.spacings.wide};
```
TimonRey, line 35: Is there a reason why we have this huge margin over the release note filter?

nkuehn: The spacing between page titel (in that case "Platform Release Notes") and the beginnning of the actual page content is supposed to be mirrored in the left column. 

In the new concept with the search bar,  the placeholder "line" in the grid for the search bar could become that space.  It's smaller than currently but I think the spacing of the release notes should be identical to the content pages.  

I suggest to you try to get the release notes into a spacing and grid setup as similar to the content pages as possibe and then show it to sven (need to do a video call because it's not on staging)

TimonRey: Yes, I had a meeting with sven yesterday. He likes the current implementation. I also showed hin the changes for the release note pages. Thanks for the explanation 👍

nkuehn:
@TimonRey please take a look over the percy test results yourself, too to check for regressions. 

For example, the home page with cards has a regression (no space between the banner and the first content heading any more:

image

Here's a robustness edge case when the index nav becomes scrollable (if it's taller than the page).  
The box looks like it's absolutely positioned which I am not sure whether it's a good idea in general.  I did not inspect your grid layout design yet, but I suggest you think about a more conservative setup that gives the box a normal box model layout place that's **inside** the container that scrolls when the page is not tall enough. 

image

image

same effect at the end of the release notes.  T

image

nkuehn:
PS: sorry, forgot the fundamentals: It's very nice in general and the effect is much like we imagined it.

nkuehn (approved):
Visual and interactive outcome is good now, very nice!

PR #938, opened 2021-04-13
Remove union type rendering feature

### LINKED ISSUE #922: Remove the union-type rendering feature.
We are not going to use the the union-type rendering feature, so we decided to remove it in the docs daily on 23.03.2021, which Nikolaus also participated in. 

By the union-type rendering feature, I mean the custom component for union-like types.

TimonRey:
closes #922


packages/gatsby-theme-api-docs/src/components/type/type.js
```diff
@@ -1,12 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FullWidthContainer } from '@commercetools-docs/gatsby-theme-docs';
-import filterOutApiTypeSubtypes from '../../utils/filter-out-api-subtypes';
 import { generateTypeURN } from '../../utils/ctp-urn';
 import { useApiTypes } from '../../hooks/use-api-types';
 import reportError from '../../utils/report-error';
 import Children from './children';
```
nkuehn, line 7: I don't see a point why we still need the "children" abstraction - can the "children" component be merged directly into "type?

nkuehn (changes requested):
- see comment, I think there is more potential for refactoring
 - the test page looks like there still is some "magic" rendering (but I might err and it's manually positioned types)

davifantasia (approved):
LGTM

PR #944, opened 2021-04-19
feat: link manually documented type locations

### LINKED ISSUE #864: Link manually documented common types from RAML generated type documentation
Both Change History and Import API sometimes use the common data types defined in the HTTP API ( https://docs.commercetools.com/api/types ). For example, LocalizedString:

Screenshot 2021-02-12 at 10 03 43

Just like CategoryKeyReference above, **we want to enable a link for LocalizedString.** Currently to do so, we need to document Localized String, in this case, in the Import API micro-site. However, we prefer not to do so for the types already documented in the HTTP API doc.

An example from Change History doc:
Screenshot 2021-02-12 at 10 16 33

TimonRey:
closes #864


.changeset/real-pots-drive.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-docs/gatsby-theme-api-docs': minor
+'@commercetools-api-specs/test': patch
+'@commercetools-website/api-docs-smoke-test': patch
```
nkuehn, line 4: neither tests nor the test websites are released artifacts.  Please remove them from the changeset.

packages/gatsby-theme-api-docs/gatsby-config.js
```diff
@@ -21,6 +21,13 @@ module.exports = (themeOptions = {}) => {
           path: path.resolve('./src/api-specs'),
         },
       },
+      {
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'type-locations',
+          path: path.resolve('./src/type-locations'),
```
nkuehn, line 28: Why do we need a special folder for a system that is restricted to using one single file with a conventional file name?  It would be exposed already on GraphQL if it 's in the data folder and you could  still give it its own custom GraphQL facade if you want.

packages/gatsby-theme-api-docs/src/hooks/use-type-location-overrides.js
```diff
@@ -0,0 +1,18 @@
+import { useStaticQuery, graphql } from 'gatsby';
+
+export const useTypeLocationOverrides = () => {
+  const queryResult = useStaticQuery(
+    graphql`
+      {
+        allTypeLocationsYaml {
+          nodes {
+            type
```
nkuehn, line 9: You must know the apiKey, too.  A same named type can be in multiple APIs

packages/gatsby-theme-api-docs/src/hooks/use-type-locations.js
```diff
@@ -56,8 +57,15 @@ export const useTypeLocations = () => {
       }
     `
   );
+  const typeLocations = convertComponentInMdxToTypeLocations(queryResult);
+  const typeLocationOverrides = useTypeLocationOverrides().reduce(
+    (list, type) => {
+      return { ...list, [type.type]: { urlAnchorTag: type.path } };
```
nkuehn, line 63: As discussed I am not happy with this data design.   The code that consumes the result of the hook will now have to have lots of null checks and special logic for type locations because overridden types will have some fields not set that types from the introspection hook would have set. 

In our review, we found that we can reduce the data structure to being just an href, not exposing the innards of the original GraphQL structures as much.

websites/api-docs-smoke-test/src/type-locations/type-locations.yaml
```diff
@@ -0,0 +1,4 @@
+- type: test__ExternalLinkTestType
```
nkuehn, line 1: TBD I don't like the data structure design. It's not designed user-first but is backwards derived from the internals of an existing implementation data structure.    Please try to learn building features like this test-first // specification-first.  You write the configuration,  the data file and the usage in the text MDX like it would be ideal and intuitive for the user.  Only then you start implementing. 

Not ideal here: 
 * a single file in a same named folder
 * It's hard to understand why the navigation and top menu yaml is a file in "data" but this one has a custom data structure in a different folder 
 * authors have to know an internal naming convention with two underscores (that's actuallly just an internal hash keying, something like this should never be exposed to a user)
 * "path" is not correct, it contains a hash too. 
 *  example data should also cover links to 
    1. other microsite
    2. completely external URLs
 * since the content is in fact a full href  or URL, consider naming it that way  (or check how authors write such stuff in other places and make it consistent - e.g. in the menu navigation data files)

nkuehn (changes requested):
Sorry, a lot of what we discussed in our grooming seems to have gotten lost. 

Take it as an exercise in learning to build a user-centric feature and not an incremental hack whose design is derived from current implementation structures.

packages/gatsby-theme-api-docs/gatsby-node.js
```diff
@@ -9,11 +9,62 @@ const fs = require('fs');
...
+  };
+
+  actions.createNode({
+    ...typeLocationData,
+    // Required fields
+    id: createNodeId(`${node.id} >>> Type location`),
+    parent: node.id,
+    children: [],
+    internal: {
+      type: `Type location`,
```
nkuehn, line 63: Not convinced of the Type name with a space - it will appear in the GraphQL API that way or auto-slugified

UpperCamelCase is the current convention: https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/gatsby-theme-docs/gatsby-node.js#L283

packages/gatsby-theme-api-docs/gatsby-node.js
```diff
@@ -9,11 +9,62 @@ const fs = require('fs');
...
+      contentDigest: createContentDigest(typeLocationData),
+      content: JSON.stringify(typeLocationData),
+      description: `Type location data`,
```
nkuehn, line 66: These are not the complete type locations, just the overrides.  The description would be a good place to tell.

packages/gatsby-theme-api-docs/src/hooks/use-type-locations.js
```diff
@@ -23,10 +24,6 @@ const convertComponentInMdxToTypeLocations = (data) =>
     return {
       ...typeLocations,
       [`${apiKey}__${name}`]: {
-        apiKey,
-        name,
-        slug,
-        urn,
         urlAnchorTag,
```
nkuehn, line 27: `urlAnchorTag` is not a good name any more.  it's a URL now and not just an anchor.

nkuehn (commented):
Quick code review, please have someone else approve the VRTs, it's unlikely I get it done today

nkuehn (changes requested):
* Documentation in READMEs missing completely
 * Unused GraphQL Facade

packages/gatsby-theme-api-docs/README.md
```diff
@@ -67,6 +67,28 @@ The file location determines the `apiKey` through which it can be addressed:
 - `src/api-specs/bar/api.raml` -> apiKey `bar`
 - `src/api-specs/baz/baz.raml` -> apiKey `baz`
 
+### Overriding Common Types
```
nkuehn, line 70:
```suggestion
### Overriding the location of certain API types
```

packages/gatsby-theme-api-docs/README.md
```diff
@@ -67,6 +67,28 @@ The file location determines the `apiKey` through which it can be addressed:
...
+
+If you want to override common types, create a file called `type-locations.yaml` under the `src/data` folder of the website. Use the following format to override the types you want:
```
nkuehn, line 72:
```suggestion
If you want to override the location to link to for certain API data types, create a file called `type-locations.yaml` under the `src/data` folder of the website. This can be useful to choose manual markdown based documentation for certain types or to link ot shared types used on other websites like public standards. Use the following format to override the type locations:
```

packages/gatsby-theme-api-docs/README.md
```diff
@@ -67,6 +67,28 @@ The file location determines the `apiKey` through which it can be addressed:
...
+
+```yaml
+- api: import
+  locations:
+    - type: AssetDimensions
+      href: /import-api/product-variant#assetdimensions
```
nkuehn, line 78:
```suggestion
      href: /product-variant#assetdimensions
```

packages/gatsby-theme-api-docs/README.md
```diff
@@ -67,6 +67,28 @@ The file location determines the `apiKey` through which it can be addressed:
...
+    - type: DiscountedPrice
+      href: /import-api/order#discountedprice
```
nkuehn, line 80:
```suggestion
      href: /../other-microsite/order#discountedprice
```

packages/gatsby-theme-api-docs/README.md
```diff
@@ -67,6 +67,28 @@ The file location determines the `apiKey` through which it can be addressed:
...
+- api: history
+  locations:
+    - type: LocalizedLabel
+      href: /api/history/change-history#localizedlabel
```
nkuehn, line 84:
```suggestion
      href: https://example.com/some/other/location/on/the/web
```

nkuehn (approved):
changes to documentation only, will approve myself and merge!

PR #960, opened 2021-05-06
Use same animation transition for all animations

### LINKED ISSUE #871: Homogenize Animation transitions and sequences
We now have four different opening animations for the top, main (left) and index (right) navigation and the search overlay.  Each use slightly different timings and sequences.  The best reviewed and agreed upon one is the new one for the top menu.   

I suggest to apply the exact same timing, sequence and transition to the main and index nav, too.  

 * Currently the left (main) is first finishing the overlay opacity and then starting to slide in the menu.  
 * The right (index) is delaying the menu slide-in but it's partially at the same time as the overlay opacity.  
 * The top nav directly starts both transitions and is best for users because it gives more immediate feedback and does not force to wait longer than needed (menu openings are frequent click flows in docs). 
 * The search appears without any transition. 

I vote for using the exact same one and rather use a fast compact one.

TimonRey:
is part of #871

### DIFF
--- a/.changeset/cool-llamas-change.md
+++ b/.changeset/cool-llamas-change.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Use same animation transition for all animations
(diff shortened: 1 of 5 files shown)


davifantasia (approved):
LGTM

PR #964, opened 2021-05-13
Visualize regex properties

### LINKED ISSUE #863: API docs: visualize "pattern properties" (regexes that allow varying object property names)
For example, the commercetools LocalizedString object type allows arbitrary keys as long as they are a valid locale code. 

The docs kit RAML API components are handling the case without error and render the regular expression as the property name:

image

So at least it works, but it's very opaque to the user because it's implicit that we don't mean this explicit string but any string that matches the given regular expression. 

RAML spec for it:  https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#additional-properties 

In incremental imrovement to the renderer would be to detect strings that start and and with a forward slash.  For example, if the prop is `/myRegexExample.*/`, the UI layout would be different and explicitly say something like "Any property matching `myRegexExample.*`"

TimonRey:
closes #863

### DIFF
--- a/.changeset/soft-moose-nail.md
+++ b/.changeset/soft-moose-nail.md
@@ -1,7 +1,5 @@
 ---
-'@commercetools-api-specs/test': patch
 '@commercetools-docs/gatsby-theme-api-docs': patch
-'@commercetools-website/api-docs-smoke-test': patch
 ---
 
 Visualize regular expression properties
--- a/.changeset/soft-moose-nail.md
+++ b/.changeset/soft-moose-nail.md
@@ -1,0 +1,7 @@
+---
+'@commercetools-api-specs/test': patch
+'@commercetools-docs/gatsby-theme-api-docs': patch
+'@commercetools-website/api-docs-smoke-test': patch
+---
+
+Visualize regular expression properties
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -28,10 +29,37 @@ const NameType = ({ apiKey, property }) => {
     apiKey,
   });
 
+  const isRegex = (regex) =>
+    regex.charAt(0) === '/' && regex.charAt(regex.length - 1) === '/';
```
nkuehn, line 33: nitpick on naming: the parameter would be better named "string".  When passing, it's unknown whether it's a regex or not.  Or, in other words:  The function provides the anser to the question "is this regex a regex" (which is pointless) :-)

.changeset/soft-moose-nail.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-api-specs/test': patch
+'@commercetools-docs/gatsby-theme-api-docs': patch
+'@commercetools-website/api-docs-smoke-test': patch
```
nkuehn, line 4: known by now:  Only provide changesets for published packages.  I.e. only those in `@commercetools-docs/`

TimonRey: Right, sorry for that 👍🏼

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
TimonRey: True 💯

nkuehn (approved):

PR #966, opened 2021-05-14
chore: deny x frame

davifantasia:
make parallel change as requested in https://github.com/commercetools/commercetools-docs/issues/886


nkuehn (changes requested):
This will not have any effect since the entry will never be reached.  And it's not even the right syntax. 

Here is the documentation:
https://vercel.com/docs/configuration#project/routes

Under "Headers" in that chapter you see that you need to set `continue: true` and put the rule to the top of the list.  Plus, in the legacy "routes" configuration style  a header is set as a key-value pair and not as an object with key and value fields. 

You can alternatively choose to migrate the whole vercel.json to the new features.  The routing configuration is not really complex.

nkuehn (approved):
+1  

plus let's not forget to do it in the content repo, too (the issue might be moved on automatically on the borad)

PR #973, opened 2021-05-19
Terms of beta policy in writing style checks

adinakleine:
Acc. to our beta policy we need to add "closed beta" and "public beta" to the automated writing style checks.

### DIFF
--- a/.changeset/brave-impalas-watch.md
+++ b/.changeset/brave-impalas-watch.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-docs/writing-style": patch
+---
+
+Add unused software release terms of beta policy in writing style checks
--- a/packages/writing-style/styles/commercetools/WordList.yml
+++ b/packages/writing-style/styles/commercetools/WordList.yml
@@ -36,7 +36,9 @@ swap:
   i18n: internationalization
   l10n: localization
   long press: touch & hold
+  open beta: public beta # acc. to our beta policy
   open-source: open source
+  private beta: closed beta # acc. to our beta policy
   regex: regular expression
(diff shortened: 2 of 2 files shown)


TohruKohrita:
Just wondering if we are allowed to simply say "beta" without closed or public, aside from the flag.

**If** not, we might also consider including this in the list. Something like the line 29. (I am assuming that Vale suggests based on this list.) After all, the attempt to use simple "beta" may occur as often as private beta and open beta.

adinakleine:
> Just wondering if we are allowed to simply say "beta" without closed or public, aside from the flag.
> 
> **If** not, we might also consider including this in the list. Something like the line 29. (I am assuming that Vale suggests based on this list.) After all, the attempt to use simple "beta" may occur as often as private beta and open beta.
> 
> On the other hand (though this is not relevant to this PR any longer), if we are allowed to keep using simple "beta", we may need to clarify what it means. (Maybe it's already done. If so, sorry for my ignorance.)

If we use "beta" in the public docs, we do mean always "public beta".  So no need to add something for this.
The term "beta" will be explained in the API contract / beta policy.

nkuehn (approved):
looks good to me, we can always add more if an issue actually occurs.

TohruKohrita:
> If we use "beta" in the public docs, we do mean always "public beta". So no need to add something for this.

Thanks for the clarification on the usage and the meaning of simple beta. 

> The term "beta" will be explained in the API contract / beta policy.

Right. In fact, this is where I first began to wonder about simple beta.

Anyways, thanks. (Also, fortunately we do not use the term beta so much in the doc. )

PR #975, opened 2021-05-20
disable search input when page is still loading

### LINKED ISSUE #974: search input active, although the page is not yet fully loaded
We have noticed that sometimes, when the page is not yet fully loaded, the search input field is still clickable, but the search dialog is not triggered. In this case the user can write in the actual field, which of course is not intended.

Kapture 2021-05-20 at 12 08 13

My suggestion is to disable the search input field when is page is not yet fully loaded.

timonrey:
closes #974 

Unfortunately, the only way to test it is to run it locally
How to test it: open the website in a new tab or do a cash update. Then you can see that you cannot click the search input until the page loads completely.

### DIFF
--- a/.changeset/beige-tables-kick.md
+++ b/.changeset/beige-tables-kick.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Disable search input field when page is still loading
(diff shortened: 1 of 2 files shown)


davifantasia (approved):
LGTM

PR #986, opened 2021-06-14
App Size: Remove dependency to two larger JS libs (moment and parse5)

nkuehn:
Following our discussion concerning the size and performance,  this PR does 

 * tweak documentation and start scripts on how to do the bundle size analysis. you can run `yarn build:analyze` as a shortcut to run the bundle analysis on 
 * remove the dependency on the full HTML parser "parse5" by doing the metaproperty parsing of code blocks with an own regex based implementation.  (I know, I voted for using the parser, but that was at a time when we needed it for other purposes anyways)
 * remove dependency on moment.js  (time library).  It's only gone from our code now but still in some bundle because it's used in the UI Kit calendar widget.   But it's not in all bundles any more now (it was due the RSS feature) but only in the release notes list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,11 +37,14 @@ Build the packages
 $ yarn build
 ```
 
-Start the [docs-smoke-test website](./websites/docs-smoke-test):
+Start the [docs-smoke-test website](./websites/docs-smoke-test) or the [api-docs-smoke-test website](./websites/api-docs-smoke-test):
 
 ```bash
+$ cd websites/docs-smoke-test
 $ yarn start
 ```
+
+When locally building the API docs smoke test site, a [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) based size analysis is run and automatically opened. Keep in mind that the analysis only covers the JS content, GatsbyJS GraphQL data live in separate JSON files in the `/public/page-data` folder, the `sq` subfolder contains the static query results.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ $ cd websites/docs-smoke-test
 $ yarn start
 ```
 
-When locally building the API docs smoke test site, a [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) based size analysis is run and automatically opened. Keep in mind that the analysis only covers the JS content, GatsbyJS GraphQL data live in separate JSON files in the `/public/page-data` folder, the `sq` subfolder contains the static query results.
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -37,12 +37,15 @@ Build the packages
 $ yarn build
 ```
 
-Start the [docs-smoke-test website](./websites/docs-smoke-test):
+Start the [docs-smoke-test website](./websites/docs-smoke-test) or the [api-docs-smoke-test website](./websites/api-docs-smoke-test):
 
 ```bash
+$ cd websites/docs-smoke-test
 $ yarn start
 ```
```
davifantasia, line 45: Start the [docs-smoke-test website](./websites/docs-smoke-test):

```bash
$ cd websites/docs-smoke-test
$ yarn start
```

Start the [api-docs-smoke-test website](./websites/api-docs-smoke-test):

```bash
$ cd websites/api-docs-smoke-test
$ yarn start
```

packages/gatsby-theme-docs/src/layouts/internals/layout-page-navigation.js
```diff
@@ -174,6 +177,7 @@ const LayoutPageNavigation = (props) => {
 
   const navigationContainer = (
     <nav
+      key="search-container"
```
davifantasia, line 180: `key="search-container"` is used twice, here and on line 160

```suggestion
      key="navigation-container"
```

davifantasia (approved):

davifantasia:
I meant to `request changes`, not approve. I tried to change it but it's not working on GitHub at the moment.

nkuehn:
> I meant to `request changes`, not `approve`. I tried to change it but it's not working on GitHub at the moment.

No problem, I dismissed and re-requested it. Thanks for the review!

packages/gatsby-theme-docs/src/layouts/internals/layout-page-navigation.js
nkuehn: copypasta...  done

README.md
nkuehn: much better, right.

timonrey (commented):
I tried to run the `create-docs-release-note` script and I got an error here:
Bildschirmfoto 2021-06-16 um 11 42 11

Looks like there there is a problem with babel here? Maybe we have different configurations. We need to keep this simple so that the tech-writers and other users can use the script easily.

nkuehn:
@timonrey  sorry for not testing properly enough on the release notes script.  I switched the script to directly loading the implementation file and changed the file to CommonJS syntax so it can be run natively in node.js too. 

https://github.com/commercetools/commercetools-docs-kit/pull/986/commits/78325c807ce504c1b2743c67e6524d668eb3e26d

timonrey (approved):
Thanks 👍🏼

PR #987, opened 2021-06-14
Refactor API data hooks

nkuehn:
the pattern of accessing the type and resource static query hook databases is now the same and uses a local property key 
based index instead of iterating the full types list on every access.


nkuehn:
@timonrey  @davifantasia  learning concerning the sizes:

image

The types are "just" 2.6 MB of data to hydrate the app, but just(!) 115kb gzipped by MacOS // 111kb on Vercel gzipped.  The extreme repetition in the contents of the GraphQL compresses very good as expected.  If we manage to get that data to be loaded only when needed (aka on pages that use the API docs) that could even be acceptable.  

The lighthouse score did not change a lot in this PR - but this is just about loading the JSON data. The actual hook and components haven't really done anything because the smoke test site is not using them yet.  There could be some inefficiencies (stuff like iterating the thousands of types every time a given type is requested by name - to be found out once used)

nkuehn:
Issue to adress the root cause: https://github.com/commercetools/rmf-codegen/issues/232

.changeset/four-kids-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+"@commercetools-docs/gatsby-theme-api-docs": patch
+"@commercetools-website/api-docs-smoke-test": patch
```
timonrey, line 3: Don't we just want to publish changes from @commercetools-docs ?

nkuehn: yes, thank you.  We have an issue open to configure changeset better

timonrey (approved):
thanks, LGTM now

Issue #984, opened 2021-06-10
Audit Log - "Change Types" page loads very slowly


timonrey:
We've noticed that the "Change Types" page loads very slowly, probably because the page has to load the types which leads to a longer loading time. Still, this needs further investigation.

Link to the page:
https://docs.commercetools.com/api/history/change-types

nkuehn:
Short first analysis in the JS profiler of the Chrome dev tools:

![image](https://user-images.githubusercontent.com/4006139/122032241-bbf12700-cdcf-11eb-8992-1e792728eb16.png)

First research opportunities:  the useTypeLocations()  hook is 
    * either super slow in itself (a lot of recursions in it, we might have tangled up in these)
    * called way too often (move calling it up in the component tree and pass the data down?)

The Call tree view rather points towards the implementation inside the hook:

![image](https://user-images.githubusercontent.com/4006139/122033366-bea04c00-cdd0-11eb-95a9-46f81ff22c89.png)

nkuehn:
@timonrey you could start with not touching any code, but refactoring the useTypeLocations() hook code into more explicitly named functions, so that the actual hook is just doing the static query (which is fast) and then calls functions.  Obvious first candidate: `typeLocationOverrides` generation (that nested reducer).  Once you found the spot you try to optimize. 

My first instinctive candidate is that you have a double nested reducer that at the _leaf_  level does a repetitive spread merging of objects that are in fact databases that you just push a new key into.   That way you do it there with a spread is "fancy dancy modern JS syntax" but I think what it's really doing is that for every insertion of a new entry in the indexed database of type locations you are copying the complete database and the new entry into a new object. 

In "hot path" code that is inside often called functions and then in these in nested level of loops, try to avoid expressive powerful language constructs.  Think like a C developer and use the low level operations of the language.  If you want to add something to an object as a key,  you  use  `myObject[expressionThatHoldsNewKey] = newValueExpression` my **feeling** is that this code is overengineered: 

```js
{   // a fresh new object to be allocated
   ...locationList, // merge in all existing properties one by one -> done in every step so you end up with N lists in memory holding 1..2...3..4..n properties
   [`${api.api}__${location.type}`]: {  // a new property to the object whose key is a single-enty array and the value is an object with the actual data
      url: location.href,
   },
}
```
I think it could the result of starting with a reducer at all. 

!!!  All guesswork where to potentially start, not really tried out !!!

nkuehn:
[...]

PR #991, opened 2021-06-18
refactor: type location overrides

timonrey:
is part of #984


nkuehn (approved):
I was not able to measure the difference in the test data because they are not big enough, but it works and should be faster theoretically.  
Did you test with the production data or have been otherwise able to re-confirm that it adresses the problem? 

But even if not: we can merge this because it's more efficient and still find other improvements if it's not enough yet.

timonrey:
I also was not able to test it with big data, but I guess we can see the difference after we merge this. That's why I want to keep the issue open until we see a sufficient performance improvement.

PR #997, opened 2021-06-21
Enable loading API spec data more selectively

### LINKED ISSUE #994: Stop injecting ApiType and ApiEndpoint components into MDXProvider by default
Currently,  `gatsby-theme-api-docs` is injecting its data-laden components into the MDXProvider by default but this is adverse to the site performance because they end up in every single page of a site that uses the theme plugin. 

We concluded to not do this any more and rather have sites import the components into the respective pages that need them.

### LINKED ISSUE #995: Provide ApiType and ApiEndpoint components that take the api data as argument instead of api key
As concluded in the performance analysis, need can arise to optimize site-specific by not having the API components load all  data of all APIs unwantedly. 

For these cases, site specific faster code would need an internal component for APITypes and Apiendpoints that directly takes the hook data as a property  (or react via react context?).   Site specific components can then do the query themselves (using a query fragment provided by the kit and providing a custom StaticQueryHook ) and pass only the needed data-

nkuehn:
closes #995
closes #994 

This is a major version because it requires explicitly importing the API documentation components in MDX pages and changes the interface from theme add-ons to the websites. 

It separates the API docs components interface into a wrapper "by key" that is the default export to MDX and uses the default Gatsby StaticQueryHooks that load the complete API specs of the site.   The hooks themselves are refactored in that they provide the required query fragments separately for websites that want to build custom hooks to optimize site performance.


packages/gatsby-theme-api-docs/shortcodes.js
```diff
@@ -0,0 +1,4 @@
+// Export only components that must be injected into the MDX provider
```
nkuehn, line 1: Nothing exported here - this is the desired outcome of the PR.

packages/gatsby-theme-api-docs/src/components/resource-method.js
```diff
@@ -21,24 +20,22 @@ const ResourceMethod = ({ apiKey, resource, method, title }) => {
   }
 
   return (
-    <FullWidthContainer>
```
nkuehn, line 24: FullWidthContainer moved directly into the "Method" component so that the "Method" can directly serve as the "pass your own data" component without having to provide yet another wrapper

packages/gatsby-theme-api-docs/src/components/resource/method/method.js
```diff
@@ -58,69 +59,73 @@ const Method = ({
   });
 
   return (
-    <SpacingsStack scale="s">
```
nkuehn, line 61: A noisy looking change in GitHub that is in fact only wrapping the existing code into a `<FullWidthContainer>`

packages/gatsby-theme-api-docs/src/hooks/use-apis.js
```diff
@@ -26,4 +26,10 @@ const useApis = () => {
   return queryResult.allRamlApi.nodes;
 };
 
-export default useApis;
+export const useApiByKey = (apiKey) => {
```
nkuehn, line 29: refactored this hook to be structured like the other main hooks which expose helper functions and the actual hook together from one file, too.

packages/gatsby-theme-api-docs/src/hooks/use-type-locations.js
```diff
@@ -9,6 +9,8 @@ const buildPageSlug = (page) => {
 };
 
 const convertComponentInMdxToTypeLocations = (data) =>
+  // TODO refactor this inefficient access pattern into a plain loop
```
nkuehn, line 12: @timonrey do you think it makes sense to do this separately or here?  I vote for separately but soon, you have routine now in this refactor.

packages/gatsby-theme-docs/utils/bootstrap-theme-addons.js
```diff
@@ -17,9 +17,10 @@ const gitignorePath = path.join(addOnsFolderPath, '.gitignore');
 
 function bootstrapThemeAddOnsForMarkdownComponents(themeAddOnNames) {
   let addOnComponentsExport = `/* eslint-disable prettier/prettier */
+/* eslint import/no-anonymous-default-export: [2, {"allowObject": true}] */
 /* THIS IS AN AUTOGENERATED FILE, DO NOT EDIT DIRECTLY */`;
   themeAddOnNames.forEach((plugin, index) => {
-    addOnComponentsExport += `\nimport * as components${index} from '${plugin}';`;
+    addOnComponentsExport += `\nimport * as components${index} from '${plugin}/shortcodes';`;
```
nkuehn, line 23: important change

recommended.code-workspace
```diff
@@ -64,6 +61,7 @@
       "ms-vscode.vscode-typescript-tslint-plugin",
       "redhat.vscode-yaml",
       "silvenon.mdx",
+      "jounqin.vscode-mdx",
```
nkuehn, line 64: when working with imports it turned out that using the "normal" MDX editor (instead of forcing the markdown editor) with this one which esLints the MDX is more helpful.  Let's try whether it's overall better or not.   At least for the kit part I liked having more code-ish editors

nkuehn (commented):
explanatory comments

packages/gatsby-theme-docs/src/plugins/rehype-mdx-section.js
```diff
@@ -1,11 +1,6 @@
 const isHastHeading = require('hast-util-heading');
 
 const isHeading = (node) => {
-  if (node.type === 'jsx') {
```
nkuehn, line 4: This was obsolete because in the meanwhile we decided to always use manual MD headings for API docs

packages/gatsby-theme-api-docs/src/components/resource/method/method.js
emmenko: In case you didn't know, there is an option to hide whitespace changes. This then looks like this:

image

packages/gatsby-theme-api-docs/index.js
```diff
@@ -1,4 +1,9 @@
...
+export { default as ApiType } from './src/components/type/type-by-key';
 export { default as ApiEndpoint } from './src/components/resource-method';
-export { default as ApiEndpointsForResource } from './src/components/resource/resource';
+export { default as ApiEndpointsForResource } from './src/components/resource/resource-by-key';
+export * from './src/hooks/use-api-types';
+export * from './src/hooks/use-api-resources';
+export * from './src/hooks/use-apis';
+export * from './src/hooks/use-type-location-overrides';
+export * from './src/hooks/use-type-locations';
+export * from './src/hooks/use-read-resource-by-resource-path';
```
emmenko, line 9: What's the reason for exporting those?

.changeset/five-pants-jump.md
```diff
@@ -0,0 +1,17 @@
...
+---
+
+Refactored the provisioning of theme extensions' components into MDX to support components that have to be imported explicitly in MDX. Auto-provided components now have to be exported via `./shortcodes` by theme add-ons.
+
+The API documentation is not automatically injecting its components into MDX any more to allow sites with large APIs to optimize the site performance (time to interactive).
+
+To migrate, all MDX pages have to explicitly import the `ApiType` and `ApiEndpoint` components. It's recommended that sites create an intermediate `/shortcodes.js` file so that the MDX import lines only read
+
+```
+import { ApiType, ApiEndpoint } from "./shortcodes"
```
emmenko, line 16: Is there a reason for naming this `shortcodes.js`? I'm not sure what the name should imply to be honest.

If this is only meant to be used within MDX files, can we simply name it `mdx.js` or `components.js`?

websites/api-docs-smoke-test/shortcodes.js
```diff
@@ -0,0 +1,8 @@
+// by convention, provide additional components that are _intended_
+// to be used in MDX content but not part of the defaults provided
+// via the MDX provider
+export {
+  ApiType,
+  ApiEndpoint,
+  ApiEndpointsForResource,
+} from '@commercetools-docs/gatsby-theme-api-docs';
```
emmenko, line 8: Isn't this the exact same thing as before? 🤔 I'm not sure I fully get the reasoning behind having a `shortcodes.js` exports vs `index.js`.

packages/gatsby-theme-api-docs/src/components/type/type.js
```diff
@@ -87,6 +84,7 @@ const ApiType = (props) => {
 
 ApiType.propTypes = {
   apiKey: PropTypes.string.isRequired,
+  apiTypes: PropTypes.object.isRequired,
```
emmenko, line 87: Isn't this an array?

```suggestion
  apiTypes: PropTypes.array.isRequired,
```

Nit: even better is to specify the shape of the array

emmenko (commented):
Thanks for the improvements!

I left some questions as some things are still confusing to me. For instance, besides the explicit imports, I fail to understand how this helps improving the performance.
Maybe I'm missing something obvious but that's my feedback so far.

packages/gatsby-theme-api-docs/index.js
nkuehn: The intent is to be enable building site specific API components that use the same data.  Every additional hook that loads nearly-the-same-but-different data just adds to the page weight.   This set of hooks is a client-side data interface to the API specs, exporting it from the package root communicates that you should always use these instead of adding own additional static queries on the same data. 

There is no specific use case I want to build right now.   So removing them would not hurt right now.

.changeset/five-pants-jump.md
nkuehn: The MDX devs call  this kind of components shortcodes (e.g. https://mdxjs.com/blog/shortcodes ) - the name is derived from wordpress which many content authors know and which introduced the concept of being able to throw some sort of tag into the middle of a text document to render a more complex UI component.
https://wordpress.com/support/shortcodes/

websites/api-docs-smoke-test/shortcodes.js
nkuehn: This is in the websites folder and the name is chosen to make the import statement in the MDX pages as intuitive and logical as possible.  There was no index.js in the website before.  

Alternatives from MDX author perspective in the MDX file: 
```mdx
// no local re-export:   -> verbose and hard to memorize and understand for authors. If you want to override the component locally in the website you have to touch all pages. 
import { ApiType } from "@commercetools-docs/gatsyby-theme-api-docs"

// local re-export from an index.js:  -> not sure, shortest but the end looks like a meaningless sequence of special characters. Would it work at all? 
import { ApiType } from "/"

// using a meaningful conventional name as a symbolic token ("always write this" - no need to expose internals)
import { ApiType } from "/shortcodes"
```

I personally found the last most meaningful but still memorizable.   Question of taste maybe, we can ask the authors for their opinion.

nkuehn: @stmeissner  do you have an opinion?

packages/gatsby-theme-api-docs/src/components/resource-method.js
```diff
@@ -1,11 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { FullWidthContainer } from '@commercetools-docs/gatsby-theme-docs';
-import useReadResourceByResourcePath from '../hooks/use-read-resource-by-resource-path';
+import { useReadResourceByResourcePath } from '../hooks/use-read-resource-by-resource-path';
 import reportError from '../utils/report-error';
 import Method from './resource/method';
 
-const ResourceMethod = ({ apiKey, resource, method, title }) => {
+const ResourceMethodByKey = ({ apiKey, resource, method, title }) => {
```
davifantasia, line 7: Is `ByKey` referencing `apiKey`?

packages/gatsby-theme-api-docs/src/components/resource/resource-by-key.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  return <Resource resourceObj={resourceObj} />;
+};
+
+ResourceByKey.propTypes = {
+  apiKey: PropTypes.string.isRequired,
+  resource: PropTypes.string.isRequired,
+};
+export default ResourceByKey;
```
davifantasia, line 21: Please what is the benefit of wrapping with `< ResourceByKey />`?

packages/gatsby-theme-api-docs/src/components/type/type-by-key.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import ApiType from './type';
+import { useApiTypes } from '../../hooks/use-api-types';
+
+const ApiTypeByKey = (props) => {
+  const apiTypeData = useApiTypes();
+  return <ApiType apiTypes={apiTypeData} {...props} />;
+};
+
+export default ApiTypeByKey;
```
davifantasia, line 10: Same here - what is the benefit of wrapping with `<ApiTypeByKey />`?

.changeset/five-pants-jump.md
davifantasia: IMHO - if `shortcodes` is just a naming convention, I don't think we should adopt it. We generally expose modules using `index.js` across multiple projects, so sticking to that would be helpful.

packages/gatsby-theme-api-docs/src/components/resource-method.js
nkuehn: Yes.  Do you think it's worth changing the names to `..ByApiKey` ?

packages/gatsby-theme-api-docs/src/components/resource/resource-by-key.js
nkuehn: See description text of the PR.  (enable using the child component in a custom wrapper that uses a hook that loads less or different data)

nkuehn:
> IMHO - if shortcodes is just a naming convention, I don't think we should adopt it. We generally expose modules using index.js across multiple projects, so sticking to that would be helpful.

@davifantasia  see my response on the same question in nicola's review

.changeset/five-pants-jump.md
emmenko: Thanks for the explanation. I didn't know about that term. As long as it's understandable and clear to the authors (assuming they are familiar with the term), it's fine by me.

packages/gatsby-theme-api-docs/index.js
emmenko: I see. Ok then.

nkuehn:
> For instance, besides the explicit imports, I fail to understand how this helps improving the performance.

@emmenko  sorry, I forgot to address this fundamental question.  

We found that the Network-transfer impact of the JSON size is not as horrible as feared because the 2-3 MB of Json are so repetitive that they compress extremely well down to 110kb or so.   But it's processing work nevertheless and from an SEO perspective the time to interactive and page weight is still a factor that should not be handled sloppyly, especially because the size of the API surface will only grow and now shrink over time. 

So we concluded in the analysis ticket that we focus on (in that order)
 1. improving the actual application performance (which does have measurable and user facing issues) - first steps already done to address the most problematic parts)
 2. making sure that the home page and pages that do not actually use API data don't load them. For that we concluded users have to put an explicit import into the MDX so that gatsby's build system does not let all other pages load the data. (this PR)
 3. enable websites that are especially exposed to have custom wrapper components with custom data loading queries that only pull what they need (also this PR, this is the "by key" wrapping that allows a site to built a custom data hook that loads less data and call the UI component with that data)

That way we stick to the gatsbyJS architecture and are good enough for the future.

packages/gatsby-theme-docs/utils/bootstrap-theme-addons.js
emmenko: I'm starting to wonder if we really need this change. The point of the add-ons was that they export components that are automatically injected into the provider.

From what I understand here, we're trying to do the same thing except for the "api-docs" add-on, which is not an add-on anymore since the change here is to import the api components explicitly.

This all export `index.js` vs `shortcodes.js` is also a bit confusing and to me it feels like a symptom to the underlying problem.

Therefore, given that we already agreed that we will have explicit imports in the MDX files, how about we do this for all add-ons and not just the api-docs?

Advantages:
* Same behavior across MDX add-ons.
* No bloated MDX provider.
* Each website will explicitly define a `shortcodes.js` with components that should be imported from the different theme packages.
* Each theme package has normal exports via `index.js`.

This obviously means that there won't be any "add-ons" with automatic imports/overrides but "just normal packages".

Would that be an option to have things more consistent? @nkuehn

davifantasia (approved):
Thanks for the explanation @nkuehn

packages/gatsby-theme-api-docs/src/hooks/use-type-locations.js
nkuehn: separate issue exists, I removed this line

packages/gatsby-theme-docs/utils/bootstrap-theme-addons.js
nkuehn: You're right in that our understanding of an add-on is including the automatic injection of components into the provider.  But I don't see it as an exclusive criterion.  To me, an add-on theme is a gatsby theme that only works in conjunction with the base theme and is assuming all the infrastructure and UI framing of the base theme.    It could for example not add new MDX components but rather override some other part of the base theme through component shadowing and would still be an add on. 

I understand your concern about consistency but I do not think the focus on consistency is improving the product in this case.  
The majority of documentation websites is not using the API docs add on and the long tail of users does not interact with it, too.   I don't see a point in making the system harder to use in the majority case and debug "just" for consistency in the codebase.

.changeset/five-pants-jump.md
nkuehn: > IMHO - if `shortcodes` is just a naming convention, I don't think we should adopt it. We generally expose modules using `index.js` across multiple projects, so sticking to that would be helpful.

@davifantasia  as described in the PR description and other comments the whole point is to separate the automatically injected components from what the theme.   That's why I have to pick some naming convention and shortcodes is the one that MDX uses in their own docs.

packages/gatsby-theme-api-docs/src/components/type/type.js
nkuehn: thanks! Fixed that wrong definition. 

specifying the shape of the array would get pretty verbose, so I avoided that.  It#s the result of this query: https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/gatsby-theme-api-docs/src/hooks/use-api-types.js#L11

websites/api-docs-smoke-test/shortcodes.js
adinakleine: One is too specific, two is missing information. Option three is fine for me too.

nkuehn: Thank you, so we keep it as implemented now.

PR #1022, opened 2021-07-12
fix: Misleading Query Parameter Type Rendering

### LINKED ISSUE #1019: Misleading Query Parameter Type Rendering - 1
For query parameters like `where` the type is rendered as `Array of String`, that is, it is expected to be passed as `where=['string1', 'string2']` - this is incorrect. The correct usage would be `where=<string1>&where=<string2>`.

Example of incorrect usage - https://commercetools-docs-git-tk-customergroup-auto-commercetools.vercel.app/api/projects/customerGroups#query-customergroups

A proposed fix is to render `where` type as a 'String`, then autogenerate additional description for usage.

davifantasia:
Closes #1019 

See the query parameters at https://commercetools-docs-kit-git-ao-query-parameter-commercetools.vercel.app/api-docs-smoke-test/endpoints/methods#query-parameters-with-multiple-instances for rendered type and additional description. In this case `sort`, `where` and `/^var[.][a-zA-Z0-9]+$/` has the additional descriptions.

Screenshot 2021-07-12 at 15 20 33

### DIFF
--- a/packages/gatsby-theme-api-docs/src/components/resource/method/query-parameters.js
+++ b/packages/gatsby-theme-api-docs/src/components/resource/method/query-parameters.js
@@ -3,15 +3,9 @@ import Parameters from './parameters';
 
 function transformQueryParameterDescriptions(queryParameters) {
   return queryParameters.map((parameter) => {
-    if (
-      parameter.description &&
-      parameter.type === 'array' &&
-      parameter.items
-    ) {
-      return {
-        ...parameter,
-        description: `${parameter.description}\n\nThe parameter can be passed multiple times.`,
-      };
+    if (parameter.type === 'array' && parameter.items) {
+      parameter.additionalDescription =
+        'The parameter can be passed multiple times.';
     }
 
     return parameter;
--- a/packages/gatsby-theme-api-docs/src/components/resource/method/query-parameters.js
+++ b/packages/gatsby-theme-api-docs/src/components/resource/method/query-parameters.js
@@ -10,7 +10,7 @@ function transformQueryParameterDescriptions(queryParameters) {
     ) {
       return {
         ...parameter,
-        description: `${parameter.description}\n\nMultiple instances can be passed. E.g. \`${parameter.name}=<${parameter.items.type}>&${parameter.name}=<${parameter.items.type}>&${parameter.name}=<${parameter.items.type}>\``,
+        description: `${parameter.description}\n\nThe parameter can be passed multiple times.`,
       };
     }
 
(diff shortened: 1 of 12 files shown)


packages/gatsby-theme-api-docs/src/components/resource/method/query-parameters.js
```diff
@@ -0,0 +1,49 @@
...
+function transformQueryParameterDescriptions(queryParameters) {
+  return queryParameters.map((parameter) => {
+    if (
+      parameter.description &&
+      parameter.type === 'array' &&
+      parameter.items
+    ) {
+      return {
+        ...parameter,
+        description: `${parameter.description}\n\nMultiple instances can be passed. E.g. \`${parameter.name}=<${parameter.items.type}>&${parameter.name}=<${parameter.items.type}>&${parameter.name}=<${parameter.items.type}>\``,
```
stmeissner, line 13: I suggest to give the hint about multiple instances right in the beginning and then append the regular description afterwards. This way we can explain the semantics of using the parameter several times, like for `where` it is a conjunction, but for `sort` it leads to a multi-sort behavior. Not sure, I did the change correct in the below suggestion, untested 😉 

```suggestion
        description: `The parameter can be passed multiple times, e.g. \`${parameter.name}=<${parameter.items.type}>&${parameter.name}=<${parameter.items.type}>&${parameter.name}=<${parameter.items.type}>\`\n${parameter.description}\n`,
```

nkuehn: I prefer stefan's wording "can be passed multiple times" too. 

Concerning the example I am not sure whether it's a good idea or just more confusing (for example, are the `<` placeholders or a necessary part of the syntax?).  

Everywhere else in the documentation we are assuming knowledge on how to build a valid URL query string (not explaining for example that the first one is after a question mark and then separated by ampersands) so I am not sure why we need to given an example of how a query string looks like here.  

Unless you have already discussed why you need the generated example, the following would suffice for me: 
`*This parameter can be passed mulitple times.*\n${parameter.description}`

davifantasia:
@stmeissner @nkuehn The only additional description is now `The parameter can be passed multiple times.`.

I have a question about similar parameters without description - in this case an example is `expand`. Currently we're rendering no description, should this current implementation. remain or we just render the additional description.

Current implementation:

Screenshot 2021-07-13 at 10 44 03

nkuehn:
@davifantasia  the hint is independent of whether there is a description or not.  It's comparable to the tags that we render in the documentation of response representation types: 

image

From that perspective, implementing it via string wrestling on markdown level is not really elegant anyways.  You can even consider directly using the style that we already have for those extra tags and shorten the statement to "Can be passed multiple times".

stmeissner (approved):
LGTM 👍

davifantasia:
Replaced by #1023

PR #1023, opened 2021-07-13
fix: Misleading Query Parameter Type Rendering

### LINKED ISSUE #1019: Misleading Query Parameter Type Rendering - 1

davifantasia:
Closes #1019 

Replaces #1022 

See the query parameters at https://commercetools-docs-kit-git-ao-query-parameter-v2-commercetools.vercel.app/api-docs-smoke-test/endpoints/methods#query-parameters-with-multiple-instances for rendered type and additional description. In this case `sort`, `where` and `/^var[.][a-zA-Z0-9]+$/` has the additional descriptions.

### DIFF
--- a/.changeset/query-parameter.md
+++ b/.changeset/query-parameter.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-api-docs': patch
+---
+
+fix: query parameters of type array now renderd as string with additional auto-generated description of usage
--- a/packages/gatsby-theme-api-docs/.eslintrc.yml
+++ b/packages/gatsby-theme-api-docs/.eslintrc.yml
@@ -1,1 +1,2 @@
-{}
+rules:
+  'react/react-in-jsx-scope': 'off'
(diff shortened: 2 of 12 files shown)


nkuehn (changes requested):
I just scanned the code but it looks like you invested the (necessary) time to refactor the query param component code a bit - that's nice because this is not the only issue we have with it.  Strictly spoken it needs to support rendering all those othe constraints like max/min etc, too. 

Concerning the visual result:  Very nice, it works better than I thought. 

But there's one regression in the string literal rendering - some of them seem to be "double-wrapped" into code blocks now:

image

davifantasia:
> But there's one regression in the string literal rendering - some of them seem to be "double-wrapped" into code blocks now:
> 
> image

It looks like this is from the original text, the writer did not have markdown text in mind. It appears that is supposed to be a list:

https://github.com/commercetools/commercetools-docs-kit/blob/853f8915072f8153fd1efa266b06b66f0360717a/api-specs/api/traits/sortable.raml#L8-L19

PR #1035, opened 2021-07-19
fix: duplicate searchInput id

### LINKED ISSUE #1028: Duplicate 'searchInput" id on html element - 1
`searchInput` id is duplicated when `ChildSectionsNav` component is used more than once on the same page.

Location - https://github.com/commercetools/commercetools-docs-kit/blob/a9aae5da783d257a7e2cee75b8d21e37749b0274/packages/gatsby-theme-docs/src/components/child-sections-nav.js#L194

See example page - https://docs.commercetools.com/import-api/product-variant

davifantasia:
Closes #1028

### DIFF
--- a/.changeset/thirty-plants-double.md
+++ b/.changeset/thirty-plants-double.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-docs/gatsby-theme-docs": patch
+---
+
+fix: duplicate searchInput id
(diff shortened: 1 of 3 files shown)


timonrey (approved):
Nice! Thanks for adding a second childSectionNav as well 💯

PR #1037, opened 2021-07-19
fix: property name and beta flag spacing

davifantasia:
Closes #1008

PR #1043, opened 2021-07-21
fix: property name and beta flag spacing

### LINKED ISSUE #1008: API type: Spacing between property name and beta flag got lost - 1
image

There should be spacing between the type name and beta flag.  It must have accidentally disappeared as part of some refactoring or other change

davifantasia:
Closes #1008 

Replaces #1037

### DIFF
--- a/packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
+++ b/packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
@@ -3,6 +3,7 @@ import styled from '@emotion/styled';
 import { css } from '@emotion/react';
 import { designSystem, Markdown } from '@commercetools-docs/ui-kit';
 import SpacingsStack from '@commercetools-uikit/spacings-stack';
+import SpacingsInline from '@commercetools-uikit/spacings-inline';
 import { BetaFlag } from '@commercetools-docs/gatsby-theme-docs';
 import { typography } from '../../../../design-system';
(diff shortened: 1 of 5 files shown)


timonrey (approved):
👍🏼

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -35,33 +36,37 @@ const NameType = (props) => {
...
+                  /
+                </span>
+              </Markdown.InlineCode>
+            ) : (
+              <Markdown.InlineCode>{props.property.name}</Markdown.InlineCode>
+            )}
+            {props.property.required && <Required />}
+          </SpacingsInline>
+          {props.property.beta && (
+            <BetaWrapper>
```
nkuehn, line 65: Why is the beta flag not part of the inline spacing wrapper but outside?

nkuehn (commented):
@davifantasia to be honest I find the effect in the VRTs a bit suspicious.   The implementation seems to affect all layouts, even if there is no beta flag present at all.  

Are we really sure it was never correct before?

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
davifantasia: Thanks for catching that, I used an unnecessary extra `SpacingsInline`

PR #1044, opened 2021-07-21
Update RMF-Codegen to 0.1.10 (for RAML API docs)

nkuehn:
https://github.com/commercetools/rmf-codegen/releases/tag/0.1.10


timonrey:
Not sure why the snapshots show so many visual changes...is there a reason for that?

nkuehn:
@timonrey  the changes look like the artifacts the "fix" for the spacing between beta and the property name caused, too.  If percy is comparing to the wrong base screenshot from main it is seeing other changes too. 

I have updated this from main now, let's see if it still occurs.   

Did we merge the beta spacer thing without checking while the complete page layout shifts?

nkuehn:
@timonrey @davifantasia   my guess was right, I had to update the changes from main to get percy to work.  

If you approve please directly merge, too.  Thanks!

timonrey (approved):

PR #1050, opened 2021-07-23
retrieve rmf-codegen version from maven metadata

jenschude:
It's somewhat annoying to always update the docs-kit then update docs repo and then maybe the docs branch.

I added a function which reads the maven metadata of the codegen cli application, retrieves the latest version according to the rmfCodegenVersion based on the semver constraint. Then downloads this version and creates a symlink to the downloaded version. The symlink is then used as stable executable target for running the codegen

### DIFF
--- a/packages/rmf-codegen/.gitignore
+++ b/packages/rmf-codegen/.gitignore
@@ -1,1 +1,1 @@
-bin/rmf-codegen-*.jar
+bin/rmf-codegen*.jar
--- a/packages/rmf-codegen/package.json
+++ b/packages/rmf-codegen/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@commercetools-docs/rmf-codegen",
   "version": "12.4.0",
-  "rmfCodegenVersion": "0.1.10",
+  "rmfCodegenVersion": "^0.1.7",
   "description": "Provides RMF-Codegen to javascript projects",
   "license": "MIT",
   "publishConfig": {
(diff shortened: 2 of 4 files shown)


nkuehn (changes requested):
I don’t think it’s a good idea to make the rmf codegen version nondeterministic. It’s random then why a ci build differs from local, why one ci build differs from another etc. 

The proper way to do this is to automate the version upgrades and open an automated pr (or even automatically release the nodejs wrapper to npm)

ideally the npm package lives in the rmf codegen repo and is released in sync automatically 
docs kit  and docs would then consume it independently and we would not have that chain of updates any more

jenschude:
It takes a semantic version range. ^0.1.7 means to take a version greater equals 0.1.7 and smaller 0.2.0

nkuehn:
I do understand semantic version selections. 

But a feature relase like 0.1.10 -> 0.1.11 still does change the site's output so the regression tests will be red and have to be approved.  If it soleley relies on the timing of package caches and installation runs on CI infrastructure they will be red without understandable reason. 
Also, selectively rolling back if a feature version is accidentally not doing the right thing is not possible. 

We could have the package be built in a way that the version can be overridden / set "hard" in the  local `package.json`  of the project using the rmf-codegen.   But in practice that would not be better maintained because all the automated version upgrades through the renovate bot would not pick it up. 

If there is an autoreleased npm package for every rmf-codegen package,  it can have the exact matching version number and the version turnaround time would also be just a day because renovate would directly pick it up in the docs repo (we can configure the renovate frequency per package if we need to)

PR #998, opened 2021-06-22
Document soft linking urn

davifantasia:
Closes #993 

https://github.com/commercetools/commercetools-docs-kit/issues/1001 to be fixed in a different PR

PR #1055, opened 2021-07-27
chore: Document soft linking URN

### LINKED ISSUE #993: Document Linking in API specs through URN format
Teams have started migrating and want to use links in RAML that point to other types.  

But the ability to link through the URN format is entirely undocumented yet, not even a mention in the README or the smoke test sites.

davifantasia:
Replaces #998 

Closes #993


nkuehn (commented):
Please also document the notation for endpoints and resolve the writing style issues that block the tests. Thanks! 

Putting it on the smoke tests site is not really useful in their current setup.  Will become relevant if we build an actual documentation website.

nkuehn (approved):
thank you!

PR #1057, opened 2021-07-27
fix: issue with tabIndex for keyboard navigation

### LINKED ISSUE #1054: Keyboard accessibility not working
I am missing browser page navigation by the Pos1/End/PgUp/PgDn keys for a long time. I am used to these keys since they work on virtually every page in the browser.
Would you mind supporting/not disabling these keys to make navigation in ct documentation a more fluent experience?

emmenko:
Attempt to fix #1054

### DIFF
--- a/packages/gatsby-theme-docs/src/components/search-input.js
+++ b/packages/gatsby-theme-docs/src/components/search-input.js
@@ -77,6 +77,7 @@ const SearchInput = React.forwardRef((props, ref) => {
   const handleBlur = () => {
     if (props.isDisabled) return;
     setIsActive(false);
+    props.onClose();
   };
   React.useEffect(() => {
     if (isLoading) {
--- a/packages/gatsby-theme-docs/src/layouts/internals/layout-main.js
+++ b/packages/gatsby-theme-docs/src/layouts/internals/layout-main.js
@@ -26,7 +26,13 @@ const Container = styled.main`
 `;
 
 const LayoutMain = (props) => (
-  <Container preventScroll={props.preventScroll}>
+  <Container
+    preventScroll={props.preventScroll}
+    // This is important to make the main container the active element when the page loads.
+    // The user can then use the keyboard navigation for example to scroll up/down.
+    autoFocus
+    tabIndex={0}
+  >
     {props.isTopMenuOpen ? (
       // GIVEN that the viewport is equal or less than laptop (sidebar not visible)
       // AND the top menu has been opened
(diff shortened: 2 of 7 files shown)


packages/gatsby-theme-docs/src/components/search-input.js
```diff
@@ -77,6 +77,7 @@ const SearchInput = React.forwardRef((props, ref) => {
   const handleBlur = () => {
     if (props.isDisabled) return;
     setIsActive(false);
+    props.onClose();
```
emmenko, line 80: Unrelated to the original issue, but when using tab navigation for the search input, when the input blurs the search dialog should close as it's not the focused element anymore.

emmenko:
So testing the deploy preview on Chrome it seems to work. On FF and Safari not really, the tab navigation is also almost nonexistent compared to Chrome.

packages/gatsby-theme-docs/src/layouts/internals/layout-main.js
```diff
@@ -26,7 +26,13 @@ const Container = styled.main`
 `;
 
 const LayoutMain = (props) => (
-  <Container preventScroll={props.preventScroll}>
+  <Container
+    preventScroll={props.preventScroll}
+    // This is important to make the main container the active element when the page loads.
+    // The user can then use the keyboard navigation for example to scroll up/down.
+    autoFocus
```
nkuehn, line 33: Works in chrome but not firefox.  It seems one still has to call `node.focus()` in javascript to focus a specific node in firefox.

nkuehn (commented):
Thanks for the improvement!  

If we want consistent behavior it requires a little more work though.  The preview of this PR only works on the initial page load, once you navigate inside the SPA the focus is not happening any more (I guess that's the intended behavior of the HTML level `autoFocus` - evaluated on HTML level page load but not after arbitrary javascript executions)

Since Firefox requires a script-based  `focus()` call too we might want to evaluation that approach?

emmenko:
I think I found some useful information on how to address this in Gastby: https://www.gatsbyjs.com/blog/2020-02-10-accessible-client-side-routing-improvements/

Let me give it a try.

emmenko:
Ok, this seems to work better now. @nkuehn @jenschude would you mind doing a quick manual test in the deploy link? Thanks

jenschude:
It seems to be better now. At least I can now directly navigate inside the page with up/down after clicking a link.

nkuehn (approved):
nice, thank you!  It works now in the (Macos only) environments and use case that I have available.

PR #1062, opened 2021-07-29
Unify type rendering of query parameters to body types

nkuehn:
This change fixes the unintended inconsistency between how numeric types of query parameters are documented vs. body JSON types.   

Two variables have been renamed. One to accomodate for the helper function being generic now and one that was less specific than possible from the beginning.


packages/gatsby-transformer-raml/src/utils/parameters-to-array.js
```diff
@@ -1,13 +1,14 @@
...
       return {
-        ...value,
+        ...parameter,
         name: key,
 
-        // transforming number type to float is specific to commercetools
-        type: value.type === 'number' ? 'float' : value.type,
-        builtinType: value.type === 'number' ? 'float' : value.type,
+        type: generateType(parameter),
+        builtinType: generateType(parameter),
```
nkuehn, line 11: this is the actual functional change

timonrey (approved):
thanks, looks better now 👍🏼

PR #1064, opened 2021-08-02
fix: move site-title id to actual site title

### LINKED ISSUE #1053: Search dialog shows css props in result list unintentionally
We've noticed that the search dialog renders css properties as text in the search results on multiple places. This happens most of the time before a section of the result list ends.

Bildschirmfoto 2021-07-27 um 11 48 15

timonrey:
is part of #1053 

In the docsearch config, we point to the `#site-title` id in order to find the header sections for the result list. Since we implemented the icons next to the site titles, we should move the id only to the actual title and not together with the icon.

### DIFF
--- a/.changeset/tiny-pandas-chew.md
+++ b/.changeset/tiny-pandas-chew.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Fix css bug in search dialog.
--- a/.changeset/tiny-pandas-chew.md
+++ b/.changeset/tiny-pandas-chew.md
@@ -2,4 +2,4 @@
 '@commercetools-docs/gatsby-theme-docs': patch
 ---
 
-Fix css bug in search dialog.
+Use another element to select the site-title for algolia search. This should fix the issue with having wrong content in the search results.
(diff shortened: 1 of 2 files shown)


davifantasia (approved):
Thanks

.changeset/tiny-pandas-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Fix css bug in search dialog.
```
emmenko, line 5: Nit: can we provide a bit more information about this fix? For instance, how does "moving an ID" fixes a CSS bug? It's not really clear...

Thanks

emmenko:
@timonrey are you sure this is fixing the issue with the search index? Or is it just an attempt? In which case, the PR shouldn't close the original issue.

.changeset/tiny-pandas-chew.md
timonrey: I explained it in the PR description. But I can change the changelog text to something more specific if you want.

emmenko: Yeah in general I would prefer to also have the changesets with good enough information.

For instance, reading this "Fix css bug in search dialog." I would think it's a fix in the css styles (colors, borders, etc.) but it's actually 1) not a change in css styles and 2) something totally unrelated to css.

emmenko (approved):

PR #1067, opened 2021-08-03
chore: update unified to v10

davifantasia:


emmenko:
@davifantasia FYI: you usually don't need to open a separate PR to migrate things, you can simply push commits to the renovate branch. Then we have everything in one place.

PR #1076, opened 2021-08-06
fix: avoid page breaking inside for firefox

### LINKED ISSUE #1052: Firefox: Very long Type names wrap over to the next column in the ChildSectionNavigator
If the name of a representation gets too long it uses the next column. This happens if the representation is the last entry in a column. 
Find an example [here](https://github.com/commercetools/commercetools-docs/pull/1192#issuecomment-887347568).

Firefox renders:
image

Chrome renders:
image

timonrey:
closes #1052

### DIFF
--- a/.changeset/four-readers-swim.md
+++ b/.changeset/four-readers-swim.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Firefox fix: Avoid very long type names from wrapping over to the next column in the child section navigator.
--- a/packages/gatsby-theme-docs/src/components/child-sections-nav.js
+++ b/packages/gatsby-theme-docs/src/components/child-sections-nav.js
@@ -94,6 +94,7 @@ const Container = styled.div`
 const ColumnContainer = styled.div`
   break-inside: avoid-column;
   margin-bottom: ${designSystem.dimensions.spacings.s};
+  page-break-inside: avoid;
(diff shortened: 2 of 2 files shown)


davifantasia (changes requested):
The problem still remains, I just checked this link with firefox:

https://commercetools-docs-kit-git-tt-fix-firefox-c3bcc4-commercetools.vercel.app/docs-smoke-test/views/two-level-index-nav#l2-subsection-21-deep

Result:

Screenshot 2021-08-06 at 11 17 06

emmenko:
One other thing I've noticed. We're currently rendering a nested structure because of the heading hierarchy:

image

Do you know why do we need to represent a nested structure? Can't we just flatten out the nesting and render a flatten list of links? This would also help with this bug.

timonrey:
From my understanding, we wanted to have a nested structure so that we can show the hierarchy of the sections.
The fix with `page-break-inside: avoid` works for all use cases except nested structures. I tried a couple of other css props but nothing seems to work in all cases for Firefox. So I guess we can try to find another way or we think about not having a nested structure in the child section nav and just show a flatten list of entries.

nkuehn (approved):
Hi, as just discussed the child section navigation is currently not used for nested structures.   The nesting is there to make it transparent that it was used over nesting but the authors currently prefer to  not do it.

davifantasia (approved):

PR #1077, opened 2021-08-06
fix: add dependency "assert" to fix development

timonrey:
local development is not working at the moment because of the missing `assert` dependency.

### DIFF
--- a/.changeset/tough-rivers-count.md
+++ b/.changeset/tough-rivers-count.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Add assert dependency.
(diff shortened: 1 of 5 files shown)


davifantasia (changes requested):
Please add `assert` to the themes instead so we manage it from the themes only.

emmenko:
Where is this dependency being used? What is exactly the error?

timonrey:
Looks like webpack can't find the `assert` dependency.
Bildschirmfoto 2021-08-06 um 14 09 52

davifantasia (approved):
Thanks

emmenko:
To clarify: the `assert` dependency is used by `mdast-util-from-markdown`, which in turn is used by `remark-parse` which was updated in #1071  (see also https://github.com/commercetools/commercetools-docs-kit/pull/1071/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR15370-R15385)

So, as a quick fix we can explicitly declare the dependency `assert`. However, we should [report the bug](https://github.com/syntax-tree/mdast-util-from-markdown/issues) and remove the dependency `assert` again once it's fixed upstream.

timonrey:
Thanks for the clarification 💯

PR #1094, opened 2021-08-27
Update RMF codegen

jenschude:


nkuehn (approved):

packages/writing-style/styles/commercetools/Acronyms.yml
```diff
@@ -55,6 +55,7 @@ exceptions:
   - RAML
   - REST
   - RFC # "Request For Comment" would confuse readers even more.
+  - RMF # "Rest Modeling Framework"
```
nkuehn, line 58: Please remove this, the writing style is not for our own internal content but for user facing content.  In that context, "RMF" must be explained before use. 

@davifantasia  the root cause is that we are still validating release notes etc. with the writing style which does not make sense. 

Can be excluded with a special comment for now (or you remove "md" extensions from writing style checking right here)

packages/writing-style/styles/commercetools/DateFormat.yml
```diff
@@ -7,5 +7,5 @@ ignorecase: true
 level: error
 nonword: true
 tokens:
-  - '\d{1,2}(?:\.|\/)\d{1,2}(?:\.|\/)\d{2,4}'
+  - '[^v]\d{1,2}(?:\.|\/)\d{1,2}(?:\.|\/)\d{2,4}'
```
nkuehn, line 10: This is a similar case but can stay since it catches a case that can happen e.g. in SDK documetnation too.

nkuehn (changes requested):

nkuehn:
@davifantasia  can you please help excluding MD files or generally files outside the websites directory from the writing style checks? Thanks !

jenschude:
> @davifantasia can you please help excluding MD files or generally files outside the websites directory from the writing style checks? Thanks !

The linting may be done independently in a separate PR to get this one here merged

davifantasia (approved):

PR #1106, opened 2021-09-15
fix(deps): replace remark-react with rehype-react

davifantasia:
It is recommended to replace `remark-react` with `rehype-react`. See https://github.com/remarkjs/remark-react#remark-react.


davifantasia:
Current issue - the markdown fragment styling is messed up because the HTML rendered is wrapped in a div like

```html
<section>
<div>
<p>paragraph</p>
<p>paragraph 2</p>
</div>
</section>
```

But HTML is currently rendered as

```html
<section>
<p>paragraph 1</p>
<p>paragraph 2</p>
</section>
```

Strangely, when I inspect using the React dev tool components tab, the comoonents remain similar.

davifantasia:
@timonrey tagging you just for you to be aware of the changes here.

PR #1152, opened 2021-11-19
chore: migrate to Yarn v3

emmenko:
More info here: https://yarnpkg.com/getting-started/migration


.percy.js
```diff
@@ -11,6 +11,12 @@ module.exports = {
       /(.*)\/views\/beta\/(.*)\.html/,
     ],
     exclude: ['site-template/*'],
+    overrides: [
+      {
+        files: /(.*)\/components\/rss-feeds\/index\.html/,
+        waitForSelector: 'table[data-table-name="rss-feed"]',
```
nkuehn, line 17: Is this related to the yarn migration or rather a fix that just happened to end up here?

nkuehn (approved):
Thank you! I would not have been able to pull this off.

.percy.js
emmenko: A bit of both. I noticed it's the same error we get e.g. here https://github.com/commercetools/commercetools-docs-kit/pull/1145

PR #1155, opened 2021-11-24
Implement support for HEAD and PATCH methods

### LINKED ISSUE #1128: API docs theme should support HEAD methods also
We now have HTTP HEAD methods in our docs that seem to be not supported by the API docs theme.

When putting this API endpoint in the MDX file:

```
<ApiEndpoint
  apiKey="api"
  resource="/{projectKey}/products/{id}"
  method="HEAD"
/>
```

I'm getting following error:
Screenshot 2021-10-21 at 11 54 08

Although the respective RAML spec does exist for route `/{projectKey}/products/{id}`:

```
head:
  securedBy:
  - oauth_2_0:
      scopes:
      - "view_products:{projectKey}"
  displayName: Check Product exists by id
  description: |-
    Checks if product with given ID exists.
  responses:
    200:
      description: |-
        The Product exists
      body:
  (codeExamples):
    curl: |-
      curl -X HEAD https://api.{region}.commercetools.com/{projectKey}/products/{id} -i \
      --header 'Authorization: Bearer ${BEARER_TOKEN}'



```

timonrey:
This Pull request
- closes #1128 
- implements support for HEAD and PATCH methods in the API docs theme.
- adds test data in the `api-docs-smoke-test`
- uses the new color for HEAD methods


api-specs/api/api.raml
```diff
@@ -137,6 +137,19 @@ resourceTypes:
...
+      - errorable
+      - deprecatable200
+    description: |
+      The Endpoint is responding a limited set of information about settings and configuration of the project.
+    securedBy: [oauth_2_0: { scopes: ['manage_project:{projectKey}'] }]
+    responses:
+      200:
+        body:
+          example: !include examples/project.example.json
+          type: Project
```
emmenko, line 152: FYI: `HEAD` requests do not have a response body usually.

Not sure if we can omit this. cc @nkuehn

api-specs/api/api.raml
```diff
@@ -137,6 +137,19 @@ resourceTypes:
         body:
           example: !include examples/project.example.json
           type: Project
+  head:
+    displayName: Get project settings with head
```
emmenko, line 141:
```suggestion
    displayName: Check if project exists
```

api-specs/test/api.raml
```diff
@@ -76,6 +76,46 @@ uses:
           body:
             type: objects.ObjectTestType
             example: !include examples/object-test-type.json
+    head:
+      displayName: Get all resources
+      description: Query/retrieve (HEAD) a specific resource or resource(s). This should return a paged query result of _blah_.
```
emmenko, line 81: Nit: I think this should be rephrased. There is no data returned.

emmenko (commented):
Looks good from what I understand.

api-specs/test/api.raml
timonrey: What would you write instead?

api-specs/api/api.raml
nkuehn: I would keep it as a smoke test because although it's not just "usually" but clearly forbidden in the the HTTP spec of HEAD nothing prevents a RAML author from actually defining a body in the RAML / openAPI.

api-specs/test/api.raml
nkuehn: @timonrey  maybe an opportunity to learn what the HEAD HTTP request method actually is ;-) 

-> does never return a body, just the headers, hence the name.

api-specs/test/api.raml
```diff
@@ -76,6 +76,43 @@ uses:
           body:
             type: objects.ObjectTestType
             example: !include examples/object-test-type.json
+    head:
+      displayName: Get all resources
+      description: Query/retrieve (HEAD) specific data.
```
emmenko, line 81:
```suggestion
      displayName: Check for existence
      description: Check if at least one resource exists matching the query condition.
```

api-specs/api/api.raml
```diff
@@ -137,6 +137,19 @@ resourceTypes:
         body:
           example: !include examples/project.example.json
           type: Project
+  head:
+    displayName: Check if project exists
+    is:
+      - errorable
+      - deprecatable200
+    description: |
+      The Endpoint is responding a limited set of information about settings and configuration of the project.
```
emmenko, line 146:
```suggestion
    displayName: Check for project existence
    is:
      - errorable
      - deprecatable200
    description: Check if there is a project matching the given key.
```

emmenko (approved):

PR #1156, opened 2021-11-26
Document dataflow in gatsby-transformer-raml readme

timonrey:


.changeset/quick-owls-chew.md
```diff
@@ -0,0 +1,5 @@
+---
```
nkuehn, line 1: I would not do releases for README changes that are not critical like completely wrong statements or omissions that prevent being able to use it at all. They can go into the next release with other stuff.   ->  delete that file.

nkuehn (approved):
Please remove the changeset file, otherwise OK.   If you want to improve, adding a few links to the folders with the respective code could be helpful for readers

PR #1158, opened 2021-11-29
chore: use gatsby cache incremental builds

emmenko:
This PR correctly sets up Gatsby incremental builds, by properly caching the `{.cache,public}` folders. It uses a custom scripts (basically the same as in the docs repo) to namespace each website workspace artifacts.

Observed behavior: build time faster up to 50-70%.


nkuehn (approved):
thanks!

timonrey (approved):
Nice! 🙌🏼

PR #1160, opened 2021-11-30
Update and Improve Algolia Docsearch Setup

nkuehn:
This does:
 * improve the DOM to be able to actually crawl the generated API docs properly  (they are not at the moment, just the descriptions but not the property names etc.),   There can be done more (e.g. method paths) but this is a minimal start
 * change to the new algolia index and api key
 * remove the user research banner from the index

This does NOT yet (to be done separately)
 * update to the new algolia docsearch frontend component (would need new styling from scratch)

**status**:  crawler query sufficiently stable now - final optimization can only be done once the DOM changes are deployed on the production docs and the crawler has run over them once.  For example, the change types has too many changes but this should be resolved with the `data-search-key` annotation de-duplicating paragraphs inside API docs tables.


packages/gatsby-theme-api-docs/src/components/description.js
```diff
@@ -17,7 +17,7 @@ const DescriptionTextContainer = styled.span`
 // as a property. Use this to embed description into a layout like
 // a table, card or bullet list.
 export const DescriptionText = (props) => (
-  <DescriptionTextContainer>
+  <DescriptionTextContainer className="embedded-api-description">
```
emmenko, line 20: Let's avoid using classnames as references. We can use e.g. `data-search-key` attributes and use the query selector like `[data-search-key="embedded-api-description"]`

packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
```diff
@@ -69,25 +74,26 @@ function ParameterRow(props) {
     isParameter: true,
   });
   return (
-    <tr key={props.parameter.name}>
+    <tr
+      key={props.parameter.name}
+      className="api-type-property" /* class name for the search crawler */
```
emmenko, line 79: Same here

```suggestion
      data-search-key="api-type-property"
```

packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
```diff
@@ -69,25 +74,26 @@ function ParameterRow(props) {
...
           <SpacingsInline scale="xs">
             <Markdown.InlineCode>{props.parameter.name}</Markdown.InlineCode>
             {props.parameter.required && <Required />}
           </SpacingsInline>
         </PropertyName>
-        <PropertyType className="name name-type">
-          {typeToRender.displayPrefix && (
-            <span className="name">{typeToRender.displayPrefix}</span>
-          )}
+        {' ' /* whitespace for the search crawler */}
```
emmenko, line 88: Nit: wrap this into a `<span>`?

```suggestion
        <span>{' ' /* whitespace for the search crawler */}</span>
```

packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
```diff
@@ -69,25 +74,26 @@ function ParameterRow(props) {
...
-          {typeof typeToRender.type === 'string' ? (
-            <span className="name">{typeToRender.type}</span>
-          ) : (
-            typeToRender.type
-          )}
+          {typeof typeToRender.type === 'string'
+            ? typeToRender.type
+            : typeToRender.type}
         </PropertyType>
+        {' ' /* whitespace for the search crawler */}
```
emmenko, line 96:
```suggestion
        <span>{' ' /* whitespace for the search crawler */}</span>
```

packages/gatsby-theme-api-docs/src/components/type/properties/rows/description.js
```diff
@@ -135,6 +135,7 @@ const Description = (props) => {
   return (
     <SpacingsStack scale="s">
       <DescriptionText markdownString={props.property.description} />
+      {' ' /* whitespace for the search crawler */}
```
emmenko, line 138:
```suggestion
      <span>{' ' /* whitespace for the search crawler */}</span>
```

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -35,7 +39,7 @@ const NameType = (props) => {
 
   return (
     <SpacingsStack scale="xs">
-      <PropertyName className="name-type">
```
emmenko, line 38: Thanks

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -66,20 +70,17 @@ const NameType = (props) => {
...
-        {isRegex(props.property.name) ? (
-          <span className="name">
-            Any property matching regular expression above
-          </span>
-        ) : typeof typeToRender.type === 'string' ? (
-          <span className="name">{typeToRender.type}</span>
-        ) : (
-          typeToRender.type
-        )}
+      {' ' /* whitespace for the search crawler */}
```
emmenko, line 73:
```suggestion
      <span>{' ' /* whitespace for the search crawler */}</span>
```

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -66,20 +70,17 @@ const NameType = (props) => {
...
+      <PropertyType>
+        {typeToRender.displayPrefix && typeToRender.displayPrefix}
+
+        {isRegex(props.property.name)
+          ? 'Any property matching regular expression above'
+          : typeof typeToRender.type === 'string'
+          ? typeToRender.type
+          : typeToRender.type}
       </PropertyType>
+      {' ' /* whitespace for the search crawler */}
```
emmenko, line 83:
```suggestion
      <span>{' ' /* whitespace for the search crawler */}</span>
```

packages/gatsby-theme-api-docs/src/components/type/properties/rows/rows.js
```diff
@@ -13,7 +13,10 @@ const Rows = (props) => {
       return undefined;
 
     return (
-      <tr key={property.name}>
+      <tr
+        key={property.name}
+        className="api-type-property" /* class name for the search crawler */
```
emmenko, line 18:
```suggestion
        data-search-key="api-type-property"
```

packages/gatsby-theme-api-docs/src/components/description.js
```diff
@@ -28,8 +28,8 @@ export const DescriptionParagraph = (props) => {
   return (
     <DescriptionParagraphContainer>
       {typeof props.children === 'string' ? (
-        <Markdown.TypographyContainer>
-          <DescriptionText markdownString={props.children} />
+        <Markdown.TypographyContainer className="body-text-api-description">
```
emmenko, line 31:
```suggestion
        <Markdown.TypographyContainer data-search-key="body-text-api-description">
```

algolia-crawler.js
```diff
@@ -0,0 +1,117 @@
+// This is a hardcopy of the Algolia crawler configuration for reference
+// The live version is authored here:
+// https://crawler.algolia.com/admin/crawlers/1b5189dd-f93c-43b7-8b45-e5ec6b0d319d/configuration/edit
+/* eslint-disable */
+
+new Crawler({
```
emmenko, line 6: So this would be used in a follow up?

emmenko (commented):
Quick code review (haven't looked at the UI changes yet)

packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
nkuehn: Why would I want that overhead?  HTML is cleanly specified how to treat whitespace and I don't need more than the effect of one piece of whitespace.

algolia-crawler.js
nkuehn: This is the server side code running on the algolia crawlers, this file is here for documentation and properly tracking changes we do.  If we accidentally delete the algolia instance we'd have to recreate this from scratch if it's not documented. 

To make clearer that it's actually docs and not code we could move it into a `/docs` folder or make it a big code block in a markdown file.  The "IDE" in the algolia backend is better for editing anyways.

algolia-crawler.js
```diff
@@ -0,0 +1,134 @@
...
+  exclusionPatterns: [
+    '**/releases',
+    '**/releases/**',
+    '**.html',
+    '**/**docs/removed-functionality**',
+    '**/**docs/removed-functionality**/**',
+  ],
+  ignoreCanonicalTo: false,
+  discoveryPatterns: ['https://docs.commercetools.com/**'],
+  schedule: 'at 22:00 every weekday',
```
timonrey, line 22: Shouldn't it be `every day`?

nkuehn: We're not working on weekends,  weekdays are enough.  We get this free from Algolia and I think we should try to use the service carefully in exchange.

emmenko (approved):

Issue #1061, opened 2021-07-29
Render a dash only if there is no content in the description


TohruKohrita:
This is an example from a query parameter doc.
<img width="780" alt="Screenshot 2021-07-29 at 16 27 45" src="https://user-images.githubusercontent.com/63338357/127510404-ce5e2526-90a8-4a5d-baf2-e814a2d2a31c.png">

We want the dash only if the description is completely empty, i.e. **no text and no blue box**.

nkuehn:
For clarification:  The description english grammar caused some confusion because it was ambiguous.  
> We want the dash only if the description is completely empty
"only if"  vs.  "dash only"  the reference of "only"? 

from context it's clear.   Let's just not render that dash at all any more.

PR #1166, opened 2021-12-01
Don't render a dash when parameter has no description

timonrey:
closes: #1061

### DIFF
--- a/.changeset/good-dingos-search.md
+++ b/.changeset/good-dingos-search.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-api-docs': patch
+---
+
+Query parameters without a description are rendered with a dash only.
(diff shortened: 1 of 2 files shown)


.changeset/good-dingos-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-api-docs': patch
+---
+
+Query parameters without a description are rendered with a dash only.
```
nkuehn, line 5:
```suggestion
API type properties without descriptions render without the dash to avoid unnecessary whitespace
```

nkuehn (commented):
for consitency the type layout should also not use the dash any more, not just the endpoint parameters

timonrey:
afaik the dash is no longer rendered anywhere

nkuehn:
> afaik the dash is no longer rendered anywhere

correct, I confused something

nkuehn (approved):

PR #1175, opened 2021-12-13
test: move link check into separate check

nkuehn:
Analogous to what we did in the docs content repo.  
Makes it more transparent whether a build really failed or whether some external site or content is broken. 

Created because I saw that one of the deps update PRs was red, but then found it's because we temporarily removed one of our marketing webpages.


.github/workflows/after-deployment.yaml
```diff
@@ -0,0 +1,30 @@
...
+  fetch-vercel-logs:
+    name: 'Vercel Build Log'
+    if: github.event.deployment_status.state != 'pending'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v2
+      - name: Install Vercel CLI
+        run: npm i --silent --no-progress vercel
+      - name: CLICK HERE TO SEE THE VERCEL BUILD LOG
+        run: FORCE_COLOR=true vercel logs -n 10000 --scope commercetools --token="${{ secrets.VERCEL_TOKEN }}" ${{ github.event.deployment_status.target_url }}
```
emmenko, line 15: This is not needed here

scripts/build.sh
```diff
@@ -19,4 +19,5 @@ cp -f ./websites/index.html ./public
 
 echo "Running link checker"
 
-yarn check-links
+# run the test but don't fail the otherwise successful build (separate check exists) -> || true
+yarn check-links || true
```
emmenko, line 23: 🤔 what's the point of having it here if we have a check anyway?

nkuehn: laziness to think about it  :-(.  Can be removed

.github/workflows/after-deployment.yaml
nkuehn: I would like to ease contributions without having to make everybody a vercel user, also in the kit.  Just one small pice in being more open for contributions.  But no emotions attached.

emmenko: I see, but I would reconsider this if we get requests/feedback from potential contributors.

For now we all have access to the logs, so I don't see a need to have this job in the docs-kit.

websites/docs-smoke-test/src/content/index.mdx
```diff
@@ -2,19 +2,36 @@
...
 
-# Narrow Cards
+# Normal cards
 
-<Cards narrow>
+<Cards>
   <Card>Marvel's The Avengers</Card>
   <Card>Avengers: Age of Ultron</Card>
   <Card>Avengers: Infinity War</Card>
-  <Card>Avengers: Endgame</Card>
```
timonrey, line 19: So Endgame did not make it anymore 😄

timonrey (approved):
LGTM

PR #1177, opened 2021-12-14
Implement formatting for regex in query parameters

### LINKED ISSUE #1009: RAML regex pattern properties: not formatted in Query Parameters
Screen Shot 2021-07-06 at 11 09 05

Expected result is that it's rendered like in a payload representation type:
https://commercetools-docs-kit.vercel.app/api-docs-smoke-test/types/general#regexpropertytesttype

Here's the existing docs kit test case:  https://commercetools-docs-kit.vercel.app/api-docs-smoke-test/endpoints/methods#get-search-resource  (not formatted yet). 

----
second issue in the same lines of code

The implementation has a bug but we oversaw it when initially merging the feature:  The data type of the respective field or parameter is lost,  the text "Any property matching regular expression above" is _replacing_ the  type information ("String" / "Integer" / etc.) instead of being added to it. 

This poses a little visual problem:  How to fit that sentence **and** the Type?   A solution that would not require a design change would be to make the text include the type.  For example:  "Any String property matching regular expression above"  //  "Any String parameter matching the regular expression above".  

Or, a more concise wording:  "Any String property matching this regular expression"  ("above" would be misleading in the query parameter layout anyways because it's next to  it, not below). 

----

plus a third issue: The text has a grammar issue, it's missing a "the":   "Any parameter matching **the** regular expression above"

timonrey:
closes #1009

### DIFF
--- a/packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
+++ b/packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
+import { css } from '@emotion/react';
 import {
   Markdown,
   designSystem as uiKitDesignSystem,
@@ -13,6 +14,13 @@ import Title from './title';
 import { DescriptionText } from '../../description';
 import Info from '../../info';
 import { typography } from '../../../design-system';
+
+const isRegex = (string) =>
+  string.charAt(0) === '/' && string.charAt(string.length - 1) === '/';
+
+const getExpressionInsideSlashes = (input) => {
+  return input.match(/\/([^;]*)\//);
+};
 
 // inline-blocks inside a block are wrapped first before wrapping inline.
 // this implements a wrapping behavior where property name and type are separated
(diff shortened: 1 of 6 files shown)


packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
```diff
@@ -14,6 +15,13 @@ import { DescriptionText } from '../../description';
 import Info from '../../info';
 import { typography } from '../../../design-system';
 
+const isRegex = (string) =>
+  string.charAt(0) === '/' && string.charAt(string.length - 1) === '/';
+
+const getExpressionInsideSlashes = (input) => {
+  return input.match(/\/([^;]*)\//);
+};
```
nkuehn, line 23: I think you should not duplicate the code but re-use it.   Either from a generic utils package or by exporting them from the type component.

packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
```diff
@@ -77,15 +85,37 @@ function ParameterRow(props) {
...
+                </span>
+                {getExpressionInsideSlashes(props.parameter.name)[1]}
+                <span
+                  css={css`
+                    color: ${uiKitDesignSystem.colors.light.textInfo};
+                  `}
+                >
+                  /
+                </span>
+              </Markdown.InlineCode>
```
nkuehn, line 105: Same here, this is 100% duplicate.  Did you consider making it a component that can be used by both?

packages/gatsby-theme-api-docs/src/utils/render-regex-property.js
```diff
@@ -0,0 +1,28 @@
+import { css } from '@emotion/react';
+import { Markdown, designSystem } from '@commercetools-docs/ui-kit';
+
+function renderRegexProperties(data) {
```
nkuehn, line 4: I'm not the react conventions guru here but to my understanding this is not a utility but a component, i.e. would rather live somewhere in the components folder and would rather have an export of the component name instad  of a `renderFooBar` naming

nkuehn (approved):
great!

PR #1187, opened 2022-01-10
Proposal for different rendering of enums

nkuehn:
This is an unrequested feature I tried as a kind of tech time to have at least written 20 lines of code in 2022. 

There was some unhappiness about the way API Enums are rendered as a definition list.  @adinakleine  and me concluded to stick with the definition list but change the key layout to String literals ("code"). 

[Preview (New)](https://commercetools-docs-kit-git-nk-enums-as-table-commercetools.vercel.app/api-docs-smoke-test/types/enums-constants-unions)
[Current Form (Old)](https://commercetools-docs-kit.vercel.app/api-docs-smoke-test/types/enums-constants-unions)


nkuehn:
discussion with @adinakleine : Let's have something that does not differ between if it has descriptions or not.  Let's use the definition list but have the term as a string literal.   More similar to the existing setup.

timonrey (approved):
I like it 🙂 looks good from the technical side

PR #1190, opened 2022-01-17
Implement tag ordering for RAML types

### LINKED ISSUE #1102: Reorder `maximum`, `minimum` tags
When tags like `minLength`, `maxLength`, `minimum`, `maximum` are modeled in RAML, the max value appears before the min value. 

eg1
eg2

**_We would want the min value to appear before the max value._**

timonrey:
closes #1102


packages/gatsby-theme-api-docs/src/utils/extract-additional-info.js
```diff
@@ -43,9 +29,24 @@ const extractAdditionalInfo = (property) => {
...
   });
 
-  additionalInfo = reorderFields(additionalInfo, ['default']);
+  const tagIdentifier = ['max', 'min', 'default'];
+  let sortedList = Object.entries(additionalInfo);
+
+  tagIdentifier.forEach((tag) => {
+    sortedList = sortedList.reduce((list, item) => {
+      if (item[0].includes(tag)) {
+        return [item, ...list];
```
nkuehn, line 38: nitpick that using array push / pop is faster because you do not clone the complete array.  But we're dealing with really small amounts of data here so don't worry.

nkuehn (commented):
@timonrey  A most general feedback:   Objects are not ordered, it's just an implementation coincidence that you rely on them being ordered here. 

Did you consider switching the data structure passed to the component to an array of objects with name / value pairs?   This would make it reliable and could also make the rendering code easier.

timonrey:
Why do you think that the objects are not ordered? I've added a couple of tests now and all scenarios seem to work.

packages/gatsby-theme-api-docs/src/utils/extract-additional-info.spec.js
```diff
@@ -0,0 +1,66 @@
...
+      minLength: 1,
+      pattern: 2,
+      maxLength: 1,
+    },
+  ];
+}
+
+function expectedTestResults() {
+  return [
+    {
```
nkuehn, line 49: you're comparing object keys here.  An object is identical no matter how it's ordered.  Javascript objects are not ordered.  You need to migrate the data structure to an array (see my previous review). 
Stringifying in the middle makes it even less predictable.  

Example array structure:
```js
const currentlyNoOrderInformation = {
  maxItems: 1,
  minItems: 2
}  // if you now change the order, it's still the same object

const orderedInformationExample = [
  { name: "maxItems", value: 1},
  { name: "minItems", value: 2}
]  // if you now change the order, it's not the same array any more

```

JSON spec (since you're serializing and deserializing here): https://www.rfc-editor.org/rfc/rfc7159.html#section-1  
> An object is an unordered collection of zero or more name/value
   pairs, where a name is a string and a value is a string, number,
   boolean, null, object, or array.

nkuehn: @timonrey  PS for more exact precision:
 * since ES 2015 _string_ keys in objects are retained in their _insertion_ order.   
 * JSON never allowed assumptions about object key ordering

In practice even JSON keeps the order in most implementations but again, it's a bit of a guess. 

What I wanted to primarily say: It's a bit of guesswork to find the root cause if you are not 100% sure the language features are used correctly.  You could start by removing the back and forth via JSON but then you would still have the unclear situation whether the `equals()` deep comparison is checking object order or not.  I think it shouldn't.

nkuehn (approved):

PR #1192, opened 2022-01-17
Migrate Builds to GitHub Actions

nkuehn:
> Status:   It works in general (on a separate vercel setup to test) but we need an transition phase where vercell still runs the dependency installation.  Issues: 
> * yarn dependency caching seems to not yet be very effective.  install step still takes 2-3 minutes. 

---

This PR migrates the production build step to GitHub actions. It's driven by Vercel's changes in in fee structure which are not favorable any more to use it as a build system.  After all it's consequential, vercel is optimized for the runtime and delivery part and the plan accounts for heavy use of functions, traffic etc..  

Advantages:
-  Simpler billing and capacity structure. Actions are paid by runtime minutes without further constraints like maximum parellelity.  And the runtime minutes have no upper boundary towards enterprise plans
-  Access to the build logs for any GitHub user without a vercel account or custom scripts that pull the logs
-  expressivity to customize and optimize further

Disadvantages:
- in the Vercel UI it's likely not as transparent any more what deployment belongs to what PR / commit
- Github Actions have no proper way of purging the cache manually if something is stuck and the cache prevents further builds from succeeding.
- Builds on github actions are less predictable in how long they run, e.g. I am seeing 14 minutes builds in GH which take 4-5 mins on vercel but sometimes just 3 minutes)

This is _not_ attempting parallelization or refactoring of the actions structure.  These goals can be addressed in separate changes.  The main purpose here is to migrate and use it in the kit development workflow for a few weeks before applying the pattern to the content repositories. 

It's using two specialized Github Actions to emulate the previous behavior: 
 * https://github.com/amondnet/vercel-action 
 * https://github.com/marketplace/actions/github-deployments  (adds explicit deployments in GitHub. Should be a feature of the vercel-action but that PR was never merged.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,13 +22,13 @@ jobs:
       # avoid warnings like "tput: No value for $TERM and no -T specified"
       TERM: xterm
       # default to being a preview deployment
-      DEPLOY_ENV: preview
+      DEPLOY_ENV: Preview
       VERCEL_ARGS: ''
 
     steps:
       - name: Determine Deploy Environment
         if: ${{ env.GITHUB_REF_NAME == 'main' }}
-        run: echo "DEPLOY_ENV=production" >> $GITHUB_ENV && echo "VERCEL_ARGS=--prod" >> $GITHUB_ENV
+        run: echo "DEPLOY_ENV=Production" >> $GITHUB_ENV && echo "VERCEL_ARGS=--prod" >> $GITHUB_ENV
 
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,8 +38,6 @@ jobs:
             .cache
             public
           key: ${{ runner.os }}-build-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-build-
 
       - name: Set custom Yarn cache path
         run: yarn config set cacheFolder .cache/yarn
@@ -73,11 +71,13 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           # vercel-args: '--prod'  # TODO only of on main branch
           vercel-org-id: ${{ secrets.VERCEL_ORG_ID}}
+          scope: commercetools
           vercel-project-id: ${{ secrets.VERCEL_PROJECT_ID}}
-          alias-domains: | #Optional
-            docs-kit-sha-${{ github.sha }}-commercetools.vercel.app
-            docs-kit-pr-{{PR_NUMBER}}-commercetools.vercel.app
-            docs-kit-branch-{{BRANCH}}-commercetools.vercel.app
+          # there are length restrictions to domains and only *.commercetools.vercel.app works
+          # reliably for generated subdomains. No branch name domain because branch names can be long.
+          alias-domains: |
+            docskit-sha-${{ github.sha }}.commercetools.vercel.app
+            docskit-pr-{{PR_NUMBER}}.commercetools.vercel.app
 
       - name: Update Deployment Status
         uses: bobheadxi/deployments@v0.6.2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,10 +1,8 @@
 name: Main workflow
-
 # The event triggers are configured as following:
 # - on branch main, trigger the workflow on every push to trigger production checks and deployment
 # - on any pull request, trigger the workflow
 # This is to avoid running the workflow twice on pull requests.
-
 on:
   push:
     branches:
@@ -23,12 +21,11 @@ jobs:
       TERM: xterm
       # default to being a preview deployment
       DEPLOY_ENV: Preview
-      VERCEL_ARGS: ''
 
     steps:
       - name: Detect Production Deploy
         if: ${{ env.GITHUB_REF_NAME == 'main' }}
-        run: echo "DEPLOY_ENV=Production" >> $GITHUB_ENV && echo "VERCEL_ARGS=--prod" >> $GITHUB_ENV
+        run: echo "DEPLOY_ENV=Production" >> $GITHUB_ENV
 
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -1,17 +1,19 @@
 name: Main workflow
 
 # The event triggers are configured as following:
-# - on branch main, trigger the workflow on every push
+# - on branch main, trigger the workflow on every push to trigger production checks and deployment
 # - on any pull request, trigger the workflow
 # This is to avoid running the workflow twice on pull requests.
+
+# TODO separate production and PR deployments (e.g. main does not need to run the tests after build and deployment)
```
emmenko, line 8: I think we can simply skip some steps if e.g. branch is not `main`

.github/workflows/main.yml
```diff
@@ -35,9 +37,9 @@ jobs:
           path: |
             .cache
             public
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-build-${{ hashFiles('**/yarn.lock') }}
```
emmenko: Why did you change this? To bump the cache?

If so, I would suggest to use a "counter" like `-v1-`. Then every time you want to bump the cache, you bump the number.

.github/workflows/main.yml
```diff
@@ -54,9 +56,45 @@ jobs:
         env:
           ENABLE_SEARCH: 'true'
 
+      - name: Initialize Deployment
+        uses: bobheadxi/deployments@v0.6.2
+        id: start-deployment
+        with:
+          step: start
+          token: ${{ secrets.GITHUB_TOKEN }}
+          env: ${{ github.ref_name }}
```
emmenko, line 63: Shouldn't this be set to `preview` (for PRs) and `production` (for `main` branch)?

.github/workflows/main.yml
```diff
@@ -54,9 +56,45 @@ jobs:
...
+
+      - name: Deploy to Vercel
+        id: deploy-to-vercel
+        uses: amondnet/vercel-action@v20
+        with:
+          vercel-project-name: commercetools-docs-kit
+          vercel-token: ${{ secrets.VERCEL_TOKEN }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # vercel-args: '--prod'  # TODO only of on main branch
```
emmenko, line 72: Just so that we don't forget.

.github/workflows/main.yml
```diff
@@ -54,9 +56,45 @@ jobs:
...
+          vercel-org-id: ${{ secrets.VERCEL_ORG_ID}}
+          vercel-project-id: ${{ secrets.VERCEL_PROJECT_ID}}
+          alias-domains: | #Optional
+            docs-kit-sha-${{ github.sha }}-commercetools.vercel.app
+            docs-kit-pr-{{PR_NUMBER}}-commercetools.vercel.app
+            docs-kit-branch-{{BRANCH}}-commercetools.vercel.app
```
emmenko: Do we need all of these? 🤔

.github/workflows/main.yml
```diff
@@ -54,9 +56,45 @@ jobs:
...
+          status: ${{ job.status }}
+          env_url: ${{ steps.deploy-to-vercel.outputs.preview-url }}
+          deployment_id: ${{ steps.start-deployment.outputs.deployment_id }}
+
+      # TODO move this into a separate workflow that holds the static checks only? They can run independently of builds
       - name: Lint text files
         run: yarn run jest --projects jest.{eslint,stylelint,test,text}.config.js --maxWorkers=3 --reporters jest-silent-reporter
 
+      # Tests run after deployment to make sure a preview URL is available for every build
+      # TODO maybe just stop the job here when on the main branch?
```
emmenko, line 97: If we decide to deploy first and test after, we can also think of running the tests against the preview URL, instead of localhost. This includes:

* e2e tests
* link checker

vercel.json
```diff
@@ -1,8 +1,9 @@
 {
   "version": 2,
   "public": true,
-  "name": "commercetools-docs-kit",
-  "alias": "commercetools-docs-kit",
+  "github": {
+    "enabled": false
```
emmenko, line 5: 💯

.github/workflows/main.yml
```diff
@@ -54,9 +54,47 @@ jobs:
...
+        with:
+          step: finish
+          token: ${{ secrets.GITHUB_TOKEN }}
+          status: ${{ job.status }}
+          env_url: ${{ steps.deploy-to-vercel.outputs.preview-url }}
+          deployment_id: ${{ steps.start-deployment.outputs.deployment_id }}
+
+      # TODO move this into a separate workflow that holds the static checks only? They can run independently of builds
       - name: Lint text files
         run: yarn run jest --projects jest.{eslint,stylelint,test,text}.config.js --maxWorkers=3 --reporters jest-silent-reporter
```
emmenko, line 94: I would still run this before we deploy, to avoid unnecessary deployments if something does not work.

emmenko (commented):
Nice, I like where this is going 💯

.github/workflows/main.yml
nkuehn: No, this is already changed to only SHA and branch. I needed to experiment with what's possible and what not.   

We wanted to have predictable URLs to be able to generate links or play around with separate projects for the websites so I took the chance here.   SHA and PR number is enough (one per build, one per branch use case but without the restriction in branch name length.

.github/workflows/main.yml
nkuehn: yes, the workflow just lacked the respective logic yet

.github/workflows/main.yml
nkuehn: OK, makes sense for the kit.  In the content repos the static checks on content are independent of the deployment so you always have  preview even if there are still formal issues.

.github/workflows/main.yml
nkuehn: let's change such things separately.  I prefer local, it's faster and more reliable.

.github/workflows/main.yml
```diff
@@ -54,17 +60,56 @@ jobs:
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          vercel-args: ${{ env.VERCEL_ARGS }}
+          vercel-org-id: ${{ secrets.VERCEL_ORG_ID}}
+          scope: commercetools
+          vercel-project-id: ${{ secrets.VERCEL_PROJECT_ID}}
+          # there are length restrictions to domains and only *.commercetools.vercel.app works
+          # reliably for generated subdomains. No branch name domain because branch names can be long.
+          alias-domains: |
+            docskit-sha-${{ github.sha }}.commercetools.vercel.app
+            docskit-pr-{{PR_NUMBER}}.commercetools.vercel.app
```
nkuehn, line 91: This might be an issue when running a production build on main.  I guess the action will resolve it to `docskit-pr-.commercetools.vercel.app`  which is not a real issue but might collide with later production deploys

nkuehn:
@emmenko this is OK for me to try out now.  I have no better idea how to test the production behavior other than merging to main and fixing issues in a new PR.   Sounds ugly though if we need multiple attempts.

emmenko:
I think it's fine to merge this and see what happens. It's just the kit repo after all, no "real production" usage.

.github/workflows/main.yml
emmenko: I would suggest to split the "Deploy to Vercel" step into 2:

* one that is configured for production (no alias domains, passing the `--prod` flag)
* one that is configured for previews

It's a bit more verbose but it's more explicit which config is used for what. Then we also don't need to do things like `echo "VERCEL_ARGS=--prod" >> $GITHUB_ENV`

timonrey (approved):
Let's try it out :)
Thanks!

.github/workflows/main.yml
nkuehn: PS: will un-do this again because getting the deployment URL from two different steps conditionally is just as messy

PR #1200, opened 2022-01-19
feat(writing-style): allow more acronyms

emmenko:


packages/writing-style/styles/commercetools/Acronyms.yml
```diff
@@ -31,6 +31,7 @@ exceptions:
   - HTML
   - HTTP
   - HTTPS
+  - ICU
```
adinakleine, line 34: Where is that used and what is it. I know only Intensive Care Unit as ICU ;-)

packages/writing-style/styles/commercetools/Acronyms.yml
```diff
@@ -45,6 +46,7 @@ exceptions:
   - MIT
   - NET # otherwise ".NET" is complaining but it's generally known among developers
   - NPM
+  - OIDC
```
adinakleine, line 49: I found this only ones in our docs? Is it planned to be used more often?

adinakleine (approved):
@emmenko I am okay with adding acronyms you want to add. Left two comments out of curiosity.

packages/writing-style/styles/commercetools/Acronyms.yml
emmenko: This refers to the "internationalization" syntax (`Intl`). It's what we use in the MC.

https://unicode-org.github.io/icu/userguide/icu/

packages/writing-style/styles/commercetools/Acronyms.yml
emmenko: We use it a bunch of times in the new Custom Apps docs.

emmenko: It's also a common acronym for user authentication topics.

packages/writing-style/styles/commercetools/Acronyms.yml
nkuehn: Not sure if ICU is that well known to not explain it the first time it occurs.  But I would leave it as "allowed" because explaining it as "International Components for Unicode" does not make anything clearer, too :-)

packages/writing-style/styles/commercetools/Acronyms.yml
nkuehn: I think "OpenID Connect" is not too long or unwieldy to not directly use it instead of an acronym.  In general, if the full form is easy to read and understand, always use the full form. 

If you really want "OIDC" we have to whitelist it though because vale would not detect the initial explanation on a page because it's not these four words)

nkuehn: sorry for the unclear text:   I vote for not using "OIDC" but spelling it out "OpenID Connect".  That is common practice on other documentation sites I remember (just spent a few hours on the topic in the self learning project two days ago)

emmenko: Ok, I'm fine with settling for that. Thanks for the feedback.

nkuehn (approved):

PR #1201, opened 2022-01-20
fix: svgr command

emmenko:
Follow up of #1165 

See https://react-svgr.com/docs/cli/


nkuehn (approved):
thank you.  I think I introduced that bug

PR #1202, opened 2022-01-20
feat: allow to render a global notification via overrides

emmenko:
FYI: This is currently needed for Custom Apps docs. See [designs](https://commercetools.invisionapp.com/console/commercetools-API-DOCS---Component-V2-ckem9zrer39id01z90qsjc8iy/ckycrxvcs018a01zidv1j1sst/play)

Screen Shot 2022-01-20 at 20 46 50

Screen Shot 2022-01-20 at 20 48 07

Screen Shot 2022-01-20 at 20 48 28

PS: I decided to not have VRTs for this (for now), as we can't really turn this easily on/off on demand.

### DIFF
--- a/.changeset/green-cycles-own.md
+++ b/.changeset/green-cycles-own.md
@@ -2,4 +2,4 @@
 '@commercetools-docs/gatsby-theme-docs': minor
 ---
 
-Allow to render a global notification using the [override](https://commercetools-docs-kit.vercel.app/documentation/configuration/overrides) feature for the `global-notification.js` file.
+Allow to render a global notification to all pages of the website. Configure it via the `globalNotification` option of the Gatsby theme.
--- a/.changeset/green-cycles-own.md
+++ b/.changeset/green-cycles-own.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': minor
+---
+
+Allow to render a global notification using the [override](https://commercetools-docs-kit.vercel.app/documentation/configuration/overrides) feature for the `global-notification.js` file.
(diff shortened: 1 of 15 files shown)


nkuehn (commented):
Can you give me some context of the use case?  It's hard to evaluate without that context and the docs do not give a lot of detail. 

From the code it looks like the implementation is for a complete website but visually it looks to me like it's for a certain given page. 

If it's per page I would  consider making it a fixed feature that's triggered from frontmatter and not a gatsby level override.    (actuall, even if it's per website I would consider making it a feature and not an ovverride)

emmenko:
> Can you give me some context of the use case? It's hard to evaluate without that context and the docs do not give a lot of detail.

Yes sorry, I'll DM you.

> From the code it looks like the implementation is for a complete website but visually it looks to me like it's for a certain given page.

It's meant to be for the complete website actually.

Using an overrides has some pros and cons yes but it's a simpler and flexible solution.

Using something else like a frontmatter is also an option yes but I felt to be a bit more restrictive about what content you can render in the notification (ideally you should be able to use markdown syntax).

However, if you prefer to go the frontmatter path I can take another look at it and change the implementation.

nkuehn (approved):
discussed with @emmenko  personally.  I'm not a fan of introducing more features based on gatsby component shadowing but think we should not invest more time before the case actually comes up a second time.  In that case we'd have to revisit the layout, too in my opinion.

emmenko:
@nkuehn I refactored it using the gatsby config. Can you take another look?

nkuehn (approved):
Thank you !

PR #1122, opened 2021-10-15
WIP Create new documentation website

timonrey:
part of #953 

This pull request creates a new website with the navigation structure.

@nkuehn could you please look over the navigation structure and make some suggestions if you want to change something. Thanks!

@naveenchkrvrthy FYI

PR #1203, opened 2022-01-20
fix: missing readmes

emmenko:
Caused by #1122

PS: The readmes contain links to the new documentation pages. It's important to have the actual file so that in the NPM website you can at least see something and click on the link.


nkuehn (approved):
Thank you very much @emmenko  for being attentive on this aspect.  I really did not even think about it.

PR #1205, opened 2022-01-24
Refactor Docs UI-Kit to Typescript

nkuehn:
I wanted to learn typescript (privately) and decided taking the docs ui-kit package would be a suitable production target to learn it in practice, too.  It's a library used by others and it helps the users of the library writing correct code faster if the lib is types (no matter if the user code is in TS or JS). 

I found only one actual bug (the size parameter to the MC UI kit spinner was wrong, it's actually "scale") but had to add a bunch of type guards which technically improve robustness.  

---
Open Issues:
  * needs a bit more  review
  * waiting for github actions to start running again


packages/ui-kit/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  extends: ['@commercetools-frontend/eslint-config-mc-app'],
+  rules: {
+    'prettier/prettier': [
+      'error',
+      {
+        parser: 'typescript',
+      },
+    ],
+  },
```
nkuehn, line 10: something is wrong here I bet.  This is one of the ways I tried to get prettier to not crash on certain ts syntax.

packages/ui-kit/package.json
```diff
@@ -17,7 +17,7 @@
   "module": "dist/commercetools-docs-ui-kit.esm.js",
   "files": ["dist", "package.json", "LICENSE", "README.md"],
   "scripts": {
-    "generate-icons": "svgr -d src/icons/generated -- src/icons/svg"
+    "generate-icons": "svgr -d src/icons/generated --typescript --filename-case kebab --no-index -- src/icons/svg"
```
nkuehn, line 20: The CamelCasing used before made eslint very unhappy when importing them where they are used.  Rabbithole: with kebab case generated components the automatic generation of the index file fails

packages/ui-kit/src/components/rss-feeds.tsx
```diff
@@ -110,11 +132,11 @@ const RssFeeds = (props) => {
       />
     );
   }
-  return <LoadingSpinner size="s">{'Loading feeds'}</LoadingSpinner>;
+  return <LoadingSpinner scale="s">{'Loading feeds'}</LoadingSpinner>;
```
nkuehn, line 135: this was the only actual bug typing found :-)

packages/ui-kit/src/utils/code-block-parse-options.ts
```diff
@@ -1,19 +1,25 @@
 import parseNumericRange from './parse-numeric-range';
 
-const getLinesForRange = (value) => {
+type OptionValue = string | boolean | number[] | undefined;
```
nkuehn, line 3: This whole area is ugly and may be refactored completely later.  The purpose of the options parsing is that it changes the type and data of a given property from string to e.g. an array of numbers.  ideally the implentation would have separate types for the input and output but that's not how the code using it is build.  

To not overengineer I left this here.

packages/ui-kit/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
```
nkuehn, line 1: Mostly from the intial automatic migration plus some the emotion thing I added to get emotion run. 

Maybe we have a template or standard in the company from like for eslint?

packages/gatsby-theme-docs/src/layouts/content.js
```diff
@@ -9,6 +9,7 @@ import { BetaFlag, ContentPagination, GlobalNotification } from '../components';
 import PlaceholderPageHeaderSide from '../overrides/page-header-side';
 import PlaceholderPageHeaderSideBannerArea from '../overrides/page-header-banner-area';
 import LayoutApplication from './internals/layout-application';
+import LayoutPage from './internals/layout-page';
```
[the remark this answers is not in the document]
nkuehn: "wald vor bäumen nicht gesehen"..

packages/ui-kit/tsconfig.json
emmenko: Only the commented code got removed, right?

packages/gatsby-theme-docs/index.js
```diff
@@ -1,6 +1,7 @@
 export { default as LayoutApplication } from './src/layouts/internals/layout-application';
 export { default as LayoutSidebar } from './src/layouts/internals/layout-sidebar';
 export { default as LayoutHeader } from './src/layouts/internals/layout-header';
+export { default as LayoutPage } from './src/layouts/internals/layout-page';
```
emmenko, line 4: Any particular reason for exporting this?

packages/ui-kit/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  extends: ['@commercetools-frontend/eslint-config-mc-app'],
```
emmenko, line 2: I would avoid having nested eslint configs. You can use `overrides` in the main `.eslintrc.js` file to target specific things.

packages/ui-kit/.eslintrc.js
emmenko: What was the error you got?

packages/ui-kit/package.json
emmenko: I'm not sure I understand the issue here...

packages/ui-kit/package.json
```diff
@@ -32,14 +32,19 @@
     "lodash.escape": "4.0.1",
     "lodash.throttle": "4.1.1",
     "prism-react-renderer": "1.2.1",
-    "prop-types": "15.8.1",
```
emmenko, line 35: In theory `prop-types` should still be a dependency, so that projects that are not using TS can still validate the props.

For that, we need to use the `babel-plugin-typescript-to-proptypes`.

https://github.com/commercetools/ui-kit/blob/482e977c7f0cf3dcd342731c965e9016da5dc417/babel.config.js#L15

packages/ui-kit/package.json
```diff
@@ -32,14 +32,19 @@
...
     "react-intl": "^5.7.0",
+    "react-is": "^17.0.2",
     "swr": "1.1.2",
     "tiny-invariant": "1.2.0"
   },
   "devDependencies": {
+    "@svgr/cli": "6.2.0",
+    "@types/react": "^17.0.38",
+    "@types/react-dom": "^17.0.11",
+    "@types/react-is": "^17.0.3",
```
emmenko, line 44: Those should be normal dependencies

packages/ui-kit/package.json
```diff
@@ -32,14 +32,19 @@
...
     "react": "17.0.2",
-    "react-dom": "17.0.2"
+    "react-dom": "17.0.2",
+    "typescript": "^4.5.5"
```
emmenko, line 47: There should be no need for this, as it's already defined in the root package.json.

packages/ui-kit/src/components/code-block.tsx
```diff
@@ -1,12 +1,11 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
-import { css, ThemeProvider, useTheme } from '@emotion/react';
+import { css, Theme, ThemeProvider, useTheme } from '@emotion/react';
```
emmenko, line 4: Types should be imported using the `type` keyword.

```suggestion
import type { Theme } from '@emotion/react';
```

The latest TS versions also support a more advanced syntax:

```suggestion
import { css, type Theme, ThemeProvider, useTheme } from '@emotion/react';
```

packages/ui-kit/src/components/code-block.tsx
```diff
@@ -181,7 +187,7 @@ const getLineStyles = (theme, options) => {
  * yarn start
  * ```
  */
-const languageAliases = {
+const languageAliases: { [key: string]: Language } = {
```
emmenko, line 190: Nit: we can also use `Record`

```suggestion
const languageAliases: Record<string, Language> = {
```

packages/ui-kit/src/components/content-notifications.tsx
```diff
@@ -82,21 +84,26 @@ const ContentNotification = (props) => {
...
+        <TypographyContainer>{children}</TypographyContainer>
       </SpacingsInline>
     </Container>
   );
 };
-ContentNotification.propTypes = {
-  type: PropTypes.oneOf(['error', 'warning', 'info']).isRequired,
-  children: PropTypes.node.isRequired,
-};
-ContentNotification.defaultProps = {
```
emmenko, line 94: Please define `defaultProps`. Example how to use it: https://github.com/commercetools/ui-kit/blob/f1bdecca464f73e94edc6b758ff271cb81695a3e/packages/components/avatar/src/avatar.tsx#L46-L54

packages/ui-kit/src/components/content-notifications.tsx
```diff
@@ -10,7 +9,7 @@ import {
 import { TypographyContainer } from './markdown';
 import { colors, tokens, dimensions } from '../design-system';
 
-const getIconByType = (type) => {
+const getIconByType = (type: string) => {
```
emmenko, line 12: Better to use the props type:

```suggestion
const getIconByType = (type: ContentNotificationProps['type']) => {
```

emmenko (commented):
Had a short quick review (haven't gone through all changes yet).

packages/ui-kit/package.json
nkuehn: Sure? We're not exporting anything that takes or returns any of these interfaces.

packages/ui-kit/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+  "compilerOptions": {
+    "jsx": "react-jsx",
+    "target": "es2016" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
+    "jsxImportSource": "@emotion/react" /* Specify module specifier used to import the JSX factory functions when using `jsx: react-jsx*`.` */,
+    "module": "commonjs" /* Specify what module code is generated. */,
+    "esModuleInterop": true /* Emit additional JavaScript to ease support for importing CommonJS modules. This enables `allowSyntheticDefaultImports` for type compatibility. */,
+    "forceConsistentCasingInFileNames": true /* Ensure that casing is correct in imports. */,
```
nkuehn, line 8: this is the one that broke the generated icons.  I'll remove it and can then revert the cruft

packages/ui-kit/package.json
nkuehn: resolved through allowing camelCased typescript imports

packages/gatsby-theme-docs/src/templates/page-content.js
```diff
@@ -16,42 +16,39 @@ const ContentCards = (props) => (
 );
 
 const PageContentTemplate = (props) => (
-  console.log(props),
```
nkuehn, line 19: only this line was changed (remove that leftover console.log) but github shows the whole block as changed.

packages/ui-kit/src/components/code-block.tsx
nkuehn: I had the impression that the index form is conventional when any string is allowed and Record is interesting when you e.g. have a specific union or so as keys.

packages/ui-kit/src/components/content-notifications.tsx
nkuehn: My understanding was that defaultProps are already on the deprecation path and that the object spread defaults I use above are the way to go now (and much more concise and javascript-ish :-) )

https://github.com/yannickcr/eslint-plugin-react/issues/2396

nkuehn:
closing this, seems stuck. New PR

PR #1209, opened 2022-01-25
fix: global notification overlap with scroll top anchor positions

emmenko:
Follow up of #1202 

Solution for the scroll anchor position has been taken from here: https://getpublii.com/blog/one-line-css-solution-to-prevent-anchor-links-from-scrolling-behind-a-sticky-header.html

PR #1211, opened 2022-01-26
fix: keep flexible height for global notification on small viewports

emmenko:
Follow up of #1209

Hopefully this is the last change.

### DIFF
--- a/packages/gatsby-theme-docs/src/layouts/internals/layout-page-release-notes-filters.js
+++ b/packages/gatsby-theme-docs/src/layouts/internals/layout-page-release-notes-filters.js
@@ -82,7 +82,6 @@ const ToggleMenuButton = styled.div`
   );
   right: ${designSystem.dimensions.spacings.m};
   cursor: pointer;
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-docs/src/layouts/internals/layout-page-release-notes-filters.js
```diff
@@ -82,7 +82,6 @@ const ToggleMenuButton = styled.div`
   );
   right: ${designSystem.dimensions.spacings.m};
   cursor: pointer;
-  z-index: ${designSystem.dimensions.stacks.base};
```
emmenko, line 85: I accidentally added this to the wrong component last time out.

nkuehn (approved):
I did not try to fully understand it.  

But it would be a good idea to make this feature part of the VRTs now that's its a configurable feature and has gone through some iterations of tuning and testing.  We have two sites tested, we could set it up e.g. on the API docs smoke test site for the VRTs.

emmenko:
Sure, we can add the VRTs as a follow up then.

PR #1217, opened 2022-01-28
Refactor uikit to typescript

nkuehn:
I wanted to learn typescript (privately) and decided taking the docs ui-kit package would be a suitable production target to learn it in practice, too.  It's a library used by others and it helps the users of the library writing correct code faster if the lib is types (no matter if the user code is in TS or JS). 

I found only one actual bug (the size parameter to the MC UI kit spinner was wrong, it's actually "scale") but had to add a bunch of type guards which technically improve robustness.


nkuehn:
Here's one visual difference worth investigating: It rather looks like a content change that we did when creating the documentation website but to be safe let's check
image

babel.config.js
```diff
@@ -1,12 +1,13 @@
+// should track https://github.com/commercetools/ui-kit/blob/main/babel.config.js as suitable
 module.exports = {
   presets: [
     [
       '@commercetools-frontend/babel-preset-mc-app',
       {
-        // TODO: change this to `automatic` when using the new runtime
-        runtime: 'classic',
+        runtime: 'automatic',
```
emmenko, line 7: This should stay as `classic` for now. We can migrate to `automatic` as a separate step.

packages/gatsby-theme-docs/index.js
```diff
@@ -1,7 +1,6 @@
 export { default as LayoutApplication } from './src/layouts/internals/layout-application';
 export { default as LayoutSidebar } from './src/layouts/internals/layout-sidebar';
 export { default as LayoutHeader } from './src/layouts/internals/layout-header';
-export { default as LayoutPage } from './src/layouts/internals/layout-page';
```
emmenko, line 4: Why removing this export?

packages/ui-kit/package.json
```diff
@@ -34,10 +34,15 @@
...
     "react-intl": "^5.7.0",
+    "react-is": "^17.0.2",
     "swr": "1.1.2",
     "tiny-invariant": "1.2.0"
   },
   "devDependencies": {
+    "@svgr/cli": "6.2.0",
+    "@types/react": "^17.0.38",
+    "@types/react-dom": "^17.0.11",
+    "@types/react-is": "^17.0.3",
```
emmenko, line 45: I would define them as `dependencies`. This way, consumers don't need to explicitly add them.

packages/ui-kit/src/components/code-block.tsx
```diff
@@ -1,12 +1,11 @@
...
 import ReactDOM from 'react-dom';
-import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
-import { css, ThemeProvider, useTheme } from '@emotion/react';
+import { css, type Theme, ThemeProvider, useTheme } from '@emotion/react';
 import Tooltip from '@commercetools-uikit/tooltip';
 import SpacingsInline from '@commercetools-uikit/spacings-inline';
 import { ClipboardIcon } from '@commercetools-uikit/icons';
-import Highlight, { defaultProps } from 'prism-react-renderer';
+import Highlight, { Language, defaultProps } from 'prism-react-renderer';
```
emmenko, line 8:
```suggestion
import Highlight, { type Language, defaultProps } from 'prism-react-renderer';
```

packages/ui-kit/src/components/content-notifications.tsx
```diff
@@ -1,5 +1,4 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { ReactNode } from 'react';
```
emmenko, line 1:
```suggestion
import React, { type ReactNode } from 'react';
```

packages/ui-kit/src/components/media-query.ts
```diff
@@ -2,8 +2,13 @@ import styled from '@emotion/styled';
...
 import { dimensions } from '../design-system';
 
+type MediaQueryProps = {
+  forViewport?: string;
+  hideIfMatch?: boolean;
+};
+
 const MediaQuery = styled.div`
-  ${(props) => {
+  ${(props: MediaQueryProps) => {
```
emmenko, line 11: For styled components, you can pass the prop type as the type argument

```ts
styled.div<MediaQueryProps>``
```

https://emotion.sh/docs/typescript#passing-props

packages/ui-kit/src/components/rss/rss-feeds.spec.ts
```diff
@@ -1,4 +1,7 @@
-import { transformData } from './rss-feeds';
+/**
+ * @jest-environment jsdom
+ */
```
emmenko, line 3: What's this for?

packages/ui-kit/src/components/rss/rss-feeds.tsx
```diff
@@ -46,41 +58,51 @@ async function fetcher(...args) {
...
-export const transformData = (data) => {
+type DatedData = { pubDate: string };
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const transformData = (data: any) => {
   // First, we need to get the oldest release note from each feed,
   // which is always the last entry of the list.
   const lastEntryOfList = data
-    .reduce((list, feed) => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .reduce((list: any, feed: string | any[]) => {
```
emmenko, line 68: FYI: usually you pass a type argument to `reduce`, instead of typing each function argument.

packages/ui-kit/src/design-system.ts
```diff
@@ -7,13 +7,76 @@ const asideColumnWidth = pageWidthSmall;
...
   const remNumber = parseInt(pxNumber, 10) / rootFontSizeNumber;
   return `${remNumber}${suffix}`;
 };
 
-export const colors = {
+// explicit types for the color part because the structure allows multiple instances
+// (light / dark,  primary / secondary for code blocks).
+// The rest of the design system is enough to be inferred
+export type ColorKey = keyof ColorSet;
+export type ColorSet = {
```
emmenko, line 20: Usually for constants you should use `as const`. There is no need to have a separate type.

```ts
export const colors = {
  // ...
} as const;
```

packages/ui-kit/src/utils/create-styled-icon.ts
```diff
@@ -50,18 +50,19 @@ const getColor = (color) => {
...
       *:not([fill='none']) {
         fill: ${getColor(props.color)};
       }
       ${getSizeStyle(props.size)}
     `
   );
-  StyledComponent.propTypes = {
-    color: PropTypes.string,
-    size: PropTypes.oneOf(['small', 'medium', 'big', 'scale']),
+  type StyledComponentProps = {
```
emmenko, line 62: No need to define this inside the function

package.json
```diff
@@ -13,7 +13,7 @@
     "build-packages": "preconstruct build",
     "build": "./scripts/build.sh",
     "vercel-build": "echo \"you're welcome - nothing to build\"",
-    "generate-icons": "yarn workspaces foreach -p --include @commercetools-docs/gatsby-theme-docs,@commercetools-docs/ui-kit,@commercetools-website/docs-smoke-test run generate-icons && prettier --write '**/generated/*.js'",
+    "generate-icons": "yarn workspaces foreach -p --include @commercetools-docs/gatsby-theme-docs,@commercetools-website/docs-smoke-test,@commercetools-docs/ui-kit run generate-icons && prettier --write '**/generated/*.{js,ts,tsx}'",
     "lint": "jest --projects jest.eslint.config.js jest.stylelint.config.js jest.text.config.js",
     "lint:js": "jest --config jest.eslint.config.js",
     "lint:css": "jest --config jest.stylelint.config.js",
```
emmenko, line 19: We should also add a `typecheck` command: https://github.com/commercetools/ui-kit/blob/c79b62bd25cd3da1ffe7fd9e39d08b8335218942/package.json#L41

And include it in CI: https://github.com/commercetools/ui-kit/blob/c79b62bd25cd3da1ffe7fd9e39d08b8335218942/.github/workflows/main.yml#L43-L44

And in lint-staged: https://github.com/commercetools/ui-kit/blob/c79b62bd25cd3da1ffe7fd9e39d08b8335218942/lint-staged.config.js#L17-L29

emmenko (changes requested):
Overall looks good, thanks!

I left some comments to fix some things and add some missing things.

packages/gatsby-theme-docs/index.js
nkuehn: by accident. added it again

packages/ui-kit/src/components/rss/rss-feeds.spec.ts
emmenko: Yes but aren't we already using `jsdom`? See https://github.com/commercetools/commercetools-docs-kit/blob/890dc3db00f97434f0cdc60d3e0efcbcbbcb33e3/jest.test.config.js#L35

packages/ui-kit/src/components/rss/rss-feeds.tsx
nkuehn: thanks. Properly typed the whole RSS thing, inference led to not needing a lot of signature on the reducers at all

packages/ui-kit/src/design-system.ts
nkuehn: The types are here because I had issues letting the code that uses it understand that "Primary" and "secondary" code block style are the same identical interface. 

this code in `code-block.tsx`:
```
  const codeBlockTheme = {
    codeBlockColors:
      colors.light.codeBlocks[props.secondaryTheme ? 'secondary' : 'primary'],
  };
```

packages/ui-kit/package.json
nkuehn: Why I made them a dev dependency:

These data structures are not part of what the ui kit exorts so consumers don't need to know them and we can save consumers an unused dependency that they just have to maintain to silence yarn warnings.  

If a package exports an interface it should be a dependency, sure.

babel.config.js
nkuehn: I took this from the MC ui-kit to have the most similar setup. At least I hope I took the right piece. Set back to classic, but had no visible issues with automatic.

package.json
nkuehn: +1

packages/ui-kit/package.json
emmenko: The uikit package does export the type declarations actually. Anyway, not super important.

packages/ui-kit/src/design-system.ts
emmenko: I pushed a commit to use `as const`. https://github.com/commercetools/commercetools-docs-kit/pull/1217/commits/9d7c327010e3bafd09aa914ce85fab3ed925b69d

packages/ui-kit/src/components/rss/rss-feeds.spec.ts
nkuehn: turned out to be a follow-up error:  `jest-preset-gatsby` was not configured for typescript so the whole config was not picked up (I had tested only that file first and oversaw that the test is not running any more in the full test suite)

nkuehn:
@emmenko  thank you for all the feedback.  I learnt quite some more.  

There's one issue remaining that I know of, `tsc-files` complains about issues that `tsc` does not complain about.  I had to comment out the check in `lint-staged` to push at all.  Did you ever see this before? 

https://github.com/commercetools/commercetools-docs-kit/pull/1217/files#diff-29c15f0b5d8d1144bd22153c302cde16c2de092d65b122a618cbdb3023336346R29-R30

emmenko (approved):
This should be good to go from my side. Let's see what CI has to say about it.

I refactored a bit the Emotion `Theme` typings to explicitly list what properties are used.

PR #1218, opened 2022-01-28
ci: use linkinator action

emmenko:
https://github.com/JustinBeckwith/linkinator-action


.github/workflows/after-deployment.yaml
```diff
@@ -12,9 +12,16 @@ jobs:
...
         with:
-          node-version: '16' # ensure a 7+ npx CLI behavior
+          node-version-file: '.nvmrc'
+
       - name: Check links
-        # the docs production domain has to be rewritten to the preview URL to allow new pages to be created without the link checker
-        # erroring on the canonical URL of the new page. Known compromise to not find links that point to the absolute production URL
-        # but the docs kit is automatically making those a relative URL in authored content so the impact is low.
-        run: npx --yes linkinator@2.x --config linkinator.remote.config.json --url-rewrite-replace=${{ github.event.deployment_status.target_url }} ${{ github.event.deployment_status.target_url }}
+        uses: JustinBeckwith/linkinator-action@v1.10.0
```
emmenko, line 18: It uses linkinator v3

.github/workflows/main.yml
```diff
@@ -161,3 +161,31 @@ jobs:
...
+        uses: JustinBeckwith/linkinator-action@v1.10.0
+        with:
+          paths: docskit-sha-${{ github.sha }}.commercetools.vercel.app
+          config: linkinator.remote.config.json
+          markdown: false
+          # the docs production domain has to be rewritten to the preview URL to allow new pages to be created without the link checker
+          # erroring on the canonical URL of the new page. Known compromise to not find links that point to the absolute production URL
+          # but the docs kit is automatically making those a relative URL in authored content so the impact is low.
+          urlRewriteReplace: docskit-sha-${{ github.sha }}.commercetools.vercel.app
+          urlRewriteSearch: https://docs.commercetools.com
```
nkuehn, line 191: the urlRewriteSearch parameter is already set in the config file, but it's ok to set it again here. 
Clearer to understand what's going on and the docs URL won't change that often :-)

.github/workflows/main.yml
```diff
@@ -161,3 +161,31 @@ jobs:
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version-file: '.nvmrc'
+
```
nkuehn, line 180: I dont think we need nodejs here.  Actions run without explicit installation of their environment.

.github/workflows/main.yml
```diff
@@ -156,3 +158,28 @@ jobs:
...
+      - name: Check links
+        uses: JustinBeckwith/linkinator-action@v1.10.0
+        with:
+          paths: docskit-sha-${{ github.sha }}.commercetools.vercel.app
+          config: linkinator.remote.config.json
+          markdown: false
+          # the docs production domain has to be rewritten to the preview URL to allow new pages to be created without the link checker
+          # erroring on the canonical URL of the new page. Known compromise to not find links that point to the absolute production URL
+          # but the docs kit is automatically making those a relative URL in authored content so the impact is low.
+          urlRewriteReplace: docskit-sha-${{ github.sha }}.commercetools.vercel.app
```
nkuehn, line 184:
```suggestion
          paths: https://docskit-sha-${{ github.sha }}.commercetools.vercel.app
          config: linkinator.remote.config.json
          markdown: false
          # the docs production domain has to be rewritten to the preview URL to allow new pages to be created without the link checker
          # erroring on the canonical URL of the new page. Known compromise to not find links that point to the absolute production URL
          # but the docs kit is automatically making those a relative URL in authored content so the impact is low.
          urlRewriteReplace: https://docskit-sha-${{ github.sha }}.commercetools.vercel.app
```

linkinator.remote.config.json
```diff
@@ -2,18 +2,19 @@
...
-    "/import-export"
+    "/.+/api$",
+    "/.+/api/",
+    "/.+/custom-applications",
+    "/.+/docs/",
+    "/.+/import-export",
+    "/.+/merchant-center",
+    "/.+/sdk",
+    "/.+/tutorials",
+    "/.+/getting-started"
```
nkuehn, line 17: Just curious: What problem does this address?  Looks like a workaround to me, do we have `/../api` style link hrefs in the final static output?

nkuehn (approved):

linkinator.remote.config.json
emmenko: I found out that paths like `/docs/` interfere with the domain (for example `https://docs-kit`), resulting in all links to be ignored.

Adding the `.+` seems to fix that. So for consistency I added it to all paths.

nkuehn: uh, ugly edge case.  Thanks.

Issue #539, opened 2020-07-03
Try improving the error message on MDX syntax errors (not crash the dev server, see the error)


nkuehn:
Currently, when the user writes an MDX page and either has an MDX syntax error (e.g. saving before closing a tag) or writes frontmatter that does not comply to the schema (e.g. an empty value) the error message is completely unrelated - in most cases it complains that certain frontmatter is missing but the root cause is an MDX syntax error in the body. 

This is very confusing to authors and if we can at least show a more specific error (or ideally even pointing to the line in the MDX that's wrong) that would be fabulous. 

A first step could be to dump the node content to the console to see whether it contains indication of the cause of the problem. 

---
**update**:  I just realized it's not even possible to recover at all from an MDX syntax error without restarting the dev server (in the content repo using the kit).  Marking this a bug therefore. 

---
**update**:  dev server is crashing in full now. we need to pull this into the priority

nkuehn:
@davifantasia  @emmenko  I did a test with the gatsby MDX starter to see what the standard behavioris:

```
gatsby new test-site https://github.com/ChristopherBiscardi/gatsby-starter-mdx-basic
cd test-site
yarn 
yarn upgrade --latest
npx gatsby develop -o
```
(you have to upgrade because older gatsby-plugin-mdx versions completely crash on MDX syntax errors)

Result when introducing a syntax error into an MDX page:

![image](https://user-images.githubusercontent.com/4006139/87790645-32447c80-c841-11ea-8857-a08d708095d8.png)

**When reverting the syntax error, the develop server recovers**

That means, the problem is introduced somewhere in our own gatsby setup. Something in our architecture approach seems to be different.   It could be many things, e.g. the fact that we generate pages ourselves and not the built-in "pages" directory behavior or one of the many plugins or the fact that we're a theme etc. pp.     It'll require some experimentation to find out.

davifantasia:
We'll revisit this to check if this issue is still the case.

nkuehn:
@davifantasia  @timonrey  I just checked and the behavior got even worse unfortunatlely,  the dev server is crashing hard e.g. if you don't close a tag and save the MDX file. 

![image](https://user-images.githubusercontent.com/4006139/130083311-fb1e02a5-3549-42d5-b1f0-5d5e9c400dc8.png)

nkuehn:
Blocked by "we have no capacity for dissecting the complete node creating process and no idea about where this could come from".

PR #1221, opened 2022-02-01
Handle MDX crashes (usually syntax errorrs), at least a little

nkuehn:
Addresses #539 

 * MDX node ceases to exist if a syntax error is introduced while dev mode is running. 
 * ContentPage node still exists but can't get its parent any more.  The connection is lost and seems to not recover although the gatsby docs claim that the API is aware of the relation to the path when writing resolvers. 
 * I catch the case and show an error message now,  capturing the way MDX generates react code.  not nice but the error is happening inside the resolver code which does not have a lot of context. 
 *  Still need to restart the server to recover but at least it does not crash without any hint to the potential reason any more. 

I can only imagine to get this fundamentally more stable by removing our own GraphQL facade layer altogether and using Gatsby more "Off the shelf".  But that would be a big refactor.   

Is this an improvement or just as horrible?    Maybe there is a middle ground that does not dynamically resolve the MDX body from the parent node but just copies it over and gives us easier error handling?


emmenko (approved):
Nice trick!

PR #1228, opened 2022-02-04
Implement subsection support in child-section-nav

### LINKED ISSUE #1215: Improve hierarchy in the child-section-nav
We want to improve the way how hierarchy in the child section navigator is displayed.
To do so, we want to adjust the design of the table content:
https://commercetools.invisionapp.com/console/share/UM35IY9N72

**Changes to be made:**
- Show a `-` before subsections as it would be subpoints
- Between the subsection groups, there should be a space of 24px. This should only happen after a section with subsections, and not after every section. See invision design for more information.

**New feature to be implemented:** 
The search feature in the child section nav does not support the search of subsections yet, so this needs to be implemented as well if we want the subsections to be used properly.

timonrey:
closes #1215

### DIFF
--- a/packages/gatsby-theme-docs/src/components/child-sections-nav.js
+++ b/packages/gatsby-theme-docs/src/components/child-sections-nav.js
@@ -127,8 +127,8 @@ const SectionNavigation = (props) => {
       {props.isSubItem ? (
         <Link href={item.url}>
           <SpacingsInline scale="xs">
-            <div>-</div>
-            <div>{item.title}</div>
+            <span>-</span>
+            <span>{item.title}</span>
           </SpacingsInline>
         </Link>
       ) : (
--- a/packages/gatsby-theme-docs/src/components/child-sections-nav.js
+++ b/packages/gatsby-theme-docs/src/components/child-sections-nav.js
@@ -7,7 +7,6 @@ import {
   createStyledIcon,
 } from '@commercetools-docs/ui-kit';
 import SecondaryIconButton from '@commercetools-uikit/secondary-icon-button';
-import SpacingsStack from '@commercetools-uikit/spacings-stack';
 import SpacingsInline from '@commercetools-uikit/spacings-inline';
 import { CloseIcon } from '@commercetools-uikit/icons';
 import { usePageData } from '../hooks/use-page-data';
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-docs/src/components/child-sections-nav.js
```diff
@@ -123,15 +124,22 @@ const SectionNavigation = (props) => {
   const item = props.tocNode;
   return (
     <>
-      <Link href={item.url}>{item.title}</Link>
+      {props.isSubItem ? (
+        <Link href={item.url}>
+          <SpacingsInline scale="xs">
+            <div>-</div>
+            <div>{item.title}</div>
```
nkuehn, line 131: By meaning, this are rather `<span>`s - segements inside the text flow and not block level layout elements.

nkuehn: By meaning, this are rather `<span>`s - segments inside the text flow and not block level layout elements.

packages/gatsby-theme-docs/src/components/child-sections-nav.js
```diff
@@ -161,16 +170,47 @@ const ChildSectionsNav = (props) => {
...
         return true;
       }
+      if (item.items) {
+        item.items.forEach((subItem) => {
+          const filterItem = subItem.title.toLowerCase();
+          const subResults = values.map((term) => {
+            if (term === '' || filterItem.includes(term)) {
+              return true;
+            }
+            return false;
```
nkuehn, line 189: Style opionion: This can be expressed much more concisely.  The `||` OR-expression is evaluating to a boolean anyways, you can directly return it.  You would not even need the body braces and the "return" then

nkuehn (commented):
- they will have to learn in production whether keeping the subsections together in a column will work for them (I think they will overdo it in some cases and find it's strange)
- I really like the filter behavior :-) 

Only superficially reviewed the implementation, esp. not the data preprocessing yet.

nkuehn (approved):
To document:
 - It supports one level of hierarchy inside the child section navigator.  Deeper nesting is not recommended and rendered as the second level.

Issue #1233, opened 2022-02-15
Provide more UI Components  and Utils from the UI-Kit


nkuehn:
For building a PoC / MVP self learning portal in the docs style it would help significantly to have much more components exposed from the UI-Kit package, which are currently inside the `gatsby-theme-docs` package. 

- [X] Markdown fragment renderer (could require allowing Javascript in the Typescript settings, the unified plugin systematics are nontrivial to type).   Addressed in #1236 (migrated)
- [ ] Maybe more of the other utils (scroll to top, sentry reporting,...)
- [X] Cards  (requires the above). Addressed in #1236  (card elements provided, but not the full "boxed" card)
- [ ] The layout internals as far as they are independent of gatsbyJS data structures.  Layouts themselves are gatsby specific.  (migrating "layout elements" is in progress:  https://github.com/commercetools/commercetools-docs-kit/tree/nk-move-layout-elements-to-uikit ) 
- [ ] The pagination buttons (just the styled components, not the full component that holds the gatsby specific  logic)  
- [ ] Icons (e.g. the burger icon) - Addressed in #1236   (icons migrated,  burger in progress in https://github.com/commercetools/commercetools-docs-kit/tree/nk-move-layout-elements-to-uikit ) 
- [ ]  SideBySideContainer  
 
Can (and likely should) be done in multiple steps and PRs

nkuehn:
FYI @FFawzy

FFawzy:
Summary: it turned out to cause more troubles than anticipated, so we will close this for now.

PR #1236, opened 2022-02-16
Refactor UI Kit: move some reusable modules to ui-kit

nkuehn:
addresses first aspects of #1233 

- moves the markdown fragment renderer from the theme to the UI Kit package as a utility function.  This required downgrading further packages from the unified / remark / rehype ecosystem to their pre-ESM versions because the UI Kit is typescript and did not apply magic to use these libraries nevertheless.  It's actually a cleanup we should have done anyways. 
- moves all icons from the theme to the UI Kit package.  There are some remaining that live in the test site, to be addressed separately. 
- separates the card into its layout and style building blocks vs. the Linking logic and "MDX shortcode" facade.  Moves the layout and style elements into the UI Kit package. 

TODO from reviews:
 - move card sizing to design system
 - move remaining icons into the ui kit (there are some remaining in the smoke test site)


packages/gatsby-theme-api-docs/src/components/description.js
```diff
@@ -18,7 +22,11 @@ const DescriptionTextContainer = styled.span`
 // a table, card or bullet list.
 export const DescriptionText = (props) => (
   <DescriptionTextContainer data-search-key="embedded-api-description">
-    {markdownFragmentToReact(props.markdownString, {}, transformURNLinksPlugin)}
+    {markdownFragmentToReact(
+      props.markdownString,
+      { a: GatsbyLink },
```
nkuehn, line 27: The "magic" gatsby link now has to be injected into the fragement renderer because the UI Kit has no understanding of gatsby or the app around it (which is the whole point of a separate UI kit)

packages/gatsby-theme-docs/src/components/card.js
```diff
@@ -1,102 +1,30 @@
...
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
-import { css } from '@emotion/react';
-import { designSystem, Markdown } from '@commercetools-docs/ui-kit';
-import markdownFragmentToReact from '../utils/markdown-fragment-to-react';
-import Link from './link';
+import {
+  Markdown,
+  markdownFragmentToReact,
+  cardElements,
```
nkuehn, line 7: I did not expose all the card component pieces individually becasue a) they don't make sense and are not usable standalone b) they would clutter the namespace too much.

packages/gatsby-theme-docs/src/layouts/internals/footer.js
```diff
@@ -3,8 +3,7 @@ import { useStaticQuery, graphql } from 'gatsby';
 import styled from '@emotion/styled';
 import SpacingsInline from '@commercetools-uikit/spacings-inline';
 import SpacingsStack from '@commercetools-uikit/spacings-stack';
-import { designSystem, MediaQuery } from '@commercetools-docs/ui-kit';
-import { LogoHorizontalSvgIcon } from '../../icons';
+import { designSystem, MediaQuery, Icons } from '@commercetools-docs/ui-kit';
```
nkuehn, line 6: Same for the Icons, they are not exported "flat" in the UI Kit namespace any more for clarity

nkuehn (commented):
explanatory comments

packages/ui-kit/src/components/card-elements.tsx
```diff
@@ -0,0 +1,159 @@
...
+ *                    <Link href={props.href} noUnderline>
+ *                        {props.textLink}
+ *                    </Link>
+ *                  </ReadMore>
+ *                </ReadMoreContainer>
+ *            </StackContainer>
+ *        </StackContainer>
+ *      </StackContainer>
+ *    </Link>
+ *  </CardContainer>
```
timonrey, line 35: Nit: Is this the right place where we should document this?

packages/ui-kit/src/components/card-elements.tsx
```diff
@@ -0,0 +1,159 @@
...
+ */
+
+// Explanation about the following sizes:
+// https://github.com/commercetools/commercetools-docs-kit/pull/427#discussion_r425442556
+// the "inContentColumn" variation prevents only one card being in one row in regular content (except mobile)
+const cardNarrowMinWidth = '242px';
+const cardRegularMinWidth = '328px';
+const cardRegularMinWidthInContentColumn = '288px';
```
timonrey, line 43: Since we are already here: why don't we move these values into the designSystem?

timonrey (approved):
LGTM. Thanks! I like the new structure. And some things are more clear now 💯

packages/ui-kit/src/components/card-elements.tsx
nkuehn: No, not for the long run. But we do not have any place  UI Kit documentation yet so I found it better to document it at all than not.  

I did not want to make this a full UI kit documentation project.  We would have to choose between the kit docs or building a storybook like the MC Ui kit documentation

packages/gatsby-theme-docs/src/layouts/internals/footer.js
emmenko: FYI: this is technically a breaking change, as the package exports changed.

Something to keep in mind for the next time.

emmenko: Also to keep in mind that grouping exports into one object might cause some issues with tree shaking. Maybe it's not the case here but better be careful with that.

nkuehn: dammit, yes - completely obvious that this is all a breaking change (assuming we had external users). 

Concerning the grouping / tree shaking, I think we need to address that once more than the current one and later likely two implementations use it.  For now it's just about decoupling gatsby bound things from non-gatsby-things and keeping it monolithic by design.

PR #1238, opened 2022-02-16
Implement reading time estimation with auto time estimation option

### LINKED ISSUE #1186: Add time estimation for tutorials
The idea is to have a timestamp below the H1 of a tutorials page that shows the reader how long it takes to read the page.
Invision link:
https://commercetools.invisionapp.com/console/share/ED30V03OA2

It can be implemented as an optional frontmatter where the writers can define a specific time. When the frontmatter is defined, the timestamp will be shown in the page under the H1.

timonrey:
closes #1186


packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -306,6 +312,12 @@ exports.onCreateNode = (
     navLevels: node.frontmatter.navLevels
       ? Number(node.frontmatter.navLevels)
       : 3,
+    showTimeToRead: node.frontmatter.showTimeToRead
+      ? Boolean(node.frontmatter.showTimeToRead)
+      : false,
+    setTimeToRead: node.frontmatter.setTimeToRead
+      ? Number(node.frontmatter.setTimeToRead)
```
nkuehn, line 319: Did you test what happens if you put something that does not parse into a Number? It would be good if the dev server would not crash in the situation but rather just ignore it

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -185,6 +185,12 @@ exports.createSchemaCustomization = ({ actions, schema }) => {
           }),
         },
         navLevels: { type: 'Int!' },
+        timeToRead: {
```
nkuehn, line 188: * Why does the GraphQL facade have to pass through both values and not just the effective time to read to be displayed?  The point of introducing a facade is to hide the business logic behind it and not to expose it. I.e. the logic "if the manual frontmatter is set, use it, otherwise use the automatic value" belongs here and not into the frontend code. 
* IMHO, if not necessary for a reasons, the resolverPassthrough should be avoided - it's the thing that makes the dev server crash if something is not OK in the source node.

websites/docs-smoke-test/src/content/views/code-blocks.mdx
```diff
@@ -1,5 +1,7 @@
 ---
 title: Code Blocks
+showTimeToRead: true
```
nkuehn, line 3: The test cases you put into the pages are all just covering the same "good cases".   You should have pages that test unlogical scenarios, too.   E.g. setting a manual time but not showing it or setting something invalid as the frontmatter.

nkuehn (changes requested):
I suggest you refactor this to have a simpler GraphQL and move the logic into the graphQL facade so that the frontend is just displaying it as instructed to.

nkuehn:
@timonrey  better request review after the checks are green ;-)

```
➤ YN0000: The field "ContentPage.showTimeToRead." was explicitly defined as non-nullable via the schema customization API (by yourself or a plugin/theme). This means that this field is not optional and you have to define a value. If this is not your desired behavior and you defined the schema yourself, go to "createTypes" in gatsby-node.js. If you're using a plugin/theme, you can learn more here on how to fix field types:
```

websites/documentation/src/content/writing/pages.mdx
```diff
@@ -44,6 +44,16 @@ We support the following key-value pairs in the frontmatter section:
 
 - `wideLayout` (boolean)<Anchor name="wide-layout" />: use to make a page responsive and compatible for large viewport sizes. When used, the page can go into a two-column content space. For more information, see `<SideBySide>` [component](#side-by-side).
 
+- `showTimeToRead` (boolean)<Anchor name="show-time-to-read" />: use to show the reading time of a page. If the `setTimeToRead` frontmatter is not defined, gatsby will analyse the page and determine the reading time.
+
+- `setTimeToRead` (number)<Anchor name="set-time-to-read" />: use to set the reading time of a page. If defined, it replaces the automatically generated value form gatsby.
```
nkuehn, line 49:
```suggestion
- `setTimeToRead` (number)<Anchor name="set-time-to-read" />: use to set the reading time of a page. If defined, it replaces the automatically generated value from gatsby.
```

nkuehn (approved):
let's go, thank you. 

I marked one typo in the docs.   If you are at it, the release note could be a little more elaborate, e.g. with a link to the documentation or the names of the frontmatter.

PR #1250, opened 2022-03-04
Fix: set max-width for enums

### LINKED ISSUE #1031: layout: side by side container left column wider than 770
The `<SideBySide ....>`  layout wrapper that we use for API docs and also manually where needed does seem to not respect the maximum width of the left (main content) column.   I think it originally did but may err and it was always the case:

image

The goal of the layout was that the left hand column of the side by side is constrained to 770px.  Since the layout of the side by side is defined in a way that the right hand column is fixed and the left one fills (at least that's how it behaves),  it may be that we simply do not calculate the width of the right hand one properly to make sure the left hand one is effectively not getting wider than 770px.

### LINKED ISSUE #1242: Text takes entire width of page in wide layout as opposed to usual
**Current Content:** There was [an instance](https://docs.commercetools.com/api/projects/types#resourcetypeid) where the text takes the entire width of the page (in wide screens), while the rest of the content in the page was restricted to only half of the page, as shown below.

Screenshot 2022-02-25 at 09 42 51 

vs.

Screenshot 2022-02-25 at 09 43 06

Something to note, the first screenshot is a text in a RAML file, while the second one is handwritten. But, as I see other RAML generated content (like [TextInputHint](https://docs.commercetools.com/api/projects/types#textinputhint)) are also restricted to only half of the page.

timonrey:
closes #1242 
closes #1031


timonrey:
The snapshot says that there is a problem with the code example `An ApiEndpoint with Object request example`.
https://percy.io/commercetools-GmbH/docs-kit-smoke-tests/builds/16952135/changed/952698385?browser=chrome&browser_ids=23&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1706&widths=1706

But I can't see this problem in the build.

.changeset/afraid-lamps-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-api-docs': patch
+---
+
+Fix enum types that take the full width of a page
```
nkuehn, line 5:
```suggestion
- Enum types are not taking the full width of the page any more in wide layout pages
- The side column now has the correct, slightly wider, width in the wide layout to match the minimum width promised for body content. 
```

nkuehn (approved):
OK, good to go

PR #1266, opened 2022-03-29
Apply all remaining writing style improvements

### LINKED ISSUE #437: Writing Style Improvements
**this issue serves as a place to collect improvement ideas for the writing style linter rules. Please add ideas as comments**

Implementation will be done periodically.

timonrey:
closes #437 finally :)

This PR implements all suggested improvements from this comment: https://github.com/commercetools/commercetools-docs-kit/issues/437#issuecomment-1076389264

It also updates the styles from the `google` and `writegood` packages.


packages/writing-style/package.json
```diff
@@ -27,5 +27,8 @@
     "7zip-bin": "5.1.1",
     "node-fetch": "2.6.7",
     "shelljs": "0.8.5"
+  },
+  "devDependencies": {
+    "prettier": "2.6.1"
```
timonrey, line 32: I've installed prettier in order to use the two scripts for updating the google and writegood styles

packages/writing-style/styles/Google/Spacing.yml
```diff
@@ -4,5 +4,5 @@ link: 'https://developers.google.com/style/sentence-spacing'
 level: error
 nonword: true
 tokens:
-  - '[.?!] {2,}[A-Z]'
-  - '[.?!][A-Z]'
+  - '[a-z][.?!] {2,}[A-Z]'
+  - '[a-z][.?!][A-Z]'
```
timonrey, line 8: Fortunately, this new change from google does not cover the `.NET` case anymore, so we don't need to override it.

packages/writing-style/styles/commercetools/Acronyms.yml
```diff
@@ -97,7 +97,6 @@ exceptions:
   - OR
   # Specials and Brands
   - YYYY # to be able to describe date formats like YYYY-MM-DD
-  - ZEIT # the zeit.co service is trademarked in uppercased letters.
```
timonrey, line 100: We don't need to add `Vercel` somwhere, right?

nkuehn: No, there's no risk of misunderstanding with an acronym because it's not uppercased

.changeset/tough-olives-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/writing-style': patch
+---
+
+Applies all suggested writing style improvements.
```
nkuehn, line 5: Please list the actual changes in minimal form. Otherwise is hard to reconcile what to adjust in the content when the release is applied.

.changeset/tough-olives-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/writing-style': patch
```
nkuehn, line 2: I think this is minor. It's not just fixing things but actually improving the behavior.

packages/writing-style/styles/Google/Spacing.yml
nkuehn: Nice trick they do here, just assuming the last letter of a sentence is normally a lowercase alphabetical one.

packages/writing-style/package.json
nkuehn: Interesting, I previously thought it's sufficient to have a CLI in the top level package.json to be able to use it in a workspace context.

packages/writing-style/styles/commercetools/WordList-Suggestions.yml
```diff
@@ -0,0 +1,8 @@
+extends: substitution
+message: "Consider using '%s' instead of '%s' (case sensitive)."
+level: suggestion
+ignorecase: false
```
nkuehn, line 4: I think these very loose rules that are just ideas should be case insensitive.  e.g. the extension case would typically be lowercase anyways

nkuehn (changes requested):
see comments, plus:
 - allowlist vs. whitelist,  blacklist vs. denylist
 - please update the vale version, too to not have unnecessarily numbers of releases. We have not updated it for over a year and there have been a lot of improvements.  Also you want to be sure that the default Google rules are compatible.

packages/writing-style/styles/commercetools/DiscriminatoryWords.yml
```diff
@@ -0,0 +1,11 @@
...
+extends: existence
+message: "'%s' is a discriminating expression. Please use a different wording."
+ignorecase: true
+level: error
+tokens:
+  - whitelist(?:ed|)
+  - blacklist(?:ed|)
+  - master
+  - slave
+  - native
```
nkuehn, line 11: please move the existing "disable" here too and consider making this rather  a replacement rule like WordList because you can then suggest the recommeded alternative term.

nkuehn (approved):
just one thing suggested, please merge yourself after considering that change

packages/writing-style/styles/commercetools/DiscriminatoryWords.yml
timonrey: I considered it and my feeling was that in some cases it is difficult to suggest the right alternative since it depends strongly on the context and what the writer wants to express with the word. For the first two it is easy, yes. But I think for the other words it is more difficult.

Issue #1268, opened 2022-03-31
Apply new ct-design in the kit


timonrey:
Created issue out of this discussion:
https://github.com/commercetools/commercetools-docs/discussions/1726

Pls replace all the provided icons, banners and the new headline color in the kit.
Don't forget to update the cards on the hompage, too.

PR #1269, opened 2022-04-05
Replace banners and icons provided by the kit

timonrey:
is part of #1268 

This pull request implements the new icons and banners in the kit.

This PR prepares everything to change the designs in the websites of the docs repo and the appkit.


packages/gatsby-theme-docs/color-presets/index.js
```diff
@@ -16,9 +21,9 @@ module.exports = {
     relativePath: 'merchant-center-user-docs',
     value: require('./merchant-center-user-docs'),
   },
-  platformDeveloperDocs: {
+  tutorialDeveloperDocs: {
```
nkuehn, line 24: I do not clearly understand why you need to rename this.  Is it just because the files have different names now? 

Background: "tutorial" as a section may not even be sure to stay an own microsite.  it was a voluntary choice to make the naming more abstract.

nkuehn (changes requested):
I think renaming the use case platform developer docs to the more specific tutorials is not needed.   It's good enough as "dev docs that are not api docs" and you avoid the breaking change.

packages/gatsby-theme-docs/color-presets/index.js
timonrey: We don't need to rename it, I jut thought it would be clearer that way. But I will keep the name then 👍🏼

nkuehn (approved):

PR #1275, opened 2022-04-07
Design changes: use new headline color for page hero title

timonrey:
is part of #1268 
This was discussed with @Shecki


websites/documentation/gatsby-config.js
```diff
@@ -16,6 +17,7 @@ module.exports = {
   plugins: [
     ...configureThemeWithAddOns({
       websiteKey: 'documentation',
+      colorPreset: colorPresets.platformDeveloperDocs.key,
```
timonrey, line 20: I used the blue banner for the documentation website now (simply looks better :))

nkuehn: Yes, but...  Grey was reserved for our various open source offerings and the docs kit is one of them.  It's not a search indexed website but I would not want it to look like an actual commercetools product.

nkuehn (commented):
I don't think this is what the designs intended.  To me the content in the discussion does not put the headline font color _over_ the banner.  The Headline color is on white because it's for the page title etc.   

@Shecki  can you confirm that the font that is over the banner is not inverted white any more but the dark headline color? 
see here:  https://docskit-pr-1275.commercetools.vercel.app/documentation/

websites/documentation/gatsby-config.js
timonrey: OK. I thought since the documentation page should be presentable it could get some color... we might consider using one of the optional banners (both orange) for the documentation page. Or we leave it as it is. Just an idea.

Shecki:
@nkuehn 
yes, this was given by the marketing person.

nkuehn:
I checked the various color schemes and  they work.  It's just the "base" theme that I find really awkward and the contrast is horribly bad:  https://docskit-pr-1275.commercetools.vercel.app/docs-smoke-test/ 

Sorry to be a stubborn old man here but I think for the case of the grey background we need a better solution

image

Shecki:
@nkuehn ok, i will bring this up to the marketing designer

Shecki:
@nkuehn @timonrey 
Ok, he is fine with white.
The readability should have more prior!

Sorry for the chnages

timonrey:
It seems like there is a problem with the new documentation logo when you look into the snapshots. I can't reproduce this bug in chrome though. But on firefox, it looks pretty buggy.

nkuehn:
TBH the font rendering of the Logo looks pretty weird in Chrome, too.   I assume that is because it is referring to our company font but the company font is not embedded into the website at all yet. 

If we have to embed the font anyways, can we maybe do it right from the beginning and separate the Text from the icon (making the text HTML text)? 

Plus, the implementation of the SVG is a bit strange, too:  It has the text in a position outside the canvas and then translates it back into the place it is supposed to be. 

image

emmenko:
I feel like we need to have a chat about SVG files in general, as many of the SVG files we usually get (also in uikit and other places) contain a lot of "unnecessary markup". Not sure what software they use (Sketch?) to export them but it would be nice in general to get "clean" versions of the SVG files with only the minimal required data.

Happy to initiate this discussion and see if we can improve things. @Shecki not sure if you're the right person to talk about this or should we reach out to someone else?

---

About the logo. I agree with @nkuehn .
What is the reason for embedding text in it? There is also the issue with the font family of course.

nkuehn:
If Sketch it seems to be a known issue, at least some people chose to invest time into a blog post to vent their frustration: https://medium.com/sketch-app-sources/exploring-ways-to-export-clean-svg-icons-with-sketch-the-correct-way-752e73ec4694  

I've put a comment in the corporate design wiki page asking for where / how to obtain the legal official licensed webfonts.

packages/ui-kit/src/design-system.ts
```diff
@@ -258,6 +264,7 @@ export const typography = {
   fontFamilies: {
     primary: "'Roboto', sans-serif",
     code: "'Roboto Mono', monospace",
+    documentationLogo: 'TTCommonsPro-Rg, TT Commons Pro',
```
emmenko, line 267: What's the plan on using this new fonts? Should they replace `Roboto`?

Also, you need to actually import them otherwise they won't be applied. Example https://github.com/commercetools/commercetools-docs-kit/blob/a4cd66c10a346aacb03cc1e019d2aba009168c14/packages/gatsby-theme-docs/gatsby-browser.js#L8-L13

emmenko: Unless I'm missing some context here, I would remove this change for now and keep it as it was until there is a formal decision.

timonrey: Yes we decided to keep the documentation logo as it is and to implement that separately.

packages/ui-kit/src/design-system.ts
```diff
@@ -121,6 +123,8 @@ export const colors: ThemeColors = {
     link: customProperties.colorPrimary25,
     linkHover: customProperties.colorPrimary,
     linkNavigation: customProperties.colorInfo,
+    // Headlines
+    headlinePrimary: '#003037',
```
emmenko, line 127: Sorry to be picky but can't we use one of our existing Design System colors, like `colorAccent30`? Why introducing a new one?

cc @Shecki 

This is how it looks (left side) compared to this new color `#003037` (right side).

image

nkuehn: Thanks for the clarification Sven.  We'll have to talk about the direction of the Docs UI Kit and design system sometime. It's currently derived from the MC UI Kit but like in this example it seems that corporate design is moving independently (but there is no design system or UI Kit implementation available for corporate elements).  

For today good enough this way.

nkuehn (approved):

PR #1276, opened 2022-04-08
fix: generate SVGR components using unique IDs

emmenko:
If multiple SVG files contain the same IDs, and these files are imported and rendered within the same page, there SVGs are not rendered as expected because the IDs conflicts with each other. For example, all SVGs have the same color.

We can configure our SVGO plugin to use the `prefixIds` plugin that prefixes all IDs with the file name, making the IDs unique.

To prevent this from happening in the future, I extended the smoke tests to render all variations of our card banners.


package.json
```diff
@@ -13,7 +13,7 @@
     "build-packages": "preconstruct build",
     "build": "./scripts/build.sh",
     "vercel-build": "echo \"you're welcome - nothing to build\"",
-    "generate-icons": "yarn workspaces foreach -p --include @commercetools-website/docs-smoke-test,@commercetools-docs/ui-kit run generate-icons && prettier --write '**/generated/*.{js,ts,tsx}'",
```
emmenko, line 16: Using `workspaces foreach` has the downside of suppressing errors, which was the case here (the smoke tests command was failing to run).

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -538,6 +538,7 @@ exports.onCreateWebpackConfig = (
                     },
                   },
                 },
+                'prefixIds',
```
emmenko, line 541: The fix for the gatsby theme, when importing `.svg` files.

svgr.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
           },
         },
       },
+      'prefixIds',
```
emmenko, line 25: Fix for the generated icons within the kit repo.

websites/docs-smoke-test/package.json
```diff
@@ -24,5 +24,8 @@
     "gatsby-cli": "4.9.0",
     "react": "17.0.2",
     "react-dom": "17.0.2"
+  },
+  "devDependencies": {
+    "@svgr/cli": "6.2.1"
```
emmenko, line 29: Fix for the `generate-icons` command

websites/docs-smoke-test/src/@commercetools-docs/gatsby-theme-docs/overrides/markdown-components.js
```diff
@@ -1,18 +1,21 @@
...
-  CardBannerOssSvgIcon as CardBannerOSSIcon,
-  CardBannerMcDevsSvgIcon as CardBannerMcDevsIcon,
-  CardBannerMcDocsSvgIcon as CardBannerMcDocsIcon,
-  CardBannerPlatformSvgIcon as CardBannerPlatformIcon,
+  CardBannerApiSvgIcon as CardBannerApiIcon,
+  CardBannerDefaultSvgIcon as CardBannerDefaultIcon,
+  CardBannerMcSvgIcon as CardBannerMcIcon,
+  CardBannerOptional1SvgIcon as CardBannerOptional1Icon,
+  CardBannerOptional2SvgIcon as CardBannerOptional2Icon,
+  CardBannerOssSvgIcon as CardBannerOssIcon,
```
emmenko, line 9: These are the new card banners.

websites/docs-smoke-test/src/content/components/cards.mdx
```diff
@@ -328,14 +328,44 @@ Icons on **narrow** cards are aligned to the left of the card regardless of whet
...
+  <Card image={<CardBannerMcIcon />}>
+    Flat card with an image
+  </Card>
+  <Card image={<CardBannerOptional1Icon />}>
+    Flat card with an image
+  </Card>
+  <Card image={<CardBannerOptional2Icon />}>
+    Flat card with an image
+  </Card>
+  <Card image={<CardBannerOssIcon />}>
```
emmenko, line 346: This results in this:

image

nkuehn (approved):
Thank you!  

It would have taken ages for me to find that this is the root cause.

timonrey (approved):
Yes, very nice! 💯 
This will fix the problem we have with the cards on the docs homepage

PR #1278, opened 2022-04-12
Allow "Cloud Native Computing Foundation" in writing style rules

timonrey:
See https://github.com/commercetools/commercetools-docs/pull/1949#issuecomment-1096762389


nkuehn:
I think this is rather a case for an exclusion comment in the respective content file. the cncf is not something we write about regularly and the way its excluded here makes it really hard to add other maybe relevant special cases.

timonrey:
Thanks, I will exclude it in the content file then.

PR #1285, opened 2022-04-28
Move `globalNotification` configuration to siteMetadata

### LINKED ISSUE #1274: plugin properties mixed into content page data nodes cause builds from stale gatsby cache data.
We recently had a situation where changes in the gatsby configuration were not actually propagating into all pages but only into some, seemingly random selection of pages.   The configuration that was changed was the global site notification.  That is mixed into the ContentPage gatsby data node.  It is also becoming part of the digest of that node.  But if a node's parent does not change, it's not even attempted to be recalculated (at least that is my understanding of cached gatsby builds and the node hierarchy).  The parent is the MDX node which does not change if a page was not touched. 

Here's my comment in the respective discussion in the content repository: 

> My suspicion for a root cause is that we're mixing a plugin configuration key (the global alert) into the Gatsby data node for the content page. But if the parent node of that content page node is not changing (the parent is the MDX node whose parent is the file node of the mdx file.), it does not trigger an update of the content page node, hence stays in the cache. If that is true, we learnt again that we overengineered the whole graphQL facade pattern in gatsby beyond our full understanding of gatsby. You cannot add information to a node that is not derived from information on other nodes that are parent nodes. Global site data would have to have read from a site data graphQL query and not mixed into the page data node.

TODO: 

1.  verify the assumption
2. discuss and decide whether to take this as the opportunity to completely remove our custom data facades or whether to just fix the bug by refactoring the data layout so that the information becomes site data and not page data.

timonrey:
closes #1274

This pull request moves the global notification configuration to the `siteMetadata` and adds documentation for the `siteMetadata` configuration options.

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -31,6 +31,10 @@ module.exports = (themeOptions = {}) => {
       author: 'commercetools',
       productionHostname,
       betaLink: '.',
+      globalNotification: {
+        notificationType: false,
+        content: false,
+      },
     },
     plugins: [
       /**
--- a/packages/gatsby-theme-docs/src/layouts/content.js
+++ b/packages/gatsby-theme-docs/src/layouts/content.js
@@ -31,7 +31,7 @@ const LayoutContent = (props) => {
   return (
     <LayoutApplication
       websitePrimaryColor={props.pageData.websitePrimaryColor}
-      globalNotification={props.pageData.globalNotification}
+      globalNotification={siteData.siteMetadata.globalNotification}
     >
       <LayoutSidebar
         {...layoutState.sidebar}
@@ -61,11 +61,13 @@ const LayoutContent = (props) => {
             allowWideContentLayout={props.pageData.allowWideContentLayout}
           >
             <LayoutGlobalNotification>
-              {props.pageData.globalNotification && (
+              {siteData.siteMetadata.globalNotification.content && (
                 <GlobalNotification
-                  type={props.pageData.globalNotification.notificationType}
+                  type={
+                    siteData.siteMetadata.globalNotification.notificationType
+                  }
                 >
-                  {props.pageData.globalNotification.content}
+                  {siteData.siteMetadata.globalNotification.content}
                 </GlobalNotification>
               )}
             </LayoutGlobalNotification>
@@ -115,10 +117,6 @@ LayoutContent.propTypes = {
   }).isRequired,
   pageData: PropTypes.shape({
     title: PropTypes.string.isRequired,
-    globalNotification: PropTypes.shape({
-      notificationType: PropTypes.oneOf(['info', 'warning']).isRequired,
-      content: PropTypes.string.isRequired,
-    }),
     websitePrimaryColor: PropTypes.string.isRequired,
     beta: PropTypes.bool.isRequired,
     isGlobalBeta: PropTypes.bool.isRequired,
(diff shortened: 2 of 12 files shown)


packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -31,6 +31,10 @@ module.exports = (themeOptions = {}) => {
       author: 'commercetools',
       productionHostname,
       betaLink: '.',
+      globalNotification: {
+        notificationType: false,
+        content: false,
```
nkuehn, line 36: I don't think it's ideal to make the default for string properties "false", esp. if the data is exposed on GraphQL you get into ugly errors because gatsby fails to build the schema.  Either give it no default (not set or null) or one that has the right type

packages/gatsby-theme-docs/src/layouts/content.js
```diff
@@ -61,11 +61,13 @@ const LayoutContent = (props) => {
             allowWideContentLayout={props.pageData.allowWideContentLayout}
           >
             <LayoutGlobalNotification>
-              {props.pageData.globalNotification && (
+              {siteData.siteMetadata.globalNotification.content && (
```
nkuehn, line 64: related to the above discussion concerning defaults:  the code is assuming "globalNotification" is always set but is just checking for "truthyness".

websites/documentation/src/content/configuration/theme.mdx
```diff
@@ -37,6 +37,28 @@ The Docs-Kit theme also receives its plugin configuration through this standard
...
 - `enableCanonicalUrls` (_optional_): indicates that the website should use canonical URLs, pointing to the `docs.commercetools.com` domain. **Default: `true`**
 
+## Available Options for the Site Metadata
+
+
+- `title` (**required**): Defines the title of the website.
+- `description` (**required**): Defines the description of the website.
+- `betaLink` (_optional_): Configures the link for _beta button_.
+- `globalNotification` (_optional_, object): Activates a global banner at the top of every page within the website. It contains two configuration options:
+  `notificationType` (String, either **info** or **warning**) defines the type and color of the banner. `content` (String) defines the text within the banner.
```
nkuehn, line 47:
```suggestion
    -  `notificationType` (String, must be either `info` or `warning`) defines the type and color of the banner. 
    -  `content` (String) defines the text within the banner.
```

websites/documentation/src/content/configuration/theme.mdx
```diff
@@ -37,6 +37,28 @@ The Docs-Kit theme also receives its plugin configuration through this standard
 
 - `enableCanonicalUrls` (_optional_): indicates that the website should use canonical URLs, pointing to the `docs.commercetools.com` domain. **Default: `true`**
 
+## Available Options for the Site Metadata
```
nkuehn, line 40: did we really not document this anywhere yet? Thanks!

nkuehn (approved):
Just by reading the code, I did not run or test it againts edge cases.

websites/documentation/src/content/configuration/theme.mdx
timonrey: We didn't 😅

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -31,6 +31,11 @@ module.exports = (themeOptions = {}) => {
       author: 'commercetools',
       productionHostname,
       betaLink: '.',
+      globalNotification: {
+        isActive: false,
+        notificationType: 'info',
+        content: '',
```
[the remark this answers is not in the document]
nkuehn: That's a pretty good idea, it makes it easier for users and the config gets typesafe. 

Just a nitpick: we're not using the  "isXxxx"  naming for booleans in the product elsewhere,  just `active`.

PR #1294, opened 2022-05-09
[Improvement idea] Add navigation to other microsites

timonrey:
At the moment the docs-smoke-test has cards on the index page that contain links to the other microsites which is useful for the deployed website since you can quickly navigate to the smoke test or documentation you need. I suggest to add these cards to all microsites (especially to the new documentation website) so that you can use the link in the index page to get to the smoke test instead of changing the URL.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "clean": "yarn workspaces foreach -p --include '@commercetools-website/*' run clean",
-    "start": "yarn workspace @commercetools-website/docs-smoke-test start",
+    "start": "yarn workspace @commercetools-website/documentation start",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "clean": "yarn workspaces foreach -p --include '@commercetools-website/*' run clean",
-    "start": "yarn workspace @commercetools-website/docs-smoke-test start",
+    "start": "yarn workspace @commercetools-website/documentation start",
```
nkuehn, line 12: Why would you want to start all four sites at once?  I could understand building but starting?

nkuehn (approved):

PR #1300, opened 2022-05-16
Proof of Concept: replace GraphQL facade layer with fields on mdx nodes

nkuehn:
**final finding**:  It does not get any better overall - the facade has significant advantages in queryability and sorting. Also, error handling does not get better by directly addressing the MDX node because that node is disappearing on mdx errors anyways. 

Superseded by improving the error handling again: https://github.com/commercetools/commercetools-docs-kit/pull/1301  

----

**Status**: Works in dev mode but the build fails, sort and filter statements on fields work in dev mode but not in the build. 

An ugly workaround to filtering would be the glob for "fileAbsolutePath" `{fileAbsolutePath: {glob: "**/src/releases/**"}}` in all the queries but it's sort of really ugly and does not solve sorting the release notes e.g. for the RSS feed. 

If it's really not possible to filter for fields in the build we could keep the facade API but try to not dynamically resolve the MDX body but rather query it from the node parent in the page queries. Maybe this makes the errors easier to catch and the schema definition could be simpler to understand if it does not have the dynamic resolution. 

----

addresses #539, maybe addresses build performance, reduces architectural complexity for easier onboarding and maintenance. 

We struggle with the dynamic resolution of the MDX body in our custom graphQL facade for long - the error behavior is hard to control and we still live with the suspicion that something in the setup causes every MDX compilation to happen twice.  @timonrey  and me decided to gradually reduce the architectural and code design complexity to more straightforward / OOTB gatsby features even if it incurs some "ugliness" or repetition.     

This PoC checks whether it's feasible to remove our custom graphQL facades on the example of the release notes (they are more complex than the content pages)

It creates the cleaned fields with defaults as before but sets them as nodeFields on the mdx node directly instead of creating another layer of data nodes. 

The intended effect has happened: the dev server is not completely crashing any more on MDX syntax errors! It's not really good error behavior yet (no useful output) but a much better starting point because the dev server is recovering cleanly at all.  (but the build fails, see above)

Also proves it's possible with relatively overseeable code mod because graphql has aliases that made the data structure only deviate in the nesting into "fields".

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -261,28 +261,34 @@ module.exports = (themeOptions = {}) => {
           feeds: [
             {
               serialize: ({ query: { site, allReleaseNotePage } }) => {
-                return allReleaseNotePage.nodes.map((node) => {
+                return allReleaseNotePage.nodes.map(({ fields }) => {
                   return {
-                    ...node,
-                    url: `${site.siteMetadata.siteUrl}${node.slug}`,
-                    guid: `${site.siteMetadata.siteUrl}${node.slug}`,
+                    ...fields,
+                    url: `${site.siteMetadata.siteUrl}${fields.slug}`,
+                    guid: `${site.siteMetadata.siteUrl}${fields.slug}`,
                   };
                 });
               },
               query: `
               {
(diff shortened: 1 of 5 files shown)


nkuehn:
Not doing, turned out to not get any better overall.

PR #1305, opened 2022-05-18
Wait for percy to take snapshots after mermaid diagrams are loaded

### LINKED ISSUE #1277: Percy shows changes because rss feed or mermaid diagram are not fully loaded yet
At the moment, many Percy snapshots show changes because the mermaid diagram or RSS feeds aren't loaded when the snapshot is taken. This is quite annoying because you always have to look into the snapshots to tell if there are actual changes or just the visual change of not seeing the diagram or RSS feed yet.

We need to somehow tell cypress/percy to take the snapshot after the components are loaded and visible.
Some adjustments in the cypress tests could solve this problem.
See more information here:
https://github.com/cypress-io/cypress/issues/5275

timonrey:
closes #1277 

This pull request adds a timeout before capturing the snapshot for the mermaid-diagrams page. This gives the mermaid diagrams 2 more seconds to load before taking the snapshot.
I tried it three times and the percy checks were always green.

Link to the documentation:
https://docs.percy.io/docs/percy-snapshot#configuration


.percy.js
```diff
@@ -16,6 +16,10 @@ module.exports = {
         include: /(.*-smoke-test.*)\/components\/rss-feeds\/index\.html/,
         waitForSelector: 'table[data-table-name="rss-feed"]',
       },
+      {
+        include: /(.*-smoke-test.*)\/components\/mermaid-diagram\/index\.html/,
+        waitForTimeout: 2000,
```
nkuehn, line 21: @timonrey  I don't think you need to guess with a timeout here.  You can use the same approach as with the RSS feed and wait for a selector.  
For example  `#body-content svg.statediagram`   try here: https://commercetools-docs-kit.vercel.app/docs-smoke-test/components/mermaid-diagram

timonrey: Ah yes, thanks. I didn't see the that there is a class defined there. 👍🏼

nkuehn (approved):

PR #1306, opened 2022-05-23
Drop in new corporate font 1:1 for design review

nkuehn:
New corporate font for review by design to evaluate whether it's worth moving towards it or not. 

The font is commercially licensed on a pageviews model, so it's statically pulled into the repo but the loading mechanism is kept as distributed by the font provider. 

Personally not impressed by the readability yet


timonrey:
I actually quite like the new font. It looks a bit more fancy and cleaner than the old one. I think I understand what you mean with the readability. But we'll probably just have to get used to it first. But that's just my opinion :)

nkuehn:
@timonrey  thanks for the feedback.  Readability just needs to be addressed by fine tuning font weights and spacing to match the dimensioning of the new font.  Can be done interactively with a designer, too because we have a proper design system with central places to adjust line height, font weight, font size etc.

nkuehn:
closing as obsolete, we should better align with the product font than with the marketing typography because the latter changes more frequently. The product has chosen "Inter V" now as the new long term font and I hope it's a better fit than TT which was tried here and did not really work for the documentation

Issue #699, opened 2020-10-07
UX: consider to copy address when clicking on Section Anchor icon (instead of navigating)


nkuehn:
Given that Chrome will remove the default display of the path in the address bar rolling out gradually starting with Chrome 86 (now) it can be expected the the whole concept of the address bar to see the full URL will further deteriorate. 

Here's how the new browser bar will show addresses (just the host, no matter where you are). 
![image](https://user-images.githubusercontent.com/4006139/95324257-bd14a080-089f-11eb-8555-4bbe3fdf32f3.png)

For example, the google cloud documentation is not actually navigating to the anchor when the anchor icon is clicked but is only copying it the to clipboard and that's it.   
Example:  https://cloud.google.com/compute/docs/instances/create-start-instance#publicimage

timonrey:
Idea approved by @Shecki

nkuehn:
Implementation proposal:  Exactly like it's done in the code block component.  Click on it and then there is a tooltip popping up confirming it was copied.  Not sure if making it re-usable is worth it but worth a look if e.g. the clipboard logic is tricky

PR #1314, opened 2022-06-06
feat: Copy section address when clicking section anchor icons

gabriele-ct:
Fixes: #699


packages/ui-kit/src/components/markdown.tsx
```diff
@@ -298,36 +300,64 @@ const TypographyContainer = styled.div`
...
+const withCopyToClipboard =
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (Component: React.ComponentType) => (props: any) => {
+    const [isCopiedToClipboard, setIsCopiedToClipboard] = React.useState(false);
+    const handleCopyToClipboardClick = (event: React.SyntheticEvent) => {
+      event.preventDefault();
+      const sectionUrl = `${window.location.href.split('#')[0]}#${props.id}`;
+      copyToClipboard(sectionUrl);
+      setIsCopiedToClipboard(true);
+      setTimeout(() => setIsCopiedToClipboard(false), 1500);
```
nkuehn, line 326: Nitpick: the timeout based implementation works but has am edge case when you keep the pointer hovering over the item longer - the text changes back then, which seems erratic. 

Did you try using the `onClose` callback of the tooltip UI-Kit component to reset the state?  It's not in the "knobs" UI of the UI-Kit storybook but in the readme. 

https://uikit.commercetools.com/?path=/story/components-tooltips--tooltip

packages/ui-kit/src/components/markdown.tsx
```diff
@@ -298,36 +300,64 @@ const TypographyContainer = styled.div`
...
-  );
-};
+          > * + * {
+            margin: 0 0 0 ${dimensions.spacings.s};
+          }
+        `}
+      >
+        <span>{props.children}</span>
+        <Tooltip
+          title={isCopiedToClipboard ? 'Copied' : 'Copy to clipboard'}
```
nkuehn, line 350:
```suggestion
          title={isCopiedToClipboard ? 'Copied' : 'Copy link to clipboard'}
```
just tuning the language

packages/ui-kit/src/components/markdown.tsx
gabriele-ct: @nkuehn works like a charm! Thank you, I guess we could implement something similar for the `code-block` component which still uses the timeout based approach: [`packages/ui-kit/src/components/code-block.tsx](https://github.com/commercetools/commercetools-docs-kit/blob/65b2345e6004dc7b69dbe76e3ede794ca5dc0906/packages/ui-kit/src/components/code-block.tsx#L213)
I'll create a ticket for that

timonrey (approved):
Thanks! Looks great! 🚀 🙌🏼

PR #1315, opened 2022-06-06
fix(method): Display API method description within the ApiEndpoint method container

gabriele-ct:


timonrey (approved):
Looks good! Thank you 🙌🏼

Issue #1260, opened 2022-03-16
Provide component for multi-column layout


stmeissner:
We have [a long list of items](https://docs.commercetools.com/api/projects/types#list-of-customizable-data-types) that 
* requires scrolling to see all items on screen resolution 1440*900px and that 
* wastes horizontal space since the list items are pretty short

We would like to present the list items in a two-, or three-column layout, but currently the Docs-Kit does not provide a suitable component for it.

Pleas add a component for [multi-column layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Columns) content.

nkuehn:
Idea:  base an initial PoC proposal on the layout behavior of the child section navigator, maybe even including the border styling.   Then show to design.

gabriele-ct:
I've created a POC simple multi column component. I think the next step is to gather feedback from UX.
The component is visible here: https://docskit-pr-1316.commercetools.vercel.app/docs-smoke-test/components/multi-column-layout

nkuehn:
Nice, thanks!  I have no opinions about the border or box but I think given that the font size is body content (as opposed to the child section nav which has a smaller font) two columns would be better readable overall.  
Would then be one column in the mobile viewport.

Shecki:
Hey, please do not go for this approach.
@gabriele-ct thank you for all the effort too build this.

- even though there is a long list, nowadays the user is not "afraid/tired" of scrolling
- a user scans to the content
- with the approach the contact is two crowded and heavy, there is no "break" for the eye
- a head map of user eye view is always strong on the left side and gets less to the right side
- with a long list we also add way more whitespace, is an ästhetic effect that gives the feeling of clean design but also saves the user from overloaded content.
He can focus on what's important

Please keep the list as it is.
If there is a list that really takes 5 sec. to scroll then we need to log on a completely new approach.
for solving situations like this.

gabriele-ct:
Hi @Shecki ,
thanks for your valuable input 👏 . I'll leave the draft PR open for now and we'll take a decision next week when @nkuehn is back 👍

nkuehn:
I have no emotions on the topic, it was brought up by content authors and I just threw the idea of a column layout into discussion.

FFawzy:
lets look for a different UI/UX solution

zbalek:
note from Nicola Molinari - saving it here for future reference:

_

[...]

PR #1316, opened 2022-06-07
feat(components): add component for multi column layout

gabriele-ct:
fixes: #1260


gabriele-ct:
I've created a POC simple multi column component. I think the next step is to gather feedback from UX.
The component is visible here: https://docskit-pr-1316.commercetools.vercel.app/docs-smoke-test/components/multi-column-layout

gabriele-ct:
@nkuehn thanks for your feedback.
I've updated the PR according to your comment
https://github.com/commercetools/commercetools-docs-kit/issues/1260#issuecomment-1153671885
Not sure about the next steps. Shall I assign it to someone specific for UX review or just go ahead adding `changeset` and make it a "proper" PR ?

PR #1321, opened 2022-06-14
Fix a vale issue with JSX in MDX and do general vale maintenance

nkuehn:
Fixes an issue with the writing style checker not treating JSX blocks in MDX as code and throwing errors that are not correct.   The "BlockExclude" is limited to self-closing tags to make sure content inside e.g. "<SideBySide>" tags is still checked. 

In addition
 - upgrade vale version
 - support the now available arm native binaries of vale 
 - fix various issues in our test websites that do not comply to the full style check
 - change the way we run the writing style check to how it's documented (not via jest) and check only the websites MDX content.


timonrey (commented):
Fails now because of the `Running linters and tests` step. Do we still need that?

nkuehn:
> Fails now because of the `Running linters and tests` step. Do we still need that?

Sorry, i forgot to adjust the way the CI is calling the step.  That's where calling it through jest came in more consistent btw.

timonrey (approved):
Thank you! Looks good now 👍🏼

PR #1326, opened 2022-06-16
Update gatsby-theme-code-examples add-on to support jsx files

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=6bEvTyh)

## Summary

This PR extends the allowed list of file types the `gatsby-theme-code-examples` Gatsby add-on currently support to include files with `.jsx` extension (Javascript files with React components syntax).

## Description

When using this add-on in the Custom Application docs repository, we found that code snippets with `.jsx` file extension were not being included.
After taking a look at this add-on source code, we realized that there is an allow list of the files mime types that it supports and this kind of files were not included.

This file extension is commonly used when developing React components using Javascript (when using Typescript, the `.tsx` file extension is used which is already included in the allowed list).


timonrey (approved):
Thank you for the contribution! 🚀

PR #1329, opened 2022-06-16
fix(raml transformer): hide fields with markDeprecated property from api generated docs

gabriele-ct:
Fixes: commercetools-docs#2051

### DIFF
--- a/.changeset/rotten-tomatoes-divide.md
+++ b/.changeset/rotten-tomatoes-divide.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-transformer-raml': minor
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/rotten-tomatoes-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-transformer-raml': minor
```
timonrey, line 2: Since it is more a bugfix than a new feature, I would say it's more a patch than a minor change

gabriele-ct: I agree 👍

timonrey (approved):
Good for now, thanks!

PR #1331, opened 2022-06-20
feat(api examples): display code examples when available in examples node

gabriele-ct:


api-specs/api/resources/custom-objects.raml
```diff
@@ -71,8 +71,17 @@ post:
       200:
         body:
           application/json:
-            example: !include ../examples/custom-object.example.json
             type: CustomObject
+
+      202:
```
nkuehn, line 76:
```suggestion
      201:
```

packages/gatsby-theme-api-docs/src/hooks/use-api-resources.js
```diff
@@ -28,6 +28,11 @@ export const useApiResources = () => {
             applicationjson {
               type
               builtinType
+              examples {
+                key
+                strict
```
nkuehn, line 33:
```suggestion
```
the "strict" information is not necessary for the frontend, it's just indicating the RAML parser and validators that the example is supposed to be validated against the schema or not (for example allowing for pseudo-code examples that are not validated)

packages/gatsby-transformer-raml/src/utils/resource/examples-to-array.js
```diff
@@ -0,0 +1,11 @@
+function examplesToArray(examples) {
```
nkuehn, line 1: Are you sure this does not already have an implementation available in the codebase, maybe a more abstract one or in the types code?

nkuehn (commented):
Thank you, I think this is actually going in the right direction but not solving the users' problem yet.

.github/workflows/main.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           path: |
             .cache
             public
-          key: ${{ runner.os }}-build-v1-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-build-v3-${{ hashFiles('**/yarn.lock') }}
```
nkuehn, line 46: why this?

nkuehn (approved):
Thank you! It was a journey but looking back it turned out to be a great first task to learn the docs stack including the API docs generation.  You touched nearly all layers of the stack here successfully.

.github/workflows/main.yml
gabriele-ct: As I said, I had some cache issue during github build, mainly related with graphql schema changes. Changing the cache key, the problem disappeared.

PR #1333, opened 2022-07-04
feat(api responses): display response code in the header of the response code block

gabriele-ct:


api-specs/test/api.raml
```diff
@@ -271,6 +271,11 @@ uses:
             body:
               type: arrays.ArrayTestType
               example: !include examples/array-test-type.json
+          201:
```
nkuehn, line 274: We could improve the "smoke-test-ish-ness" of this endpoint by using not only examples that are referred to "locally" at the endpoint but also omitting them in some (like our actual API spec mostly does) and having the kit render the generic example of that type.   

I would not hesitate to even add a fourth 2xx response for that case and making the description explicity explain what it is targeting  (also here your just copied "ArrayTestType  is used because it is big and long." instead of describing why the same resonse and example is there another time)

packages/gatsby-theme-api-docs/src/components/resource/method/request-response-examples.js
```diff
@@ -46,13 +48,9 @@ const RequestResponseExamples = (props) => {
       {responsesCodeExamples.map((codeExample) => {
```
nkuehn, line 48: since we're here: can we rely on the responses being ordered numerically by code?

packages/gatsby-theme-api-docs/src/components/resource/method/request-response-examples.js
```diff
@@ -21,6 +22,7 @@ const RequestResponseExamples = (props) => {
 
         if (apiType.examples) {
           responsesCodeExamples.push({
+            code,
             typeDisplayName,
             value: apiType.examples[0].value,
           });
```
nkuehn, line 28: This logic still looks to me like we're not actually using the example that is directly referred to in the API.   It's just pulling the examples from the type.   

I suggest to first harden that smoke test endpoint (the one named "artificially complex") with examples and modeling that have the respective use case described in the actual examples JSON and descriptions before proceeding to make sure we're not running circles.

nkuehn (commented):
TL;DR:  Fixes one aspect (can't see what example refers to what response code) but does not address that "local" examples at the endpoint are simply ignored.  For that we need to touch the data processing side of things.

api-specs/test/api.raml
gabriele-ct: Yes, I've added an additional 2xx response  to cover the scenario and added a more explicit description

packages/gatsby-theme-api-docs/src/components/resource/method/request-response-examples.js
gabriele-ct: Yes, you are correct, as you said [here](https://github.com/commercetools/commercetools-docs-kit/pull/1333#pullrequestreview-1027864950) The code, at the moment, is ignoring the `examples` node. This means that currently we always render the example from the type.
This PR is meant to just add the response code to the header of the code example block, it doesn't add any logic regarding the handling of explicit examples.

PR #1334, opened 2022-07-05
Provide component shadowing for the avatar component

timonrey:
This pull request implements the new overridable avatar component. This is needed in order to override the component in the learning-tech websites to display the avatar (or not, if the user is not logged in yet).

### DIFF
--- a/packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
+++ b/packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
@@ -13,7 +13,7 @@ import {
   Icons,
 } from '@commercetools-docs/ui-kit';
 import { SearchDialog, SearchInput, Overlay } from '../../components';
-import Avatar from '../../overrides/avatar';
+import PlaceholderAvatarArea from '../../overrides/avatar';
 
 const SearchIcon = createStyledIcon(Icons.SearchSvgIcon);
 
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -13,6 +13,7 @@ import {
   Icons,
 } from '@commercetools-docs/ui-kit';
 import { SearchDialog, SearchInput, Overlay } from '../../components';
+import Avatar from '../../overrides/avatar';
```
gabriele-ct, line 16: I see we often use a `Placeholder` naming for these empty components.
See [here](https://github.com/commercetools/commercetools-docs-kit/blob/80b8e988f87e2dad0a788f5f36b93971504729fd/packages/gatsby-theme-docs/src/layouts/content.js#L9)

So my suggestion would be to import it as
```
import PlaceholderAvatar from '../../overrides/avatar';
```

or, even better
```
import PlaceholderAvatarArea from '../../overrides/avatar';
```

Today we might just want to add an avatar icon, so it makes sense to call it `Avatar` but a more generic name would add some flexibility.
`AvatarArea` could help us to identify the location, but keeping the flexibility to add anything in it.
e.g. "Add a logout button in the avatar area" sounds better than "Add a logout button in the avatar"

Anyway, it's just naming, up to you if you want to change it or not.

PR #1359, opened 2022-08-26
Introduce learning plugins to the kit

timonrey:
fix commercetools/learning-tech#161

The plugins are now configured and are providing all the functionalities as expected.
Two things we still need to consider:
- We still need to add the two dependencies from the ui-kit separately, otherwise you get an error.
- The import statement for the quiz component in the mdx file is also still needed to show the quiz in the page.


gabriele-ct:
@timonrey and @FFawzy 
I managed to have a successful build by setting the plugin config vailes manually (not sure why the environment variable set in Vercel are not visible).
Apart from this, using an incognito window I get to the login screen but after the login I can't see any auth0 cookie being set and it looks I'm logged out even though the calls to auth0 are successful.
I try to believe that has something to do with gatsby SSR, we might try to use the login popup instead.
Anyway, the bottomline is that I've not been able to diagnose the issue:
- auth0 side all seems correct
- I can't see any error reported  client side

That leave out gastby setup, but I'm running out of ideas

FFawzy:
I've spent the day today trying to investigate this and I finally found the problem
2 actually

1- there is a problem with the redirect of the avatar component 
if you run the doc-kit locally from the root folder and try to login, the login process succeeds but not able to finish the last piece of the process which is redirection to the quiz section so it exits.
but if you run the microsite (docs-smoke-test) by itself .. the login works and redirection works as expected
so there is some weird Gatsby behaviour here and it needs more investigation 
(I did not have the time to look at this today) 

2- second problem is CORS issue, our API is not handling the preflight requests and not setting the right CORS settings so I'm working on a fix for this and will create a pull request in the learning-tech repo

websites/docs-smoke-test/gatsby-config.js
```diff
@@ -1,5 +1,11 @@
 const isProd = process.env.NODE_ENV === 'production';
 const shouldEnableSearch = process.env.ENABLE_SEARCH === 'true';
+
+if (process.env.NODE_ENV === 'development') {
+  require('dotenv').config({
+    path: `.env`,
+  });
+}
```
emmenko, line 8: You should be able to use `.env.development` and Gatsby will load these env variables automatically. There should be no need to manually load them.

Can you double check? https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/

websites/documentation/gatsby-config.js
```diff
@@ -20,5 +20,12 @@ module.exports = {
       additionalPrismLanguages: ['java', 'scala', 'csharp', 'swift', 'php'],
       addOns: ['@commercetools-docs/gatsby-theme-code-examples'],
     }),
+    {
+      resolve: '@commercetools-docs/gatsby-theme-sso-ui-kit',
+      options: {
+        auth0Domain: 'commercetools-professionals.eu.auth0.com',
+        auth0ClientId: 'nZwQWPaoTq8IIk67VV8o0Ska9lGp4hnW',
```
emmenko, line 27: Is that ok to hardcode them?

websites/docs-smoke-test/package.json
```diff
@@ -18,8 +18,12 @@
     "@commercetools-docs/gatsby-theme-code-examples": "19.3.0",
     "@commercetools-docs/gatsby-theme-constants": "19.3.0",
     "@commercetools-docs/gatsby-theme-docs": "19.4.0",
+    "@commercetools-docs/gatsby-theme-learning": "1.0.0",
+    "@commercetools-docs/gatsby-theme-sso-ui-kit": "1.0.0",
     "@commercetools-docs/ui-kit": "19.3.0",
+    "@commercetools-uikit/avatar": "15.2.1",
     "@commercetools-uikit/icons": "^15.0.0",
+    "@commercetools-uikit/radio-input": "15.2.1",
```
emmenko, line 26: Are these deps required?

emmenko (commented):
Just had a quick look 🙂

websites/documentation/gatsby-config.js
timonrey: This is just for testing...will be removed as soon as the env variables work as expected.

websites/docs-smoke-test/package.json
timonrey: In the initial implementation of the learning plugin, I had to extra add those dependencies to make it work. But I will double check once we've solved the other issues

websites/docs-smoke-test/gatsby-config.js
FFawzy: I was actually reading this doc too and it says [here](https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/#:~:text=For%20example%2C%20when%20configuring%20a%20plugin%20in%20gatsby%2Dconfig.js%3A) that this is an example of how to provide keys to plugins which is exactly our use case and they had to manually load them ... this is why I did so

it is still failing in the [build](https://github.com/commercetools/commercetools-docs-kit/runs/8151852021?check_suite_focus=true#step:9:231) so this might actually mean that its not able to load the production variables too so it might be even worth to try their code snippet
`require("dotenv").config({
  path: `.env.${process.env.NODE_ENV}`,
})`

CC: @timonrey

timonrey:
Unfortunately, the env variables still don't seem to work on production...
https://github.com/commercetools/commercetools-docs-kit/runs/8151852021?check_suite_focus=true#step:9:229
I will hardcode them for now to be able to continue.

websites/docs-smoke-test/src/content/components/quiz.mdx
```diff
@@ -0,0 +1,9 @@
+---
+title: Quiz component
+---
+
+import { Quiz } from '@commercetools-docs/gatsby-theme-learning'
```
FFawzy, line 5: do you want to check what's wrong with the shortcodes and why you have to import it here?
or you want to do it in a separate PR ?

gabriele-ct (approved):
Thank you @timonrey 💯

FFawzy (approved):
🎊 🍻

emmenko:
Niiiice! 🥳

PS: when I click on logout, I get redirected to the root website but I would expect to be redirected to the page I was.

websites/docs-smoke-test/src/content/components/quiz.mdx
timonrey: Will be handled in a separate PR 👍🏼

PR #1365, opened 2022-09-09
[Self-learning] configure learning plugin with addOns

timonrey:
closes https://github.com/commercetools/learning-tech/issues/196


gabriele-ct (approved):
👏 thank you for fixing this!

PR #1367, opened 2022-09-15
Allow regex properties with any type

### LINKED ISSUE #1357: Regex properties currently only work for string types
I would like to render a property whose key is only defined with a regular expression like demonstrated in the [API smoke test](https://commercetools-docs-kit.vercel.app/api-docs-smoke-test/types/general#regexpropertytesttype).

The respective RAML spec looks like this:

```
properties:
  /^[a-z]{2}(-[A-Z]{2})?$/:
    type: string
```
The difference to my use case is that the type of my property is not `string`, but some object:

```
properties:
  /^[a-z]{2}(-[A-Z]{2})?$/:
    type: SearchKeyword[]
```
That causes following error on the page:

> Error in function NameType in ../../node_modules/@commercetools-docs/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js:58
>
> typeToRender.type.toLowerCase is not a function

Could you please add support for data types other than string also?

timonrey:
closes #1357

### DIFF
--- a/.changeset/brown-bears-unite.md
+++ b/.changeset/brown-bears-unite.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-api-docs': patch
+---
+
+Allow Regex properties to be used with any type, not only string.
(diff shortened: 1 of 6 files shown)


gabriele-ct (approved):
Super 👍

PR #1372, opened 2022-10-10
Migrate to new UIKit design tokens

emmenko:
In UIKit `v15.3.0` we refactored the internal usage of design tokens to provide better support for theming.
This was the first step to enable the work in the design system and MC redesign.

In short, we rely on CSS variables to define the design tokens and to override them (when defining a new theme).

In theory all design tokens are backwards compatible but in the DocsKit we are using our own abstraction on top of it, so we need to adjust certain things.
For instance, we don't need to use Emotion theming anymore and instead rely on using and overwriting CSS variables. This makes the setup simpler.

The changes here are mostly about replacing Emotion theming with CSS variable overrides.


tsconfig.json
```diff
@@ -1,14 +1,15 @@
...
-    "target": "es2019",
     "jsxImportSource": "@emotion/react",
+    "lib": ["ESNext", "DOM"],
     "module": "ESNext",
     "moduleResolution": "Node",
-    "esModuleInterop": true,
     "strict": true,
     "skipLibCheck": true,
-    "allowJs": true
+    "target": "ES2022"
```
gabriele-ct, line 12: shall we concern about browser compatibility here?

gabriele-ct (approved):
to the best of my knowledge all seems fine 👍

gabriele-ct (approved):

nkuehn:
Can one of you that has reviewed this in detail add the changeset file with the respective compatibility?  I can only guess whether this is minor / major / patch. Thanks !

emmenko:
Bumping `patch` should be enough

PR #1387, opened 2022-10-17
ci: set correct deploy URL for prod and preview

emmenko:


nkuehn:
Sorry, I'm not understanding what the difference is in the behavior and what issue this addresses

emmenko:
On `main` the cypress tests fail because the base url is not defined.

image

nkuehn (approved):
I see the difference now in the logic, thanks

PR #1388, opened 2022-10-18
chore: refresh uikit deps, keep react-intl v5

emmenko:
This might help reducing the bundle size.

What I did is running the `yarn up -R` command which forces a refresh of all matching dependencies to their latest allowed version. This is useful when some dependencies are defined with version ranges but there are different versions installed (you can inspect it with `yarn why`).

```
yarn up -R '@commercetools-uikit/*'
yarn up -R '@emotion/*'
yarn up -R '@babel/*'
```


gabriele-ct (approved):
let's see if it helps

PR #1393, opened 2022-10-18
chore(plugins): upgrade to the latest version of learning-plugin

gabriele-ct:


nkuehn (approved):
I do wonder what we need it for in the documentation website, but that's a separate discussion. 

Thank you!

gabriele-ct:
> I do wonder what we need it for in the documentation website, but that's a separate discussion.
> 
> Thank you!

We do, it's in progress together with the performance investigation

PR #1395, opened 2022-10-19
Ga site title

gabriele-ct:
I've enabled some demo of the functionality visibe here:
https://docskit-pr-1395.commercetools.vercel.app/docs-smoke-test/
and here:
https://docskit-pr-1395.commercetools.vercel.app/api-docs-smoke-test/

(update:exchanged build specific preview links with the permalink for the PR)


gabriele-ct:
This is the visual result in case of site matched:
- desktop /tablet width:
Screenshot 2022-10-19 at 17 00 45

- mobile width:
Screenshot 2022-10-19 at 17 01 15

- small mobile:
Screenshot 2022-10-19 at 17 01 24

packages/gatsby-theme-docs/src/data/top-menu.yaml
```diff
@@ -1,7 +1,7 @@
 - menu-title: Developer Center
   items:
     - label: HTTP API
-      href: /../api
```
gabriele-ct, line 4: just for demo, it has to be rolled back

packages/gatsby-theme-docs/src/data/top-menu.yaml
```diff
@@ -24,6 +24,6 @@
 - menu-title: Merchant Center
   items:
     - label: Documentation
-      href: /../merchant-center
```
gabriele-ct, line 27: just for demo, it has to be rolled back

nkuehn: You can adjust the megamenu content in the kit repository to, for example, have an actual structure like
 * Documentation
   *  Docs Kit Docs
 * Test Sites
    * API Docs Smoke Test
    * Docs Smoke Test
    * Site Template
  
It would fix an old long-standing issue with the test and docs site - it's completely confusing that they have the actual commercetools docs megamenu. 

The Top Bar would say  "Documentation -> Docs Kit Docs"  or "Test Sites -> API Docs Smoke Test" then.

nkuehn: (I know, it opens a rabbithole on how the megamenu content is actually to be distributed into the actual content repositories.  But as you can see, the one here is out of date anyways.

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -180,8 +181,71 @@ const SearchContainer = styled.div`
...
+const SiteContextTitle = styled.div`
+  margin-right: ${designSystem.dimensions.spacings.s};
+`;
+
+const SiteTitle = styled.div`
+  margin-right: ${designSystem.dimensions.spacings.s};
+  ${({ hasContext }) =>
+    hasContext &&
+    `
+    font-weight: 600;
```
nkuehn, line 193: Please use a design token for font weight and not a hardwired value.  Please use the one for semibold 

https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/ui-kit/src/design-system.ts#L310

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -180,8 +181,71 @@ const SearchContainer = styled.div`
...
+  `}
+`;
+
+const TitleContainer = styled.div`
+  ${({ hasContext }) =>
+    hasContext &&
+    `
+    @media screen and (${designSystem.dimensions.viewports.mobile}) {
+      font-size: 14px;
```
nkuehn, line 202: please use design tokens

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -1,10 +1,10 @@
 import React, { forwardRef } from 'react';
+import { useStaticQuery, graphql } from 'gatsby';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 import { AngleDownIcon, AngleUpIcon } from '@commercetools-uikit/icons';
 import IconButton from '@commercetools-uikit/icon-button';
-import SpacingsInline from '@commercetools-uikit/spacings-inline';
```
nkuehn, line 7: Why did you need to switch to manual spacing?   Maybe you don't need to do that any more when always having a default value for context?

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -180,8 +181,71 @@ const SearchContainer = styled.div`
...
+  margin-right: ${designSystem.dimensions.spacings.s};
+  ${({ hasContext }) =>
+    hasContext &&
+    `
+    font-weight: 600;
+  `}
+`;
+
+const TitleContainer = styled.div`
+  ${({ hasContext }) =>
```
nkuehn, line 198: see below, there should rather be a fallback value for UI consistency.  You can simplify the code then.

nkuehn (commented):
Thanks, good first iteration!  See my various detail comments, some just input about design patterns, some actual change suggestions. 

Visually, the spacing around the ">" sign is not consistent yet and the way the line breaks on smaller screens is not robust yet:

image

To closely control what breaks together with what you can apply a trick that is wrapping `<span style="display: inline-block;">`  around the elements that should not be separated and have the surrounding element allow wrapping (either flexbox based or plain inline content wrapping).  We've implemented that CSS pattern in the API docs to have the HTTP API path wrap in a meaningful way: https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/gatsby-theme-api-docs/src/components/resource/method/url.js#L26-L39

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
gabriele-ct: I changed to a flexbox approach so that the title automatically rearrange on 2 lines in case the viewport is too narrow. In order to do that I had to get rid of the `SpacingsInline` which, if I don't understand wrong, is arranging all the elements in a single row and doesn't allow the content to "flow" on multiple rows.

> Maybe you don't need to do that any more when always having a default value for context?
Not sure I understand what you mean with default value for context

packages/gatsby-theme-docs/src/data/top-menu.yaml
gabriele-ct: don't feel like jumping into this rabbithole, at least in the scope of this ticket. I've changed the `menu-title` so that we don't beed to make any string comparison to display the Title in the top bar. If we want to address other issues related to the menu configuration I'd rather create a separate ticket properly groomed with well defined requirements.

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
nkuehn: A default value for the actual SiteContext if a given microsite is not appearing explicitly in the top menu

gabriele-ct:
I've tried `<span style="display: inline-block;">` but didn't work out well (the caret was always on the second line). So I used  `inline-flow` and seems to work as expected

nkuehn:
cool, I learnt something new - the two value display CSS property options.   

`flow` is experimental but for such an edge case that's ok - gracefully degrades still functional

nkuehn (approved):
It's beautiful - thank you for bearing with me through the details

packages/gatsby-theme-docs/src/data/top-menu.yaml
emmenko: @nkuehn Do we already have a ticket to address the "distribution" of the production top menu data?

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -224,14 +273,25 @@ const LayoutHeader = forwardRef((props, ref) => {
...
-            <SpacingsInline alignItems="center">
-              <span>{props.siteTitle}</span>
-              {props.isTopMenuOpen ? (
-                <AngleUpIcon size="medium" color="info" />
-              ) : (
-                <AngleDownIcon size="medium" />
+            <TitleContainer hasContext={!!siteContextTitle}>
+              {siteContextTitle && (
+                <SiteContextTitle>
+                  {siteContextTitle}&nbsp;&nbsp;&gt;{' '}
```
emmenko, line 279: Nit: have you tried other approaches than `&nbsp;&nbsp;&gt;{' '}`? For example using `::after`?

image

packages/gatsby-theme-docs/src/hooks/use-top-menu-items.js
```diff
@@ -0,0 +1,32 @@
...
+
+  const siteContextMap = new Map();
+  data.allTopMenuYaml.nodes.forEach((node) => {
+    node.items.forEach((item) => {
+      if (item.href.startsWith('/../')) {
+        const minisiteSegment = item.href.split('/')[2];
+        if (minisiteSegment) {
+          siteContextMap.set(minisiteSegment, node.menuTitle);
+        }
+      }
```
emmenko, line 25: Can you explain what this logic is for? 🤔

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -224,14 +273,25 @@ const LayoutHeader = forwardRef((props, ref) => {
             onClick={props.toggleTopMenu}
             onKeyPress={handleTopMenuButtonKeyPress}
           >
-            <SpacingsInline alignItems="center">
-              <span>{props.siteTitle}</span>
-              {props.isTopMenuOpen ? (
-                <AngleUpIcon size="medium" color="info" />
-              ) : (
-                <AngleDownIcon size="medium" />
+            <TitleContainer hasContext={!!siteContextTitle}>
```
emmenko, line 276: Can you give an example when the site context title is not defined? Is it only for backwards compatibility? Either way we can maybe leave a comment explaining the use case.

emmenko (commented):
Tested a bit in the preview URL 👌

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
gabriele-ct: I haven't but I can give it a go. Thank you 👍

packages/gatsby-theme-docs/src/hooks/use-top-menu-items.js
gabriele-ct: sure
```
      if (item.href.startsWith('/../')) { // this excludes absolute urls 
        const minisiteSegment = item.href.split('/')[2]; // this gets the first non ".." segment of the url, normally matching the minisite baseUrl segment
        if (minisiteSegment) {
          // creates a map mapping a specific minisite to his ancestor as defined in the `top-menu.yaml`
          siteContextMap.set(minisiteSegment, node.menuTitle);
        }
      }
```

The idea behind this is that depending on the minisite I'm on, I've got a reverse map `siteContextMap` where I can lookup if the minisite belongs to Composable Commerce or Composable Frontend (or in main general look up what's the ancestor of a given minisite).

packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
gabriele-ct: I implemented this as it was a requirement of the ticket:
https://github.com/commercetools/commercetools-docs/issues/2374

With the current setup I believe we'll always have the context available.  I'll add a comment to explain that

PR #1396, opened 2022-10-19
chore: update Percy to v1

emmenko:


emmenko:
It seems in the v1 version the Percy static snapshots are all NEW.

Someone should go over them and make sure they are all as expected. cc @commercetools/docs-tech

nkuehn:
closing, implemented in #1531

PR #1399, opened 2022-10-19
ci: try perf improvements to speed up build times

emmenko:
Trying some stuff mentioned here: https://www.gatsbyjs.com/docs/how-to/performance/improving-build-performance/


.github/workflows/main.yml
```diff
@@ -63,6 +63,7 @@ jobs:
         env:
           ENABLE_SEARCH: 'true'
           NODE_OPTIONS: '--max-old-space-size=6144'
+          GATSBY_EXPERIMENTAL_QUERY_CONCURRENCY: 32,
```
nkuehn, line 66: For your new M2 max pro enterprise CPU :-)

websites/api-docs-smoke-test/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@commercetools-website/api-docs-smoke-test",
   "version": "6.0.21",
   "private": true,
+  "browserslist": [">0.25%", "not dead and supports es6-module"],
```
nkuehn, line 5: I wondered, too whether we can do that already.    I think it's OK because the site degrades gradually if the JS fails (not completely but at least you can move around the content if you found it through google). 

We'd have to apply it manually to the repositories using the kit, right?

emmenko: > I wondered, too whether we can do that already.

I think we should be good. I checked the supported browsers and these are all old versions https://browserslist.dev/?q=PjAuMjUlLCBub3QgZGVhZCBhbmQgc3VwcG9ydHMgZXM2LW1vZHVsZQ%3D%3D

> We'd have to apply it manually to the repositories using the kit, right?

I haven't looked much into it but yes I suppose so.

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -8,6 +8,8 @@ const fs = require('fs');
 const path = require('path');
 const { createFilePath } = require('gatsby-source-filesystem');
 const { ContextReplacementPlugin } = require('webpack');
+const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
```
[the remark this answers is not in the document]
emmenko: We're planning to remove `moment` (which is actually a [maintenance-only library](https://momentjs.com/docs/#/-project-status/)) from our tooling at some point.
See also https://wiki.commercetools.com/display/shield/%5BRFC%5D+New+standard+package+to+deal+with+dates+and+numbers

So I see this more of a temporary thing (we have something similar in the MC as well).

.github/workflows/main.yml
emmenko: 🤪

I wanted to try to max value first and see what happens 🙌

emmenko: In any case, with no cache now it took 7-8 min. Seems already a small improvement.

websites/api-docs-smoke-test/package.json
nkuehn: I checked the readme, it can live in the package.json of a parent directory, too or in a separate config file in a parent directory:  https://github.com/browserslist/browserslist#queries   

That would be less maintenance than having it in every website (esp in the commercetools-docs repo which is getty pretty big now)

nkuehn (dismissed):

nkuehn:
I was a bit too optimistic approving this  :-(   The build fails with a low level nodejs error

````
FATAL ERROR: Error::ThrowAsJavaScriptException napi_throw
 1: 0xb09980 node::Abort() [/opt/hostedtoolcache/node/16.14.0/x64/bin/node]
 2: 0xa1c235 node::FatalError(char const*, char const*) [/opt/hostedtoolcache/node/16.14.0/x64/bin/node]
 3: 0xa1c23e  [/opt/hostedtoolcache/node/16.14.0/x64/bin/node]
 4: 0xad139b napi_fatal_error [/opt/hostedtoolcache/node/16.14.0/x64/bin/node]
 5: 0x7f37f018afea  [/home/runner/work/commercetools-docs-kit/commercetools-docs-kit/node_modules/@parcel/cache/node_modules/@lmdb/lmdb-linux-x64/node.napi.glibc.node]
 6: 0x7f37f018b9e9  [/home/runner/work/commercetools-docs-kit/commercetools-docs-kit/node_modules/@parcel/cache/node_modules/@lmdb/lmdb-linux-x64/node.napi.glibc.node]
 7: 0x7f37f01c352a  [/home/runner/work/commercetools-docs-kit/commercetools-docs-kit/node_modules/@parcel/cache/node_modules/@lmdb/lmdb-linux-x64/node.napi.glibc.node]
 8: 0xab44dd  [/opt/hostedtoolcache/node/16.14.0/x64/bin/node]
 9: 0xd53cfe  [/opt/hostedtoolcache/node/16.14.0/x64/bin/node]
10: 0xd5511f v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [/opt/hostedtoolcache/node/16.14.0/x64/bin/node]
11: 0x15f0b79  [/opt/hostedtoolcache/node/16.14.0/x64/bin/node]
```

It seems to be inside the native lmdb node extension.  My first guess is that the concurrency setting is too aggressive for the build environment vs. our local machines?

emmenko:
I can try taking another look in the next days, some new things got merged which might have affected this.

nkuehn (approved):
I adjusted the settings and resolved the conflicts. Good to go and helps with our current bundling performance issues.

PR #1400, opened 2022-10-20
Ga video component

gabriele-ct:
closes commercetools/learning-tech#235


packages/gatsby-theme-docs/package.json
```diff
@@ -39,6 +39,7 @@
     "@mdx-js/react": "1.6.22",
     "@sentry/browser": "7.15.0",
     "@svgr/webpack": "6.4.0",
+    "@videojs/themes": "^1.0.1",
```
gabriele-ct, line 42: if we ever going to have a custom theme we can get rid of this dependency

packages/gatsby-theme-docs/src/components/video-player-client-side.js
```diff
@@ -0,0 +1,57 @@
...
+    };
+  }, [playerRef]);
+
+  return (
+    <div data-vjs-player>
+      <video ref={videoRef} className="video-js vjs-big-play-centered" />
+    </div>
+  );
+};
+VideoPlayer.propTypes = {
```
gabriele-ct, line 50: We need to think a bit better the interface of the Component.
Since everything gets passed to the videoJs in a single `options` object, I thought to implement a  list of `important` props that needs to be explicit for the editors to be able to play with them, but also accept the full option object to cater for corner cases. I implemented a override mechanism in a way that the explicit props are always overriding the option object

websites/docs-smoke-test/src/content/components/video-player.mdx
```diff
@@ -0,0 +1,15 @@
...
+---
+
+# Video Player
+
+<VideoPlayer options={{
+    controls: true,
+    responsive: true,
+    fluid: true,
+}}
+  videoUrl="https://bitdash-a.akamaihd.net/content/sintel/hls/playlist.m3u8"
```
gabriele-ct, line 12: Just a sample video and sample thumbnail

packages/gatsby-theme-docs/src/components/video-player-client-side.js
nkuehn: I am not a fan of officially allowing direct interaction with the underlying video player library at all.   Conceptually everyhing in the docs kit is conventions based and KISS for authors.    Plus, staying able to swap the underlying library is a plus, too. 

Getting the default behavior right  (i.e. very good) through feedback is worth more to me. 

Looking at https://videojs.com/guides/options/  I'n not seeing something that I want to be different per-content.

websites/docs-smoke-test/src/content/components/video-player.mdx
```diff
@@ -0,0 +1,15 @@
+---
+title: VideoPlayer component
+---
+
+# Video Player
+
+<VideoPlayer options={{
```
nkuehn, line 7: My "wishlist" Author interface would be: 

```jsx
<Video 
   url="https://bitdash-a.akamaihd.net/content/sintel/hls/playlist.m3u8" 
   posterPath="foobar-module/intro-video-poster.jpg"  // a path to an image in the repo following the same conventions as when we refer to other images 
/> 
```
- No autoplay (does not work reliable anyways and browser vendors do not want it) https://videojs.com/blog/autoplay-best-practices-with-video-js/
-  just "video" 
- No low level options passthrough (see separate comment)
- "Poster" describes it better,  a "thumbnail" is a small size version of a bigger image. That's also the name in the HTML standard https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

emmenko:
Can I ask what are the requirements here and what is the reason for going with a video library?

Are we hosting video files as well or only embedding them?

packages/gatsby-theme-docs/src/components/video-player.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import LoadingSpinner from '@commercetools-uikit/loading-spinner';
```
nkuehn, line 3: Concerning the loading UX, we don't have to get it 100% perfect but the current preview has too much layout shift and general "weirdness".   Can you at least take a look at the quiz component, which is already doing something similar 
- reserve the space before the client component is loaded based on the assumption that it's a 16:10 video in full responsive width - maybe frame it with a grey hairline 
- center the loading animation. 
- consider using a 16:10 aspect ratio example video.  We will only very unlikely use a wide format like that example you used, rather more squarish than the other direction.

FFawzy:
@emmenko 
as part of the self learning initiative, some of the courses will include videos so we will need to produce, host and deliver them.
see this ticket for investigation 
https://github.com/commercetools/learning-tech/issues/221

packages/gatsby-theme-docs/src/components/video-player-client-side.js
```diff
@@ -0,0 +1,57 @@
+import React, { useRef, useEffect } from 'react';
+import videojs from 'video.js';
+import PropTypes from 'prop-types';
+import 'video.js/dist/video-js.css';
+
+const prepareVideoOptions = (options, videoUrl, autoplay, poster) => {
```
nkuehn, line 6: 1. Can you hide the default caption and (whatever the headphone symbol is) controls?  They seem to appear independent of whether there are captions, or at least I am not seeing any captions when I activate one.   If, then we'll use a single button for english captions on/off (as discussed today, no perspective for non-english content).   Done through passing options to the component tree: https://videojs.com/guides/components/#default-component-tree  (scroll down from there to see the options and their value.  Sometimes it's just passing `false` for a component to remove it)

2. I suggest to hardwire language to english ( https://videojs.com/guides/options/#language ) - it would be confusing if all our UI controls text is english but the player uses browser localized strings.

websites/docs-smoke-test/src/content/components/video-player.mdx
```diff
@@ -0,0 +1,15 @@
+---
+title: VideoPlayer component
```
nkuehn, line 2: I suggest to make that a full width layout page to be able to test (and maybe fix) how it behaves if the page is full width for two-column content on a big screen.

packages/gatsby-theme-docs/src/components/video-player-client-side.js
nkuehn: PS: I ran across a nice hack to remove the PiP on devices that don't support it (typically mobiles):  

`{ controlBar: { pictureInPictureToggle: !!document.pictureInPictureEnabled } }`

websites/docs-smoke-test/src/content/components/video-player.mdx
nkuehn: stupid question: Is the GitHub octocat the default poster image inside videojs or did you set that or is it in the playlist you linked?     In any case - it's not exactly ideal as a default (other peoples' brand).

nkuehn:
> Can I ask what are the requirements here and what is the reason for going with a video library?
> Are we hosting video files as well or only embedding them?

The de facto standard that the stream hosting vendors offer is HLS. We're not going to self-host videos (all the transcoding into resolutions and formats and the actual HLS support for skipping, preload etc).  Native HLS is only Safari, they are still quabbling around at apple and google.   

https://caniuse.com/?search=hls

We can still switch to "raw" HTML5 behind the scenes in this component but for immediately now we need a low-effort approach that has all the controls, caters for all the browser quirks, controls work responsive etc.

nkuehn:
PS: If you feel ambitious and want to explore listening to the events here's the documentation on how to track custom events into google analytics:  https://www.gatsbyjs.com/plugins/gatsby-plugin-google-analytics/#trackcustomevent-function  

Label could be the page title.

websites/docs-smoke-test/src/content/components/video-player.mdx
gabriele-ct: No, it was just a test since locally the static images are not working, so I wanted and absolute url. I'll replace it with a commercetools logo

packages/gatsby-theme-docs/src/components/video-player-client-side.js
gabriele-ct: Regarding the first point:
The headphone symbol (audio tracks) and the CC symbol (subtitles) are automatically sown/hidden depending on the presence of multiple audio tracks and CC files. In the example I added, both are present to you get the 2 buttons enabled
I added an additional video (unfortunately I've only found a live stream) that doesn't have additional audio tracks or CC and the 2 buttons are hidden.

The other points are done

packages/gatsby-theme-docs/src/components/video-player.js
gabriele-ct: @nkuehn I tried to do that but I'm not 100% sure to understand where I should place the container. I still see quite a lot of layout shift. Any ideas?

gabriele-ct:
> PS: If you feel ambitious and want to explore listening to the events here's the documentation on how to track custom events into google analytics: https://www.gatsbyjs.com/plugins/gatsby-plugin-google-analytics/#trackcustomevent-function
> 
> Label could be the page title.

I'm unsure if this was mentioned in the initial discussion, I'm not sure what's the expected outcome. probably better tackling this on a separate ticket

packages/gatsby-theme-docs/src/components/video-player.js
```diff
@@ -0,0 +1,56 @@
...
+import { customProperties } from '@commercetools-uikit/design-system';
+import styled from '@emotion/styled';
+
+const VideoPlaceholder = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 100%;
```
nkuehn, line 11:
```css
aspect-ratio: 16 / 9; 
width: 100%;
```
should do  the job, no need to do script based calculations

packages/gatsby-theme-docs/src/components/video-player.js
```diff
@@ -0,0 +1,56 @@
...
+  useLayoutEffect(() => {
+    if (!ref || !ref.current) {
+      return;
+    }
+    const width = ref.current.offsetWidth;
+    if (width) {
+      const height = width * (10 / 16); // 16:10 ratio
+      setPlaceholderHeight(height);
+    }
+  }, []);
```
nkuehn, line 32: see above, CSS is enough and produces less potential for layout shift and flicker

packages/gatsby-theme-docs/src/components/video-player.js
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    <>
+      {isClientSide && (
+        <React.Suspense
+          fallback={
+            <VideoPlaceholder height={placeholderHeight} ref={ref}>
```
nkuehn, line 39: I think the placeholder belongs around suspense and not inside it to have no layout shift.

websites/docs-smoke-test/src/content/components/video-player.mdx
```diff
@@ -0,0 +1,20 @@
...
+title: Video component
+wideLayout: true
+---
+
+# Video Player
+
+## Video 1
+<Video
+  url="https://customer-ytbpo1yna9xohg5m.cloudflarestream.com/e1aa2e24614b242afdd4afb36a0c2f75/manifest/video.m3u8"
+  posterPath="https://customer-ytbpo1yna9xohg5m.cloudflarestream.com/e1aa2e24614b242afdd4afb36a0c2f75/thumbnails/thumbnail.jpg?time=6s"
```
nkuehn, line 11: if it takes URLs or paths dynamically the propname would better be just "poster" (back to the original).  a URL is definitely not a "path"

nkuehn (commented):
sorry for the delayed feedback

nkuehn:
PS: sorry, I forgot the following in the review:   I'm not convinced that going full width on a full width page is a good idea.  The video typically does not fit the screen in height any more. 

There's CSS somewhere in the page content wrapper that decides which kinds of content elements are constrained to the main column width. 

image

Adding a test case where the video sits inside the `<SideBySide>` wrapper could also help - for example putting the video as the second element in sidebyside, which moves it to the right column.  

Also, a test case where no poster is defined would help robustness

packages/gatsby-theme-docs/src/components/video-player.js
gabriele-ct: I think I now got it right

packages/gatsby-theme-docs/src/components/video-player.js
```diff
@@ -0,0 +1,46 @@
...
+import { customProperties } from '@commercetools-uikit/design-system';
+import styled from '@emotion/styled';
+
+const VideoPlaceholder = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 100%;
+  aspect-ratio: 16 / 9;
+  height: ${(props) => props.height}px;
```
nkuehn, line 13: is the height line a leftover from the previous implementation?

packages/gatsby-theme-docs/src/components/video-player.js
```diff
@@ -0,0 +1,46 @@
...
+import LoadingSpinner from '@commercetools-uikit/loading-spinner';
+import { customProperties } from '@commercetools-uikit/design-system';
+import styled from '@emotion/styled';
+
+const VideoPlaceholder = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 100%;
+  aspect-ratio: 16 / 9;
```
nkuehn, line 12: It's pretty good not but still has a little bit of shift in the first example video because 16:9 does not include the control bar of the video player (30px).   
you can add a bottom padding of 30px, that will make it fit. 
```suggestion
  aspect-ratio: 16 / 9;
  padding-bottom: 30px;
```

Alternatively a youtube-ish layout with the control bar fading in only on hover could be done but I think we should not overengineer the interaction pattern.

nkuehn (approved):
Iteration is great :-) - it's starting to be smooth and I can imagine authors may like the side by side setup in some situations. 

I propsed a small improvement in the CSS.  

Only missing aspect to go live is to add it to the documentation website.  You could decide to extend and rename the existing "Images and diagrams" page ( https://docskit-pr-1400.commercetools.vercel.app/documentation/writing/images ) towards Embedding Media in general.    Or some other place you find suitable (but I'd like to avoid adding to many small pages, that makes it rather hard to navigate)

Approved assuming what's left here is just hygiene.   Overall very nice to see it work.

packages/gatsby-theme-docs/src/components/video-player.js
gabriele-ct: Checking the current CSS applied to the controls turns out the height is set to `3em`
I set the same to the bottom-padding of the placeholder.

Screenshot 2022-10-26 at 14 36 59

BTW, I still see layout shift. It seems like it happens within the videojs implementation. As soon as the client component loads, just before it shows the poster, I can see the black area becoming smaller, and just after that jumping back to the normal size.
I'll keep looking into it although the overall effect is now much better 👍

packages/gatsby-theme-docs/src/components/video-player.js
gabriele-ct: Indeed 🤦

packages/gatsby-theme-docs/src/components/video-player.js
nkuehn: Thanks.  It's a bit frustrating the the videjs player has layout shift in itself while loading (it seems it's first reserving the space for the video and only then adding the control bars).  

So that remaining bit is out of our control, at least in the current scope to not overdo it before we have actual content.

FFawzy:
is it possible/better to have the PIP when we scroll away automatically, or optional for a user?

also I noticed when we have PIP and navigate to a different page the video stays with the user (because its technically a separate window) BUT the video stops playing and freezes, is that intentional or a bug?

gabriele-ct:
> is it possible/better to have the PIP when we scroll away automatically, or better keeping it optional for a user?

@nkuehn already replied. Just wanted to add that is should be possible using the native
https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/requestPictureInPicture API

We could consider it an option behind a flag. 

> also I noticed when we have PIP and navigate to a different page the video stays with the user (because its technically a separate window) BUT the video stops playing and freezes, is that intentional or a bug?

Good candidate for a separate ticket to implement what @nkuehn suggested. going to create it in the backlog

nkuehn:
> is should be possible using the native
https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/requestPictureInPicture API

It's called "request" for a reason.  It triggers a user confirmation normally.    Compare to "exitPictureInPicture" not being called "requestExitPictureInPicture".   

Thanks for opening the new issue!

Issue #1341, opened 2022-07-13
Issue with style check for JSX elements with a `.` in the name


emmenko:
Seems to be related to these recent changes: #1321 

<img width="490" alt="image" src="https://user-images.githubusercontent.com/1110551/178750627-5bbd217b-62dd-4aa5-8721-36b8b71e15ed.png">

I think the regex needs to be adjusted.

timonrey:
Yes, thank you. I will add it to the board 👍🏼

emmenko:
Found another regression:

<img width="578" alt="image" src="https://user-images.githubusercontent.com/1110551/193021573-5388e8d7-8f40-4f8c-bacc-02bb62d19615.png">

nkuehn:
Sorry, got lost between bigger projects.  Unfortunatly we have to find the MDX expressions ourselves   The place to go is 
here: https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/writing-style/.vale.ini#L15  

The current regex is only covering self-closing tags because only these were a practical issue until now.   We'll have to make it more paranoid and add opening and closing tags, too. 

> Found another regression:

Github Style links (without explicit markdown link syntax) are not supported,  ideally generally not used e.g. for refactoring, finding links etc.   If you want just the text, ideally make it a string literal  `https://foo/bar`

emmenko:
> Github Style links (without explicit markdown link syntax) are not supported, ideally generally not used e.g. for refactoring, finding links etc. If you want just the text, ideally make it a string literal `https://foo/bar`

Not sure I get the point. This is not a normal link, it's using a code block to show the URL format.

<img width="475" alt="image" src="https://user-images.githubusercontent.com/1110551/193828211-ba0ce8f6-a5b8-4aa3-b619-c0391127af92.png">

https://github.com/commercetools/merchant-center-application-kit/blob/20c0be80ea2bb959c9f298cb7f6784841acef948/website/src/content/concepts/merchant-center-api.mdx?plain=1#L134-L136

How do you suggest to proceed with this?

gabriele-ct:
@emmenko and @nkuehn 
are you sure the first error (regarding `<Foo.Bar>` tag) is still relevant?
I've tried pulling latest main and I don't get that error:


![Image](https://user-images.githubusercontent.com/106664971/193827894-64c76d5d-5ed6-409f-a69a-7ed185b40dcd.png)

[...]

PR #1405, opened 2022-10-22
fix(writing-style): use `TokenIgnores` instead of `BlockIgnores`

emmenko:
Attempt to fix the issue with Vale + MDX (JSX).

Previous attempt was done in #1321 but this introduced a regression (see #1341).

It seems thought that using `TokenIgnores` instead of `BlockIgnores` fixes both issues.

See examples below in docs-kit:

**Without ignore regex**

Screenshot 2022-10-22 at 15 12 27

**With `TokenIgnores` regex**

Screenshot 2022-10-22 at 15 13 04

I don't know if this was the actual original issue that led to #1321, maybe @nkuehn can confirm and double check.
Thanks!


timonrey (approved):

emmenko:
@nkuehn Is this change ok for you as well?

nkuehn:
I haven't found the time to analyze it for real but intuitively it feels more correct to me because JSX can be interleaved with markdown even inside what markdown considers a "Block"

emmenko:
Should we give this a try or do you have other suggestions?

gabriele-ct:
I'm fine with it. I'm going to merge this and we see how if goes. FYI @nkuehn

PR #1414, opened 2022-11-09
Move banner-layout component to overrides

### LINKED ISSUE #1413: Make the whole banner area overridable
We decided that we don't want to show the user research banner in the frontend microsites. However, if we override the banner components with an empty component, the layout still shows a border on the left.
So we need to be able to override the whole banner area in the top-right corner if we don't want to show any banners or other content.

timonrey:
closes #1413


gabriele-ct (approved):
LGTM

Issue #1410, opened 2022-10-28
Dev mode: useLayoutEffect warning is back


nkuehn:
Dev mode now throws the following warning:
```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. 
This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. 
To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. 
See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```
The only place where we directly use it is still the sidebar, but there is a specific comment and handling for the case.  Maybe we need to revisit the possible workarounds / fixes from the react docs and pick a different one, .

gabriele-ct:
Some findings:
- the warning definitely doesn't come from the sidebar. I even tried to remove entirely the code and I still get the warning when the homepage loads. As additional confirmation, disabling the _isomorphic hack_ in sidebar, I get a lot more of warnings.
- The warning is certainly generated by some code in `@commercetools-docs/gatsby-theme-docs` package as even the most minimal implementation (`websites/site-template`) generates that warning and the only dependency it has is indeed `@commercetools-docs/gatsby-theme-docs`
- My hypothesis is that it might be related to some dependency we use that is not ssr safe, but I can't figure out what is actually causing the error
- Looking at the error stack trace I can guess it's something related with the MDXProvider (see below the stack trace)

```
at ThemeProvider (/Users/gabrieleantonini/projects/commercetools-docs-kit/websites/site-template/.cache/page-ssr/routes/render-page.js:626:79)
    at LayoutApplication (/Users/gabrieleantonini/projects/commercetools-docs-kit/websites/site-template/.cache/page-ssr/routes/render-page.js:55655:32)
    at LayoutContentHomepage (/Users/gabrieleantonini/projects/commercetools-docs-kit/websites/site-template/.cache/page-ssr/routes/render-page.js:55446:78)
    at ErrorBoundary (/Users/gabrieleantonini/projects/commercetools-docs-kit/websites/site-template/.cache/page-ssr/routes/render-page.js:53951:5)
    at ThemeProvider (/Users/gabrieleantonini/projects/commercetools-docs-kit/websites/site-template/.cache/page-ssr/routes/render-page.js:54971:258)
    at IntlProvider (/Users/gabrieleantonini/projects/commercetools-docs-kit/websites/site-template/.cache/page-ssr/routes/render-page.js:127796:47)
[...]

PR #1418, opened 2022-11-10
refactor: load CSS variables via global import instead of using `<ThemeProvider>`

emmenko:
Temporary workaround for #1410

### DIFF
--- a/packages/gatsby-theme-docs/globals.css
+++ b/packages/gatsby-theme-docs/globals.css
@@ -1,0 +1,6 @@
+@import '@commercetools-uikit/design-system/materials/custom-properties.css';
+
+:root {
+  --font-family-default: 'Roboto', sans-serif;
+  --font-family-body: 'Roboto Mono', monospace;
+}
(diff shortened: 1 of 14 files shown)


packages/gatsby-theme-docs/globals.css
```diff
@@ -0,0 +1,6 @@
+@import '@commercetools-uikit/design-system/materials/custom-properties.css';
+
+:root {
+  --font-family-default: 'Roboto', sans-serif;
+  --font-family-body: 'Roboto Mono', monospace;
+}
```
emmenko, line 6: Overriding some design tokens here.

packages/ui-kit/src/components/globals.tsx
```diff
@@ -2,10 +2,17 @@ import React from 'react';
...
+type TGlobalsProps = {
+  websitePrimaryColor: string;
+};
+
+const Globals = (props: TGlobalsProps) => (
   <Global
     styles={css`
+      :root {
+        --website-primary-color: ${props.websitePrimaryColor};
+      }
```
emmenko, line 14: This is defined here as it depends on the prop.

kark (approved):
👍

emmenko:
@commercetools/docs-tech For awareness

PR #1420, opened 2022-11-11
fix(content): active section calculations

gabriele-ct:


packages/gatsby-theme-docs/src/hooks/use-active-section.js
```diff
@@ -15,16 +15,32 @@ const getSectionElements = () =>
 
 const offset = calculateOffset();
 
+const isInViewport = (element) => {
+  const rect = element.getBoundingClientRect();
+
+  return (
```
gabriele-ct, line 21: This has to be fine tuned and properly tested, it's here just to give an idea

nkuehn: I think a comment verbalizing what the expression tests in layout terminology would help maintenance  - and me being able to review it ;-) .

Testing the console output seems to yield the correct output (at least without more intensive testing e.g. with scrollbars when no touch mouse is attached and the laptop is closed)

packages/gatsby-theme-docs/src/layouts/internals/page-content-inset.js
```diff
@@ -15,6 +15,9 @@ import { designSystem, Markdown } from '@commercetools-docs/ui-kit';
 const PageContentInset = styled.div`
   max-width: ${(props) =>
     props.maxWidth || designSystem.dimensions.widths.pageContentWide};
+  section {
```
nkuehn, line 18: if we're sure it does not lead to accidental scrollbars appearing in certain situations this looks like a better approach than trial-and-erroring offsets.  Consider commenting because we're making use of a side effect, the usage is not about the main purpose of the overflow property.

packages/gatsby-theme-docs/src/hooks/use-active-section.js
```diff
@@ -15,16 +15,32 @@ const getSectionElements = () =>
...
+};
+
 const useActiveSelection = () => {
   const [activeSection, setActiveSection] = React.useState();
   const onScroll = React.useCallback(() => {
     const sectionElements = getSectionElements();
     let nextActiveSection;
+    const visibleSections = [];
     sectionElements.forEach((section) => {
       if (section.getBoundingClientRect().top - offset < 0) {
```
nkuehn, line 36: is the offset still needed when expanding the section area?

packages/gatsby-theme-docs/src/hooks/use-active-section.js
```diff
@@ -15,16 +15,32 @@ const getSectionElements = () =>
...
+  );
+};
+
 const useActiveSelection = () => {
   const [activeSection, setActiveSection] = React.useState();
   const onScroll = React.useCallback(() => {
     const sectionElements = getSectionElements();
     let nextActiveSection;
+    const visibleSections = [];
     sectionElements.forEach((section) => {
```
nkuehn, line 35: Just to document my confusion further:  The forEach over the sections looks like the last section in the `sectionElements` result is the one that ends up being highlighted through `setActiveSection`, but our observation is that the topmost one is highlighted if multiple are in view.

nkuehn:
P.S: to my comment:  From now on you can make silly jokes about me talking a lot about an intersection observer based implementation which actually never was implemented because we took an existing scrollSpy hook that turned out to be performant enough for the purpose.  

Side hint to the quiz preloading discussion:  can use the same scrollspy hook if that is easier but we need to keep an eye on scoll performance on weak devices.

gabriele-ct:
@nkuehn I'd like your opinion on the latest changes.

**Note**: when user reaches the bottom of the page, no scroll events get triggered even when selecting the specific items in the navigation menu (that's why I also need the `hashchage` event listener)

The solution feels quite hacky and still has glitches. For example, if a user clicks the last item and then scrolls up, the last item remains selected (because of the url hash) until it disappears of view, and then when the user scrolls down, it get selected straight away when it's in view, even if the user didn't explicitly clicked on it (again, because of the url hash).

Before going down the road of rewriting the browser history (to get the correct hash in the url), I'd like to know your opinion and how far makes sense to go

packages/gatsby-theme-docs/src/hooks/use-active-section.js
```diff
@@ -1,32 +1,97 @@
...
+        visibleSections.push(section);
       }
     });
-    setActiveSection(nextActiveSection);
+    setActiveSection(
+      calclulateActiveSection(visibleSections, locationHash, sectionElements[0])
+    );
+    // by resetting the url hash it prevents some incorrect behaviours when the hash
+    // is changed but the page cannot scroll (bottom of the page)
+    if (window.location.hash) {
```
gabriele-ct, line 62: See the inline comment @nkuehn 
Not sure if you're going to like it.
This solves all the glitches with the section navigation menu but hides the hash from the url.
The hash functionality still works as before, it is just the URL that gets rewritten just after removing the hash. The UX result is that when pressing a link you don't see the hash in the url.
I don't think it's a biggie, we still have the copy url functionality in place, but wanted to check

nkuehn: Your gut feeling is right, that's not how links in a content website should work IMHO, especially we can't have them behave inconsistently depending on whether a link leads to a new microsite, a new page or the same page.    

You'll need to explain me what the "glitches" would be.

gabriele-ct: reverted.

gabriele-ct:
Ready for review. 🙇

FFawzy (approved):
LGTM 👍

PR #1423, opened 2022-11-15
Fix width of release notes date filter

### LINKED ISSUE #1382: Regression: Release notes filter inputs overflow invisibly
By chance I found a regression that makes the release notes filter unusable in certain viewports:

https://commercetools-docs-kit.vercel.app/docs-smoke-test/releases

image

I have not been able to seen when it came into place so we'll have to fix it without knowning the original cause.

timonrey:
closes #1382

### DIFF
--- a/.changeset/slow-laws-watch.md
+++ b/.changeset/slow-laws-watch.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Fix the width of the date filter on the release notes page.
--- a/packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
+++ b/packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
@@ -17,6 +17,13 @@ const DateLabel = styled.label`
   line-height: ${designSystem.typography.lineHeights.small};
 `;
 
+// Override min-width from the uikit date component
+const DateInputContainer = styled.div`
+  div span div {
+    min-width: 0;
+  }
+`;
+
 const ReleaseNotesFilterDates = () => {
   const [filterParams, setFilterParams] = useReleaseNotesFilterParams();
   const maximumDate = IsoDateFormat.format(new Date());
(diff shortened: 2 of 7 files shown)


gabriele-ct:
@timonrey I might be wrong but I think your fix is only addressing part of the problem. In fact, if you try to open the date selector component you get this bad behaviour
Screenshot 2022-11-16 at 10 01 33

I also believe that the component who dictate the min-width should be the calendar component. If you try to make it smaller it will start looking weird and "broken".
Maybe we should review the application layout allowing more space for the calendar component.
If we know that only specific pages use this component, maybe treat it as a special case to avoid changing the layout for all the application pages.
What do you think @nkuehn and @emmenko ?

timonrey:
@gabriele-ct Thank you for mentioning this. I also don't like to modify the uikit components. I think the best solution would be to allow the right-hand column to have more space on the release notes page. But I would also like to avoid changing the layout for the whole website.

emmenko:
Might be worth looking into the uikit component to see if we can add an option to decide the position of the calendar. For example, in this case it should render right-aligned.

gabriele-ct:
@timonrey I still see the date filter cut out when resolution width is between 1024px and ~1100px.
Maybe it's possible to tweak a little bit more the breakpoint in a way that the left navbar disappears earlier?

gabriele-ct (changes requested):
Please see my previous comment

nkuehn:
Quick log from a conversation among me an @timonrey :  UI Kit does not intend to allow smaller widths for the date picker any time soon, it would be a change in the UX concept (the MC Ui Kit date picker "drops down" in the same width instead of being a floating overlay).  

Timon will spend a bit of timeboxed research to style the native HTML date picker to a good enough look (not the part that pops up, just the input when not opened).    Safari supports it, too by now so it's become a valid option if we can live with slight style deviations.  Mobile support is great on the plus side.

gabriele-ct:
Closing as the actual solution was implemented here: https://github.com/commercetools/commercetools-docs-kit/pull/1446

PR #1434, opened 2022-11-22
fix(video): wait for video element to be visible before run visual re…

gabriele-ct:
…gression

### DIFF
--- a/.percy.js
+++ b/.percy.js
@@ -22,7 +22,7 @@ module.exports = {
       },
       {
         include: /(.*-smoke-test.*)\/components\/video-player\/index\.html/,
-        waitForSelector: '#body-content div.video-js',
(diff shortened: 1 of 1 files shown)


.percy.js
```diff
@@ -20,6 +20,10 @@ module.exports = {
         include: /(.*-smoke-test.*)\/components\/mermaid-diagram\/index\.html/,
         waitForSelector: '#body-content svg.statediagram',
       },
+      {
+        include: /(.*-smoke-test.*)\/components\/video-player\/index\.html/,
+        waitForSelector: '#body-content div#vjs_video_998',
```
gabriele-ct, line 25: waits for the last video in the page to be visible

timonrey (approved):

PR #1437, opened 2022-11-23
Ga awareness poc

gabriele-ct:


gabriele-ct:
This is not needed anymore as the final implementation is on its way

PR #1439, opened 2022-11-24
Display product name in google search results

### LINKED ISSUE #1422: Search result should display product name alongside the minisite name
As result of the frontend docs migration  the following issue should be addressed.  In the search results the microsite name should be the combination of product and microsite as it is in the page header now.   (e.g.  “Composable Commerce > HTTP API”.
That requires either finding an existing DOM element that is present without javascript and addressing that or creating a special meta tag for it that the crawler configuration can read out.
Similar / related:  We may want to consider adjusting the autogenerated page name (which appears in search results and the tab name), too accordingly.

timonrey:
closes #1422 

Any further suggestions on how we want to display the google search results are welcome :)

### DIFF
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -33,6 +33,7 @@ module.exports = (themeOptions = {}) => {
       betaLink: '.',
       excludeFromSearchIndex: false,
       beta: false,
+      productName: 'commercetools',
       globalNotification: {
         active: false,
         notificationType: 'info',
--- a/packages/gatsby-theme-docs/gatsby-config.js
+++ b/packages/gatsby-theme-docs/gatsby-config.js
@@ -33,7 +33,6 @@ module.exports = (themeOptions = {}) => {
       betaLink: '.',
       excludeFromSearchIndex: false,
       beta: false,
-      productName: 'commercetools',
       globalNotification: {
         active: false,
         notificationType: 'info',
--- a/packages/gatsby-theme-docs/src/components/seo.js
+++ b/packages/gatsby-theme-docs/src/components/seo.js
@@ -10,13 +10,7 @@ import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
 import { useSiteData } from '../hooks/use-site-data';
 import useTopMenuItems from '../hooks/use-top-menu-items';
-
-const getSiteContextTitleByPath = (sitePathsMap, sitePath) => {
-  const siteSegment = sitePath.replace('/', '');
-  if (sitePathsMap.has(siteSegment)) {
-    return sitePathsMap.get(siteSegment);
-  }
-};
+import getSiteContextTitleByPath from '../utils/get-site-context-title';
 
 const SEO = (props) => {
   const siteData = useSiteData();
--- a/packages/gatsby-theme-docs/src/components/seo.js
+++ b/packages/gatsby-theme-docs/src/components/seo.js
@@ -9,9 +9,22 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
 import { useSiteData } from '../hooks/use-site-data';
+import useTopMenuItems from '../hooks/use-top-menu-items';
+
+const getSiteContextTitleByPath = (sitePathsMap, sitePath) => {
+  const siteSegment = sitePath.replace('/', '');
+  if (sitePathsMap.has(siteSegment)) {
+    return sitePathsMap.get(siteSegment);
+  }
+};
 
 const SEO = (props) => {
   const siteData = useSiteData();
+  const siteContextMap = useTopMenuItems();
+  const siteContextTitle = getSiteContextTitleByPath(
+    siteContextMap,
+    siteData.pathPrefix
+  );
   const excludeFromSearchIndex =
     props.excludeFromSearchIndex ||
     siteData.siteMetadata.excludeFromSearchIndex;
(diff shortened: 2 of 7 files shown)


packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -33,6 +33,7 @@ module.exports = (themeOptions = {}) => {
       betaLink: '.',
       excludeFromSearchIndex: false,
       beta: false,
+      productName: 'commercetools',
```
timonrey, line 36: default value.

packages/gatsby-theme-docs/src/components/seo.js
```diff
@@ -62,7 +62,7 @@ const SEO = (props) => {
   ].filter(Boolean);
   return (
     <Helmet
-      titleTemplate={`%s | ${siteData.siteMetadata.title} | commercetools`}
+      titleTemplate={`%s | ${siteData.siteMetadata.title} | ${siteData.siteMetadata.productName}`}
```
timonrey, line 65: I personally like the order here:
First the page title > then the website title > then the parent product name.

nkuehn (commented):
TBH I'm not very happy with the approach to have two parallel configuration paths for the megamenu dropdown / page nav at  top vs. the page title metadata and algolia search metadata.  People would have to keep top-nav.yaml and that new configuration flag here manually in sync.  

It's pretty safe to assume that it will go out of sync pretty soon in operation. 

I think the place were the top menu switcher pulls the site context is here: https://github.com/commercetools/commercetools-docs-kit/blob/b20ca8b5594277977438f3d3715d352bd9465fa2/packages/gatsby-theme-docs/src/layouts/internals/layout-header.js#L235

gabriele-ct:
> 
> I think the place were the top menu switcher pulls the site context is here:
> 
> https://github.com/commercetools/commercetools-docs-kit/blob/b20ca8b5594277977438f3d3715d352bd9465fa2/packages/gatsby-theme-docs/src/layouts/internals/layout-header.js#L235

Yes, that's where we get the "context" information which would correspond to the product name

packages/gatsby-theme-docs/src/components/seo.js
```diff
@@ -62,7 +75,9 @@ const SEO = (props) => {
   ].filter(Boolean);
   return (
     <Helmet
-      titleTemplate={`%s | ${siteData.siteMetadata.title} | commercetools`}
+      titleTemplate={`%s | ${siteData.siteMetadata.title} | ${
+        siteContextTitle || 'commercetools'
```
timonrey, line 79: We need a fallback because the home-homepage doesn't have a site context. This only works on the microsites.

websites/documentation/src/content/configuration/theme.mdx
```diff
@@ -38,9 +38,10 @@ The Docs-Kit theme also receives its plugin configuration through this standard
 - `betaLink` (_optional_, string): Configures the link for _beta button_.
 - `excludeFromSearchIndex` (_optional_, boolean): indicates that the website should not be indexed by crawlers. This option sets the `robots="noindex"` meta attribute on all pages of the site. If true, the search box in the top navigation is not displayed. **Default: `true`** Note that the on-site Algolia Docsearch is not respecting the instruction, it is purely based on click-trough reachability from the home page.
 - `globalNotification` (_optional_, object): Activates a global banner at the top of every page within the website. It contains three configuration options:
-    -  `active` (Boolean): flag to activate the global notification banner.
-    -  `notificationType` (String, must be either `info` or `warning`) defines the type and color of the banner.
-    -  `content` (String) defines the text within the banner.
+
+  - `active` (Boolean): flag to activate the global notification banner.
+  - `notificationType` (String, must be either `info` or `warning`) defines the type and color of the banner.
+  - `content` (String) defines the text within the banner.
```
timonrey, line 44: Pls ignore, just prettier.

packages/gatsby-theme-docs/src/components/seo.js
```diff
@@ -9,9 +9,22 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
 import { useSiteData } from '../hooks/use-site-data';
+import useTopMenuItems from '../hooks/use-top-menu-items';
+
+const getSiteContextTitleByPath = (sitePathsMap, sitePath) => {
```
gabriele-ct, line 14: I think the logic is the same as implemented in `layout-header.ts`. If that's the case, instead of duplicating the function, you could import it from the other file, or if you prefer, you could move it to a util module.

gabriele-ct (approved):
Just a minor code duplication issue that I'd recommend to fix. The rest is good

nkuehn:
@timonrey  @gabriele-ct  this is not implementing what we discussed in the grooming (But unfortunately I did not take care that we document what was spoken, so its not in the ticket text).  

1. We can't just completely remove the company name from the google search listing titles.   In google, it's out of context and "Frontend" could be anything and everything. 
2. the official full name of our products is "commercetools Frontend" and "commercetools Composable Commerce".  The Branding guide is clear that the "commercetools" can only be omitted if its' very obviously in the visual context of our brand (for example, the header bar of the docs site).   But in google, it has to be  for example  "Cart | HTTP API | commercetools Composable Commerce",  not  "Cart | HTTP API | Composable Commerce".   

Sorry for the pull back, but that's what we have branding rules for.

timonrey:
@nkuehn No worrries, thank you for keeping the branding rules in mind. I will fix that in another PR.
Just wanted to say that at least "commercetools Composable Commerce" will probably be too long to display in the search result header, it will most likely be cut off at the end. But I guess we can live with that.

nkuehn:
> will probably be too long to display in the search result header

For clarification: That's for the google results, means the title meta tag.   It's cut off in deep pages but for example on the microsite landing pages it will be visible because they have just one level (to be confirmed by the way). 

Apropos since you're here: What we forgot to address, too in this PR is the search results in Algolia docsearach.  in these you don't want the "commercetools" name because it's happening on our own branded site

PR #1443, opened 2022-11-25
Search result improvements

### LINKED ISSUE #1444: Display the product name in the algolia search results
related to #1422 
We need to adjust the aloglia search results so that we can display the product name on level 1.
That requires adding an extra element which the crawler can read out to get the product name and website name.

timonrey:
closes #1444 

This pull request changes the site-context implementation in `layout-header` to have an element that Algolia can use to read the full site context as it appears in the header.

If we split the information into two elements, the crawler is not able to read out one element to get a clean string.

### DIFF
--- a/algolia-crawler.js
+++ b/algolia-crawler.js
@@ -29,9 +29,10 @@ new Crawler({
           // FYI selector documentation: https://github.com/cheeriojs/cheerio#selectors
           recordProps: {
             lvl0: {
-              selectors: $(
-                'meta[name=commercetools:title-for-onsite-search]'
-              ).attr('content'),
+              selectors:
+                $('meta[name=commercetools:title-for-onsite-search]').attr(
+                  'content'
+                ) || '#site-title',
               defaultValue: 'General Topics',
             },
             lvl1: 'article h1',
--- a/algolia-crawler.js
+++ b/algolia-crawler.js
@@ -29,7 +29,9 @@ new Crawler({
           // FYI selector documentation: https://github.com/cheeriojs/cheerio#selectors
           recordProps: {
             lvl0: {
-              selectors: '#site-title',
+              selectors: $(
+                'meta[name=commercetools:title-for-onsite-search]'
+              ).attr('content'),
               defaultValue: 'General Topics',
             },
             lvl1: 'article h1',
--- a/packages/gatsby-theme-docs/src/components/seo.js
+++ b/packages/gatsby-theme-docs/src/components/seo.js
@@ -61,6 +61,10 @@ const SEO = (props) => {
       name: 'twitter:description',
       content: metaDescription,
     },
+    {
+      name: 'commercetools:title-for-onsite-search',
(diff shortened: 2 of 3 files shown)


packages/gatsby-theme-docs/src/layouts/internals/layout-header.js
```diff
@@ -255,6 +233,9 @@ const LayoutHeader = forwardRef((props, ref) => {
     siteData.pathPrefix
   );
 
+  // The siteContext is also being used to display the site context in algolias search results.
+  const siteContext = `${siteContextTitle} > ${props.siteTitle}`;
```
timonrey, line 237: I'm not sure if we can reproduce the exact style of the text inside of the top bar as it was before. For example the bold site title. Suggestions would be welcome :)

gabriele-ct: Maybe a possible solution would be to leave untouched the current UI implementation and create a hidden div, with the  `site-context` id,  containing just the text you need and then select it using the algolia crawler selector

nkuehn: We spoke about the implementation approach in the grooming and other locations,  I recommended a custom HTML head meta attribute using a name that is not in common use yet ( https://wiki.whatwg.org/wiki/MetaExtensions ), e.g. "commercetools:title-for-onsite-search" or whatever you think makes sense.

algolia-crawler.js
```diff
@@ -29,7 +29,9 @@ new Crawler({
           // FYI selector documentation: https://github.com/cheeriojs/cheerio#selectors
           recordProps: {
             lvl0: {
-              selectors: '#site-title',
+              selectors: $(
+                'meta[name=commercetools:title-for-onsite-search]'
+              ).attr('content'),
```
nkuehn, line 34: you can try out very simply in the browser dev console. The search bar at the bottom of the DOM explorer is evaluating CSS selectors.   

attr("content") looks good following the cheerio docs.  

You might want to make it a fallback logic for deployment (e.g. app kit goes live separately and would have no site name at all for a while)

packages/gatsby-theme-docs/src/components/seo.js
```diff
@@ -61,6 +61,10 @@ const SEO = (props) => {
       name: 'twitter:description',
       content: metaDescription,
     },
+    {
+      name: 'commercetools:title-for-onsite-search',
+      content: `${siteContextTitle} > ${siteData.siteMetadata.title}`,
```
nkuehn, line 66: You need to cover the case when the current microsite does not have a site context title available.  In that case just stick to the title alone.

algolia-crawler.js
timonrey: I would have waited with the actual code change in Algolia until I applied the update to all repositories. But it wouldn't hurt anyway 👍🏼

nkuehn (approved):
nice.  Please test the crawler config interactively after going live and update the result here, it's a manual process anyways.

PR #1445, opened 2022-11-28
feat(links): auto link to endpoint URN

gabriele-ct:


linkcheck-local-skip-list
```diff
@@ -11,3 +11,4 @@ https:\/\/support.commercetools.com
 \/.+\/sdk
 \/.+\/tutorials
 \/.+\/getting-started
+\#ctp:.+:endpoint
```
gabriele-ct, line 14: I had to disable linkcheck for all the links containing endpoint anchor because linkcheck seems to fail when checking those links: https://github.com/commercetools/commercetools-docs-kit/actions/runs/3574162285/jobs/6009092000
I assume it fails because the hash part of the link (which is url escaped) doesn't match exactly what contained in the anchor id (which is not url escaped)

packages/gatsby-theme-api-docs/src/components/resource/method/method.js
```diff
@@ -64,7 +67,7 @@ const Method = ({
           <TitleWithAnchor id={id}>{title}</TitleWithAnchor>
         ) : (
           // eslint-disable-next-line jsx-a11y/anchor-has-content, jsx-a11y/anchor-is-valid
-          <a name={id}></a>
+          <a id={id} name={id}></a>
```
gabriele-ct, line 70: If I just use the name attribute, the anchor doesn't work

packages/gatsby-theme-api-docs/src/hooks/use-endpoint-locations.js
```diff
@@ -0,0 +1,88 @@
...
+      const url = slug && urn ? `${slug}#${urn}` : '';
+
+      console.log('adding', `${apiKey}__${resource}__${method}`, url);
+      endpointLocations[`${apiKey}__${resource}__${method}`] = { url };
+    });
+  }
+  locationsAreIndexed = true;
+};
+
+// const convertTypeLocationOverrides = (typeLocationData) => {
```
gabriele-ct, line 39: overrides seem to be designed to work for types.
Do we need to support overrides on endpoints as well?
In that's the case, we need to have a different format of .yaml file as endpoints expects different parameters. Maybe a `endpoint-locations.yaml` file?

packages/gatsby-theme-api-docs/src/utils/ctp-urn.js
```diff
@@ -46,12 +54,15 @@ export const parseTypeURN = (urn) => {
...
-  apiKey = '',
-  path = '',
-  method = '',
-}) => {
-  return `${prefix}:${encodeURNComponent(apiKey)}:endpoint:${encodeURNComponent(
+export const generateEndpointURN = (
+  { apiKey = '', path = '', method = '' },
+  isDom
+) => {
+  if (isDom) {
```
gabriele-ct, line 61: I had to add this param to avoid escaping the actual URN identifier in the DOM, but still escape it for the URL.

gabriele-ct:
@nkuehn I'd like your input here. I've added few comments to highlight some possible discussion points.

nkuehn (commented):
A first few impressions.  You got the overally setup right, just that stupid question of encoding and IDs may need discussion (I am unsure whether I am on the right path).

websites/api-docs-smoke-test/src/content/layouts/endpoint-large-desktop.mdx
```diff
@@ -19,6 +19,11 @@ Lorem ipsum dolor sit amet, everti feugiat gloriatur sed et, an omnes saepe inte
 
 Id quo virtute epicurei phaedrum. No volumus atomorum sit, sonet sensibus ne quo. Tale porro cu qui, ridens option praesent ius ei, nam brute iracundia cu. Aeque soluta necessitatibus te quo, cu vel aperiri euismod delectus. Eu quod invenire mnesarchum sed, his ne decore mollis audire, sit ea aeque zril doctus.
 
+## Test with URL Linking
+
+The [following link](ctp:test:endpoint:/{projectKey}/resource/{id}:POST) should point to the `resource` API endpoint.
```
gabriele-ct, line 24: @nkuehn I tried adding endpoint and type link in standard mdx content but it doesn't seem to be supported. 
I'm not sure about how hard it could be adding support for type and endpoint links in standard mdx content. Shall we extend the scope of this ticket covering the use case you pointed out or would you rather add a separate ticket for that?

nkuehn: good catch,  I think we simply never completed the feature when it was first implemented.  
The "magic" link handling that makes a lot of the content writing "just work" is here:  https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/gatsby-theme-docs/src/components/link.js

What's blocking is the dependency structure between api docs theme and the general theme. The link resolution logic is pretty monolithic, hard to override / extend by an optional plugin.  

I suggest you leave it out here and open a tech maintenance issue to consider merging the two plugins.  It's in line with the general strategy to reduce the complexity.  The original intent was to not bloat the bundle size of every docs page with all the functionality that is not really needed but since that was decided gatsby's abilities to manually control bundling has gotten much better and we should be able to have just one theme with all features and still make sure the big pieces are loaded separately.

gabriele-ct: Thank you. I've created a placeholder ticket https://github.com/commercetools/commercetools-docs-kit/issues/1450 linking to this thread

nkuehn (approved):
nice!

Please add a changeset before merging

PR #1446, opened 2022-11-29
Replace `date-filter` ui-kit component with new custom component as release note filter

### LINKED ISSUE #1382: Regression: Release notes filter inputs overflow invisibly

timonrey:
Follow up from the discussion here:
https://github.com/commercetools/commercetools-docs-kit/pull/1423

will close #1382 

This pull request replaces the date-input filter from the ui-kit with a new custom native html date picker.

Preview deeplink:  https://docskit-pr-1446.commercetools.vercel.app/docs-smoke-test/releases

### DIFF
--- a/packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
+++ b/packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
@@ -99,6 +99,8 @@ const ReleaseNotesFilterDates = () => {
     });
     setFromFilterDate(undefined);
     setToFilterDate(undefined);
+    document.getElementById('from-filter-date').value = '';
+    document.getElementById('to-filter-date').value = '';
     scrollToTop();
   }
 
--- a/packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
+++ b/packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import styled from '@emotion/styled';
-import { designSystem, IsoDateFormat } from '@commercetools-docs/ui-kit';
+import { designSystem } from '@commercetools-docs/ui-kit';
 import SpacingsStack from '@commercetools-uikit/spacings-stack';
 import SpacingsInline from '@commercetools-uikit/spacings-inline';
 import useReleaseNotesFilterParams from '../hooks/use-release-notes-filter-params';
@@ -48,11 +48,19 @@ const ClearAll = styled.button`
   }
 `;
 
+/**
+ * Return the date in yyyy-mm-dd format
+ */
+const isoYMD = (inDate) => {
+  const date = inDate || new Date();
+  return date.toISOString().substring(0, 10);
+};
+
 const ReleaseNotesFilterDates = () => {
   const [filterParams, setFilterParams] = useReleaseNotesFilterParams();
-  const [fromDate, setFromFilterDate] = useState();
-  const [toDate, setToFilterDate] = useState();
-  const maximumDate = IsoDateFormat.format(new Date());
+  const [fromDate, setFromFilterDate] = useState('');
+  const [toDate, setToFilterDate] = useState('');
+  const maximumDate = isoYMD();
 
   useEffect(() => {
     if (filterParams.fromFilterDate) {
@@ -116,29 +124,25 @@ const ReleaseNotesFilterDates = () => {
   function handleOnFromFilterDateChange(e) {
     let date;
     try {
-      date = IsoDateFormat.format(new Date(e.target.value));
+      date = isoYMD(new Date(e.target.value));
     } catch (err) {
-      return;
+      if (fromDate) {
+        setFromFilterDate('');
+      }
     }
-    if (date.length === 10) {
-      setFromFilterDate(date);
-    } else if (date.length !== 10 && fromDate) {
-      setFromFilterDate(undefined);
-    }
+    setFromFilterDate(date);
   }
 
   function handleOnToFilterDateChange(e) {
     let date;
     try {
-      date = IsoDateFormat.format(new Date(e.target.value));
+      date = isoYMD(new Date(e.target.value));
     } catch (err) {
-      return;
+      if (toDate) {
+        setToFilterDate('');
+      }
     }
-    if (date.length === 10) {
-      setToFilterDate(date);
-    } else if (date.length !== 10 && fromDate) {
-      setToFilterDate(undefined);
-    }
+    setToFilterDate(date);
   }
 
   function handleOnBlur(event, field) {
--- a/packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
+++ b/packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
@@ -34,7 +34,7 @@ const DateInputField = styled.input`
 `;
 
 const ReleaseNotesFilterDates = () => {
-  const [filterParams, setFilterParams] = useReleaseNotesFilterParams();
+  const [, setFilterParams] = useReleaseNotesFilterParams();
   const maximumDate = IsoDateFormat.format(new Date());
 
   return (
@@ -48,8 +48,6 @@ const ReleaseNotesFilterDates = () => {
             type="date"
             id="from-filter-date"
             max={maximumDate}
-            value={filterParams.fromFilterDate || ''}
-            onClick={handleClick}
             onChange={handleOnFromFilterDateChange}
           />
         </div>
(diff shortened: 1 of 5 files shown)


gabriele-ct (approved):
Looks good 👍

nkuehn (changes requested):
Summary:  overall nice and I am optimistic we can take this to the end, but:

 - please try to use the native min/max features
 - we have a bug with keyboard input in the year field. it works flawlessly in the barebones HTML spec example ( https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#value ) but in our implementation it takes only the first number types and then exits the field.  We need to revisit the `onChange` handler maybe?  I am a bit clueless. `onInput` fires even more often.  Something about React state and rerendering maybe? 
 - a small inconsistency in browsers is that some do show the picker on click and some don't. There is a new  `showPicker()` function on the element in the new living HTML standard documents but it's not a standard yet.  It is implemented in current chrome though.  We could try calling it  on click if it exists for maybe a little more consistent behavior and more similar to the previous UI which defaulted to a picker and not text input, too.  
  

----
Screenshots for the record (all current versions of browsers)

**current chrome on mac** (works reliably)
image


**current firefox on mac** (works but not really good for entering the year through keyboard) Windows behaves and looks identical
image


**current safari on mac** (works but also here the keyboard experience is nonfunctional in the year input). Windows behaves and looks identical. 
image

**Chrome on Android**:  very nice with the native picker, but no way to unset the date (reloading the page keeps it)

**Safari on iOS**:  very nice with the native picker but no way to unset the date, too (there is a "reset" UI element but it does not reset, maybe it just resets the native picker to today's date? )

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
```diff
@@ -28,29 +44,38 @@ const ReleaseNotesFilterDates = () => {
       <SpacingsStack scale="xs">
         <DateLabel htmlFor="from-filter-date">From</DateLabel>
         <div>
-          <DateInput
+          <DateInputField
+            type="date"
             id="from-filter-date"
+            max={maximumDate}
             value={filterParams.fromFilterDate || ''}
+            onClick={handleClick}
```
timonrey, line 52: I added the `showPicker` function only to the first date input so that we can test it and decide if we want to have it or not. When the picker is open, there is no way to write the date directly into the field, you would have to select it via the picker (in chrome). If we don't use it, you would still be able to write it directly into the field (which is still a bit problematic because of the bug we mentioned above).
Preview:
https://docskit-pr-1446.commercetools.vercel.app/docs-smoke-test/releases

timonrey: > we have a bug with keyboard input in the year field. it works flawlessly in the barebones HTML spec example ( https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#value ) but in our implementation it takes only the first number types and then exits the field. We need to revisit the onChange handler maybe? I am a bit clueless. onInput fires even more often. Something about React state and rerendering maybe

I tried to find out what the problem is and even without the `onChange` handler, the input field gets submitted (or reseted) every time you reach the year field. I also looked into the ui-kit component and I noticed that the behaviour is slightly different. It seems like the date is only being submitted after pressing enter.

@emmenko maybe you have an idea how we can solve this?

nkuehn: > even without the onChange handler, the input field gets submitted (or reseted) every time you reach the year field. 
That's really weird given that the plain HTML examples in the HTML reference docs don't do that.  Does React or some library we use magically attach some kind of even handlers or does some parent component of this input form trigger a re-render in certain situations?

gabriele-ct: @timonrey 
This strange behaviour definitely has to do with the fact that the field is "controlled"meaning that the value is set by `filterParams.fromFilterDate` which is in turn changed by the handleOnFromFilterDateChange. Probably in some case the date is either not valid or it causes the render in the field to become weird and the field starts misbehaving. 
Try to comment out `value` and `onChange` props, and you'll see it working as expected.
One solution could be to call `setFilterParams` only when the date has a valid date value.

Let me know if you need further support on this

nkuehn: > One solution could be to call setFilterParams only when the date has a valid date value.

That is a good idea in any case because there should never be situations where are shareable URL contains an invalid date.  But be aware that it's not exactly trivial to "validate".   E.g.  the case that my UI stuck on was  0020-02-21 ,   that is a valid date.  It's roundabout where christ was born but formally a valid date.  

maybe you need to introduce component level state ( useState // useEffect based) and not only rely on the URL being the UI state.   If the state "loop" is managed inside react it the "controlled" aspect mabye behaves more predictable because it does not loop through the browser engine any more but react has full control?

nkuehn:
@timonrey  concerning the problem that resetting the fields is not consistently prossible in the native UIs you could introduce the identical "clear"  link like the one above the filter fields  (or, maybe even better,  move that "clear all" link up to the top of the form and reset the whole form - I found that slightly weird anyways)

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
gabriele-ct: Here's an example (ignore the use of `moment`)
The idea is: you apply your logic, if the date is valid, you format it and you pass it as value to the input
https://codesandbox.io/s/funny-chebyshev-nf9dv?file=/src/App.js

timonrey:
@gabriele-ct I have implemented all the suggested changes and I managed to fix the bug with your help 🙌🏼 
The only thing that is still not perfect is that the clear button does not clear the date inside of the input fields. It clears the release not filters for dates and topics as expected, but the input fields still have the date inside.

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
```diff
@@ -17,48 +17,107 @@ const DateLabel = styled.label`
...
+      toFilterDate: undefined,
+      filterTopics: [],
+    });
+    setFromFilterDate(undefined);
+    setToFilterDate(undefined);
     scrollToTop();
   }
 
+  function handleOnFromFilterDateChange(e) {
+    const date = IsoDateFormat.format(new Date(e.target.value));
```
gabriele-ct, line 108: This line is potentially unsafe, in case `e.target.value` gets some non date weird value.
Validating a date it's not trivial so I'd just add a `try..catch` block, returning in case of catch.
Same thing in the `handleOnToFilterDateChange` function.

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
```diff
@@ -17,48 +17,107 @@ const DateLabel = styled.label`
...
+    if (date.length === 10) {
```
gabriele-ct, line 109: for safety let's check that date is not undefined
`if (date && date.length === 10) {`

gabriele-ct (changes requested):
I'd add some additional safety checks

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
timonrey: The date input only allows numbers to be written, and once the date has 10 chars, we can be sure that it's at least a valid date format. But if you think it would make sense, I can add it 👍🏼

packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js
```diff
@@ -93,7 +92,6 @@ const ReleaseNotesFilterDates = () => {
             type="date"
             id="to-filter-date"
             max={maximumDate}
-            value={toDate}
```
gabriele-ct: @timonrey Removing value attribute and setting the values directly using DOM selector is a code smell, it means that something else is not working properly. I'll try to have a better look later but this last commit seems not to go in the right direction

gabriele-ct: @timonrey I pushed a commit. The sync between the local state and the application state happens onBlur. I think you can start having a look to it once it's build and fine tune it. But try to stay away from direct DOM interactions unless there's no other solution.

HTH.

nkuehn:
@gabriele-ct  thank you for finding the problem.  The main interaction with the date pickers is fine now. 

@timonrey  the "clear all" only clears the date input now, that was not really the idea behind moving it.

gabriele-ct (approved):

gabriele-ct:
@timonrey good to go for me.
I've fixed some corner cases and simplified a little bit the code

nkuehn (approved):
Thank you, it's very nice now - in fact in some aspects better than before (keyboard input, clear all for all inputs)

PR #1448, opened 2022-11-30
feat(api): support union type

gabriele-ct:


gabriele-ct:
Previous behaviour:
Screenshot 2022-12-01 at 10 55 48

Current behaviour:
Screenshot 2022-12-01 at 09 41 27

packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
```diff
@@ -18,6 +19,20 @@ import { typography } from '../../../design-system';
 const isRegex = (string) =>
   string.charAt(0) === '/' && string.charAt(string.length - 1) === '/';
 
+const isTypeUnion = (strType) => {
+  return typeof strType === 'string' && strType === 'Union';
```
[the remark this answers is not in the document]
gabriele-ct: I set this myself: https://github.com/commercetools/commercetools-docs-kit/pull/1448/files#diff-ef887826b9c4c7b2ac90b9c6ad39a3e32b7166a99aad307eeba76d8f8fd64039R7

packages/gatsby-transformer-raml/src/utils/type/generate-type.js
```diff
@@ -1,4 +1,11 @@
+function isUnion(type) {
+  return typeof type.type === 'string' && type.type.indexOf('|') >= 0;
+}
+
 function generateType(type) {
+  if (isUnion(type)) {
+    return 'Union';
```
nkuehn, line 7: @gabriele-ct  OK, understand (referring to the thread above).   Conventionally, primitives are lowercase in RAML and we conventionally name custom types PascalCase   

https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md/#built-in-types

For consistency, please check where the renaming of the RAML primitives names into our display convention happens (graphQL server or frontend code).  E.g.  `integer` becomes `Integer` somewhere and `string` turns into `String`  (and stuff like a  `number` with `format: float` becomes `Float` in the UI)

gabriele-ct: Thank you @nkuehn ,
this _renaming_ happen precisely in the function you're quoting:
https://github.com/commercetools/commercetools-docs-kit/blob/3c7f7f7567d79a4fe40cc9bd210a8e58d349a4c8/packages/gatsby-transformer-raml/src/utils/type/generate-type.js

The detection of `isUnion` happen at this stage too, so I decided to use capitalized `Union` as it's the result of a processing of raml type, it seemed coherent to me.

nkuehn: OK!  So the whole renaming happens "server-side" anyways (build-time technically spoken)

nkuehn (approved):
Nice!  Another topic checked of for the week :-)

PR #1453, opened 2022-12-05
feat(api code): sticky code examples column

gabriele-ct:


packages/gatsby-theme-docs/src/components/side-by-side.js
```diff
@@ -33,9 +38,20 @@ const SideBySideContainer = styled.div`
 
 const SideBySide = (props) => {
   const pageData = usePageData();
+  const siteData = useSiteData();
 
   if (pageData.allowWideContentLayout)
-    return <SideBySideContainer>{props.children}</SideBySideContainer>;
+    return (
+      <SideBySideContainer
+        stickyMargin={
```
gabriele-ct, line 46: bare minimum margin. Set to 0 when there's no global notification banner, otherwise it's set exactly to the notification height (including padding). We could decide to remove this logic and just add a fixed higher margin (like the example in Stripe does)

packages/gatsby-theme-docs/src/layouts/internals/page-content-inset.js
```diff
@@ -16,7 +16,7 @@ const PageContentInset = styled.div`
   max-width: ${(props) =>
     props.maxWidth || designSystem.dimensions.widths.pageContentWide};
   section {
-    overflow: auto;
+    overflow: clip;
```
gabriele-ct, line 19: `overflow: auto` has been added by this PR: https://github.com/commercetools/commercetools-docs-kit/pull/1420/files

I've verified that the active selection still works as expected.

Reason behind `clip` value:
CSS `position: sticky` works only when there's no overflow restrictions applied at any level in the ancestors. see https://www.terluinwebdesign.nl/en/css/position-sticky-not-working-try-overflow-clip-not-overflow-hidden/

As possible solution `clip` is a sort of middle ground that seems to be working solving both the issues (active section and sticky content)

gabriele-ct:
Ready for first round of review. Example page: https://commercetools-docs-84dfah1in-commercetools.vercel.app/api-docs-smoke-test/layouts/endpoint-large-desktop

packages/gatsby-theme-docs/src/components/side-by-side.js
nkuehn: I suggest a top margin, definitely in the case without the top banner but also a little more when used with the top banner.  Here's an example from the docs kit docs:

image

Given that the primary use case is in the API docs I think  the `1rem`  padding that the code example has towards the grey background area anyways would be one logical choice.  In other contexts in body content, 0.5rem is a common padding too depending on the situation but I think it's visually better to have a bit more spacing in doubt. 

@Shecki opinons?  That was a long standing feature wish from you, so you have that last word :-) .  See this page and make the screen very wide: 
https://commercetools-docs-84dfah1in-commercetools.vercel.app/api-docs-smoke-test/layouts/endpoint-large-desktop

nkuehn (commented):
Wow, it's beautiful and CSS is great.   

But you're right, I think it needs more spacing.

packages/gatsby-theme-docs/src/components/side-by-side.js
gabriele-ct: `1rem` looks good to me, in both the situations @nkuehn .
if you agree we can go ahead with this spacing and then fine tune it if needed

gabriele-ct:
@nkuehn 
Sticky behaviour applies only to api endpoint code and types examples. The previous implementation was targeting too generically the second item in the side by side which was targeting any content (I noticed it in the video wide screen page example).
Having added the `data-is-wide-sticky` data attribute logic, will give us more granularity on what we want to be sticky and what not

packages/gatsby-theme-docs/src/layouts/internals/page-content-inset.js
nkuehn: nice research.  Please consider documenting the rationale as a code comment. This is the kind of things that get really hard to re-engineer if later someone else touches the CSS.

nkuehn: Update:  in most pages it's an increased spacing when an H2 heading is directly following an H1 heading.  But there's also one case in the images docs where it's in a place where an H2 heading is after an image.

gabriele-ct: @nkuehn The VRT changes you see are basically reverting the VRT changes introduced with the PR regarding selected section. (the cause is the change of the overflow property).
As you can see here are the original VRT changes:
https://percy.io/commercetools-GmbH/docs-kit-smoke-tests/builds/23092707/changed/1289339663?browser=chrome&browser_ids=31&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=1706&widths=1706

and here the latest:
https://percy.io/commercetools-GmbH/docs-kit-smoke-tests/builds/23496946/changed/1313464350?browser=chrome&browser_ids=31&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1706&widths=1706

Basically we're reverting the UI changes but the functionality still works as expected

PR #1457, opened 2022-12-09
fix(tables): gracefully degrades links in table headers in mobile view

gabriele-ct:


packages/ui-kit/src/components/markdown.tsx
```diff
@@ -171,9 +185,9 @@ const Table = styled.table`
       return React.Children.toArray(rowHeadersChildren).reduce(
         (styles, elem, index) => `
         ${styles}
-        td:nth-of-type(${index + 1})::before { content: "${
-          reactIs.isElement(elem) && elem.props.children
-        }"; }
+        td:nth-of-type(${
```
gabriele-ct, line 188: injecting content using CSS has certain limitations https://developer.mozilla.org/en-US/docs/Web/CSS/content
I'm not sure if there's a safe way to inject an actual a href link, I haven't found one. Maybe there's workaround but I feel that if we **need** proper links support in the headers for mobile, we should consider creating a dedicated "adaptive" component that renders only in certain viewports.

Otherwise my opinion is that stripping any link is the safest bet considering we're not targeting primarily mobile platform.

I'm open to other options/solutions

gabriele-ct:
See a working example here:
https://commercetools-docs-o9nb1g3tu-commercetools.vercel.app/docs-smoke-test/views/markdown#tables

The idea is to render the link only when not in mobile view and fallback on  text only for mobile

nkuehn (approved):

Issue #1181, opened 2022-01-03
Unwanted hyphen at the end of Anchor Links on Beta tagged headers


timonrey:
Anchor links for Beta-tagged entries have a hyphen (-) added to the end of their `href`.

![image](https://user-images.githubusercontent.com/77231096/147738096-f32f8806-4f36-4b9c-a8e4-66600619179f.png)

Transfered issue from the [docs repo](https://github.com/commercetools/commercetools-docs/issues/1674)

timonrey:
Slack discussion for this issue:
https://commercetools.slack.com/archives/GG00MD84Q/p1640802801001600

timonrey:
@adinakleine @industrian Can you please describe the practical problem for you?

industrian:
> @adinakleine @industrian Can you please describe the practical problem for you? It would be quite complicated to fix and we would like to not do it if its just for aesthetical reasons.

Adding @stmeissner 

This problem essentially breaks links. But doesn't count as a broken link as the page itself loads.

Example:

1. Go to https://docs.commercetools.com/api/projects/cartDiscounts 
2. Click CreatedBy (in the screenshot). 
![image](https://user-images.githubusercontent.com/77231096/148200326-094777f1-e22d-4e59-9aa2-b9ae5a63814e.png)
3. This link expects to go to an anchor called "createdby", but as this anchor doesn't exist, it just takes you to the top of the Common API Data Types page.

I'm not sure how these pages are generated, but the issue appears to be that a space is being added to the title. This space is then causing a hyphen to be added to the anchor links.

![image](https://user-images.githubusercontent.com/77231096/148200992-0423250b-c557-4600-9344-116a0974de53.png)

nkuehn:
@industrian  the contente that has the link is handwritten, that means the author has to put the exact achor string of the target location.  If the target location has a dash at the end the anchor in the hadwritten markdown has to have a dash at the end.  It's pretty mechanic work, yes. 

My conclusion from the ticket is that this is more of an author ergonomics issue than a bug.   It's not intuitive that some target anchors have a dash at the end.  And it's annoying to have to touch all manual links if something goes out of beta. 

I'd keep it as a feature request to try to somehow wrap the standard library used for "slugify" of headings and text to avoid the extra dash.  It's happening in many places in the codebase and therefore needs thorough testing.   I would prefer to do it after we have anchor link testing back ( #1185 )

industrian:
[...]

PR #1459, opened 2022-12-14
feat(mdx): Upgrade to mdx v2

gabriele-ct:
addresses #1472


gabriele-ct:
@nkuehn, @timonrey  here some recap to where I got:
To start with I'm focusing on making the mdx v2 working on the docs smoke test website which is sort of the most simple (it doesn't use introspection)
I followed the `gatsby-plugin-mdx` plugin upgrade guide (checked if all is fine several times). I got to the point that the build (for the specific site) works with no errors, but I can't get any content on the page, apart from the content coming from frontmatter tags.
The mdx plugin seems to be working fine, at least the compile part: I get some warning if I introduce some non mdx v2 compliant content on a mdx file.
I even created a super basic page example (`views/mdxtest.mdx`) to see if for any reason the compiler was failing on more complex pages but still, no luck.
One of the most significant change is this: https://github.com/commercetools/commercetools-docs-kit/pull/1459/files#diff-020a60f4c5bb2f367453ebce61d0589acb834c8e58f9fe0569d04f52f033c451R32
documented in here https://www.gatsbyjs.com/plugins/gatsby-plugin-mdx/#updating-page-templates
Basically the plugin should inject a `children` prop into the layout component with the transformed body content of the page. However I do see that prop injected, but it's undefined all the times.

I have tried to disable all the remark plugins, all the rehype plugins, basically use the base config of `gatsby-plugin-mdx`, but still I don't see any content being transformed.
I'm at the stage that I'm blocked, so I'm wondering if anyone has any idea about what I'm doing wrong

gabriele-ct:
Latest findings:
- opposite to what we thought initially, it's correct to use `require.resolve` which is *not* actually resolving the component but returning the resolved path as string. require works correctly and it's what we actually want in this case
- What was wrong it was the actual graphql result destructuring which was resulting in an undefined property. Now it works fine

Where we are?
- Most of the pages are working correctly displaying the content as expected.

What is still not working?
- Import of mdx components  such as:

> import ListItems from "../../topics/list-items.mdx"
> 
> <ListItems />

is still not working, it's throwing an warning at compile time.

It seems to be more related to the webpack config but I cannot find a solution

nkuehn:
Nice!   I'm sorry that I led you on the wrong path. 

`<Card>` only being allowed to live inside `<Cards>` was originally intentional, we just recently started abusing them as a general purpose "frame around things".   FYI @timonrey  - maybe the solution to the problem with the shadow is to rather dial back what I changed to the `<Cards> ` CSS and introduce a separate `<Box>` or something.  

Crazy to hear that many things already render :-)

Battling the ESM imports problems at build time might turn out to be a blocker by the way, maybe have an eye on that before tackling too many details.   MDX v2 is ESM-only and in the past I was unssuccessful to use ESM only modules in the project in other places.

nkuehn:
@gabriele-ct  I was curious and tried it out locally.  I pushed one trivial fix to enable tables again and clicked around. 

The issue with the mermaid diagrams is most likely to be found in https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/gatsby-theme-docs/src/plugins/remark-mdx-mermaid.js  , which generates MDX v1 AST which is now not properly understood any more and hence rendered as text.   

I think the spec is here: https://github.com/syntax-tree/mdast-util-mdx-jsx#syntax-tree  

The new node type for a block element would be "mdxJsxFlowElement", but please check.  We may need to adjust more of our custom AST transformers too,  check for everything that handles  `type: "jsx"` nodes

packages/gatsby-theme-docs/utils/api-tag-info.js
```diff
@@ -0,0 +1,25 @@
+const unified = require('unified');
```
gabriele-ct, line 1: @nkuehn didn't get that far... unfortunately.

Limiting to only methods.mdx file for the moment. If I attempt to parse the file as is. It fails with an error in one of the remark-mdx dependencies.

If I simplify the code removing the `\n` in `ApiEndpoint` tag. I only get the headings tags in the AST.

any ideas?
You can try it out by pulling the branch and running `rm -rf .cache && yarn start` in the `api-docs-smoke-test` microsite

gabriele-ct: thank you @nkuehn.
re the missing topic folder, I have solved adding an empty topic folder to the api microsite locally (forgotten about git and empty folders 🤦 ).

I had had a look to the `createProcessor` but in the example https://github.com/unifiedjs/unified#processorparsefile seemed to do exactly the same thing I was doing using unified, so I didn't try it.
As you say, it's probably a side effect of non compatible versions.

I'll move on tomorrow starting from here. Thank you.

nkuehn: If we require folders we generally autogenerate them at startup, too to avoid confusing errors for authors.   At the top of `gatsby-node.js` there's some code that ensures the minimum conventional filesystem structure for a site. 

The docs about the processor API you linked are actually weird and a bit misleading because if you get a processor instance you don't actually have to do the chaining starting from `unified().`.  Maybe they just copy / pasted some documentation :-)

nkuehn:
@gabriele-ct  nice, at least it builds.  There's quite some artifacts though in the content.  Just  quick first partial scan
 - extra space at the top of every page content (the index nav is not top aligned any more with the content body start). percy shows this at the very bottom of the page but it's not actually there. Needs to be resolved. Maybe MDX v2 adds some wrapper around the content? 
  - nested content in quotation blocks:  Does not have to work, this was really just a smoke test
  - issues with certain code examples (for example the markdown one becoming actual markdown).  This needs to be robust because it's used a lot.

nkuehn:
I expected TOC issues - it's not unlikely that you have to re-do the slug generation for the anchor links manually.    Might address #1181 as a side effect

gabriele-ct:
@nkuehn The TOC issues seem to have been fixed with the latest development.
I've been looking at the percy issues and I could track one down to the infostring parsing (https://github.com/mdx-js/mdx/pull/257)
Basically it seems that the functionality above is not working with the mdx upgrade. The result is that any inline attribute added to the mdx file is not passed as prop to the underlying React component.

e.g.
```md title="you write:" secondaryTheme
```

doesn't get any `title` or `secondaryTheme` prop.

The issue doesn't seem to be under our control, unless I'm missing something. What do you think?

gabriele-ct:
> Do you get the language and the raw infostring (=metastring) somewhere at all?

That's the issue. We don't get the metastring, at least from what I can see...

gabriele-ct:
@nkuehn in the https://mdxjs.com/playground/ 
you can see how, for example, the following code gets compiled:
```
```md title=hello themeTwo
This is some code
```
```

At `mdast ` level the title and theme information are available, but they don't get passed to the complied jsx version ...
Shall we somehow pre-process the ast on node create?

**Comment by nkuehn** [comment]
@gabriele-ct  thanks for the link to the playground. It really seems the "meta" informat in the mdast is not passed on to the the components in the compiled output.  

Lots of issues and discussions around "meta" and "code": https://github.com/mdx-js/mdx/search?q=code+meta&type=issues .  The MDX maintainers pointed to. 

https://mdxjs.com/guides/syntax-highlighting/  and  https://github.com/remcohaszing/remark-mdx-code-meta  for further reading.   Looks like they intend to have us use a plugin  or extend ourselves because it's not a hard part of the MD standard.

**Comment by gabriele-ct** [comment]
@nkuehn 
Despite the initial troubles in making the es6 module working, it seems now to be loading and executing the plugin.
However, the jsx compilation fails. And here comes my first question:
- should the plugin be loaded as `remarkPlugins` in  the `gatsby-plugin-mdx` (as I did)? or it should be used somewhere else?

I added some console logs into the plugin itself and the AST tranformation happens as expected, however I cannot fully understand why is triggering the compiler error.
I pushed the changes so you can have a look locally.

The transformer code is very simple:
https://github.com/remcohaszing/remark-mdx-code-meta/blob/main/index.ts

**Comment by gabriele-ct** [comment]
also, on a separate topic. It looks like `dl` `dt` `dd` tags are not correctly mapped to the react components defined in `packages/gatsby-theme-docs/src/markdown-components.js`

```
dl: Markdown.Dl,
  dt: Markdown.Dt,
  dd: Markdown.Dd,
```

I tried to figure out why but I could not find the reason. That is affecting the styling resulting in VRT issue

File: packages/gatsby-theme-api-docs/src/components/type/enum.js [hunk 2]
```diff
@@ -23,7 +23,7 @@ const Enum = ({
         {description && (
           <DescriptionParagraph>{description}</DescriptionParagraph>
         )}
-        <Markdown.Dl>
+        <Markdown.DescriptionList>
```
gabriele-ct, line 26: As agreed I changed the `<d*>` components name to something more descriptive name

packages/gatsby-theme-code-examples/src/multi-code-example.js
```diff
@@ -7,18 +7,18 @@ import {
 } from '@commercetools-docs/ui-kit';
 import useCodeExamples from './use-code-examples';
 
+const isCodeExampleNode = (alienNode) =>
```
gabriele-ct, line 10: In a couple of place in the code, we need to check what's the React type of a specific component. Unfortunately `component.type`, in case of functional components is, indeed, a function which name gets mangled in minification. 
`component.type.name === 'MyComponent'` works in dev mode but not once built.
So the problem I'm trying to address here is to `guess` if the component is a `CodeExample` component by checking ig the mandatory `href` prop is defined.

It's weak but I couldn't find any better soultion

packages/gatsby-theme-api-docs/src/components/type/enum.js
nkuehn: How they are named internally in the "Markdown" module does not really matter, what matters is under what name they are injected into the MDX provider.    (e.g. the "Markdown." prefix is never exposed)

packages/gatsby-theme-docs/bin/create-release-note.js
```diff
@@ -125,7 +125,7 @@ topics:
 ${response.topics.map((topic) => `  - ${topic}`).join('\n')}
 ---
 
-Please write the content here. To use "Read More", insert <!--more-->.
+Please write the content here. To use "Read More", insert {/* more */}.
```
gabriele-ct, line 128: [EDITOR ALERT] This is a syntax change that needs to be notified to the editors

packages/gatsby-theme-code-examples/src/multi-code-example.js
nkuehn: Good enough!.  As said, it's just a small optimization to catch most common types of errors before the dev server crashes or the app breaks otherwise

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -5,6 +5,15 @@ const defaultOptions = require('./utils/default-options');
 
 const isProd = process.env.NODE_ENV === 'production';
 
+const wrapESMPlugin = (name) =>
```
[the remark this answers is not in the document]
nkuehn: This is great!  Would it be horrible scope creep to do this with the other remark / rehype plugins in one go and remove all those pesky manual renovate version constraints?

packages/gatsby-theme-docs/src/components/card.js
```diff
@@ -47,6 +24,68 @@ WrapWith.propTypes = {
   children: PropTypes.node.isRequired,
 };
 
+const isReactLink = (alienElement) =>
```
gabriele-ct, line 27: same case as previously described. Need to check if a React element is a `Link`. I do it by checking if it contains an `href` prop

packages/gatsby-theme-docs/src/markdown-components.js
```diff
@@ -41,9 +41,6 @@ const components = {
   ul: Markdown.Ul,
   ol: Markdown.Ol,
   li: Markdown.Li,
-  dl: Markdown.Dl,
```
gabriele-ct, line 44: we don't support direct use of `<d*>` tags anymore

websites/api-docs-smoke-test/src/content/endpoints/endpoints-for-resource.mdx
```diff
@@ -5,26 +5,26 @@ description: Test endpoints
 
 import { ApiEndpointsForResource } from "/shortcodes"
 
-# /{projectKey}/resource
+# /\{projectKey}/resource
```
gabriele-ct, line 8: [EDITORS ALERT] This syntax change needs to be notified to the editors

gabriele-ct:
Ready for a first round of review. I still need to add mdx v2 eslint plugin and decide how to handle the deprectaiton of `gatsby-transformer-mdx-introspection` package

packages/gatsby-theme-docs/gatsby-config.js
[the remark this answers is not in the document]
gabriele-ct: I can give it a try. Hopefully it will not create huge breaking changes

nkuehn:
> how to handle the deprectaiton of `gatsby-transformer-mdx-introspection` package
Suggestion:  
3. tell changesets that this package changes (patch version)
1. change the repository URL in package.json  of the introspection plugin to `https://github.com/commercetools/commercetools-docs-kit/tree/@commercetools-docs/gatsby-transformer-mdx-introspection@15.0.5/packages/gatsby-transformer-mdx-introspection`  (last current release, alternatively use the release that this change here will product). This makes this the source URL in the package.  
2. change README.md with a quote at the top telling that the package is deprecated because the functionality can now easily be achieved directly through reading from the MDX v2 AST. 
3. merge and publish together with MDX v2
4. I or nicola go to npmjs.com and change the package to deprecated (optional)
5. new PR, removing the code.  The final NPM release will point to the tag that is the last code version

websites/api-docs-smoke-test/src/content/endpoints/endpoints-for-resource.mdx
nkuehn: That's a bit pesky but we have no chance.   The curly brace notation for project key and region are all over the place in the docs.

packages/gatsby-theme-docs/src/markdown-components.js
```diff
@@ -59,6 +56,9 @@ const components = {
   a: Link,
 
   // Official react components to be used in MDX files
+  DescriptionDetails: Markdown.DescriptionDetails,
+  DescriptionTerm: Markdown.DescriptionTerm,
```
nkuehn, line 60: See above, maybe: 

```suggestion
  Details: Markdown.DescriptionDetails,
  Term: Markdown.DescriptionTerm,
```

packages/gatsby-theme-docs/utils/api-shortcode.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const shortcodes = []; // {name: shortCodeName, attributes: [{name: foo, value: bar}]}
+
+  const isShortcode = (node) => {
+    return (
+      node.type === 'mdxJsxFlowElement' && shortcodeList.includes(node.name)
+    );
+  };
+
+  // https://unifiedjs.com/explore/package/unist-util-is/#test
```
nkuehn, line 18: I think this comment line is obsolete.  I originally used unist-util-is but then did it as the custom function in lines 12-16

packages/gatsby-theme-docs/utils/api-shortcode.js
```diff
@@ -0,0 +1,38 @@
+// loaded via require because we use the pre-ESM version at the moment.
+// to be changed once upgraded to a latest version
```
nkuehn, line 2: revisit if the other loading strategy turns out to work.

websites/docs-smoke-test/src/content/views/beta.mdx
```diff
@@ -2,7 +2,7 @@
 title: Beta page
 beta: true
 showTimeToRead: true
-timeToRead: true
```
nkuehn, line 5: I think this was a smoke test against authors confusing the two time to read frontmatters.

websites/docs-smoke-test/src/topics/complex-blockquote.mdx
```diff
@@ -1,11 +1,16 @@
+---
+title: Complex blockquote
```
nkuehn, line 2:
```suggestion
title: Nested markdown in blockquote
```

websites/documentation/src/content/writing/markdown.mdx
```diff
@@ -667,14 +667,14 @@ Only icons provided in the [UI Kit](https://uikit.commercetools.com/?path=/story
 
 ```md title="you write:" secondaryTheme
 <Cards>
-  <Card image={<CardBannerMcDocsIcon />} icon={<DummyIcon />}>
+  <Card image={<CardBannerMcIcon />} icon={<DummyIcon />}>
```
nkuehn, line 670: Thank you for fixing the docs here!

nkuehn (commented):
Looks like it's overall in very good shape already!  A bunch of comments concerning things that look like leftovers only.

packages/gatsby-theme-api-docs/src/components/type/enum.js
```diff
@@ -32,20 +32,20 @@ const Enum = ({
...
-                  </Markdown.Dt>
+                  </Markdown.DescriptionTerm>
                   {enumDescription && enumDescription.description && (
-                    <Markdown.Dd>
+                    <Markdown.DescriptionDetails>
                       <DescriptionText
                         markdownString={enumDescription.description}
                       />
-                    </Markdown.Dd>
+                    </Markdown.DescriptionDetails>
```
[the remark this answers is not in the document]
gabriele-ct: 👍

nkuehn:
P.S. to my review:  What we need to test, document and maybe try to optimize is the behavior of the dev server when a file with syntax errors is saved. At the moment the whole dev server is not recoverable although we have added error handling that at least shows a custom message and tells users to restart the dev server.  Improving this would be a massive plus

packages/gatsby-theme-docs/gatsby-config.js
```diff
@@ -110,6 +119,14 @@ module.exports = (themeOptions = {}) => {
...
         },
       },
+      // Topics content pages (.mdx)
+      {
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'topics',
+          path: path.resolve(`./src/topics`),
+        },
+      },
```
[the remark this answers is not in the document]
gabriele-ct: TBH this is really something it's not clear to me. 
This part wasn't here in the first place but I started seeing build errors like the following one if I don't add this entry:
Screenshot 2023-01-23 at 11 47 30

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -74,7 +76,10 @@ const resolverPassthrough =
       id: source.parent,
     });
     if (mdxNode) {
-      const field = type.getFields()[fieldName];
+      const field = type.getFields()[fieldName]; // TODO: migration
```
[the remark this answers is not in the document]
gabriele-ct: I had just threw some todos around in the very beginning in places that I felt could break or needed revision. This is actually fine

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -251,7 +270,7 @@ exports.onCreateNode = (
     parent.internal.mediaType === 'text/mdx' &&
     parent.sourceInstanceName === 'releaseNotes';
   if (isReleaseNotesPage) {
-    const excerptSplit = node.rawBody.split('<!--more-->');
+    const excerptSplit = node.body.split('{/* more */}');
```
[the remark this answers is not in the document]
gabriele-ct: added to the wiki

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -285,6 +305,10 @@ exports.onCreateNode = (
     return;
   }
 
+  const { createProcessor } = await import('@mdx-js/mdx');
+  const processor = createProcessor();
```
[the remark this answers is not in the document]
gabriele-ct: yes, I moved it to `onPreBootstrap` as I need an async function. It now gets executed only 1 time

packages/gatsby-theme-docs/gatsby-node.js
```diff
@@ -330,15 +361,17 @@ exports.onCreateNode = (
...
     return basePath;
   }
 
   if (node.frontmatter.slug) {
     return trimTrailingSlash(`${basePath}/${node.frontmatter.slug}`);
   }
 
-  const date = node.frontmatter.date ? node.frontmatter.date.split('T')[0] : '';
+  const date = node.frontmatter.date
+    ? node.frontmatter.date.toISOString().split('T')[0]
```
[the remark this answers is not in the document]
gabriele-ct: OK, it just feels weird to add a ui-kit dependency to a part of the gatsby build. Plus it's a trivial operation anyway and

websites/docs-smoke-test/src/content/views/beta.mdx
gabriele-ct: Ah, got it. I noticed it because is creating a warning in the gatsby build about types. But I reverted it

packages/gatsby-theme-docs/gatsby-config.js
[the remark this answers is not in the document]
nkuehn: **my first guess, but we can talk and try out if you want: **

That's a content error that became visible because MDX v2 seems to be stricter. 

Files inside `/topics` are not pages but intended to be imported into and re-used in multiple actual MDX files that are pages.   We decided on "topics" as a naming convention because in tech writing the concept of putting together documentation from prefabricated content  blocks is called "topic based writing".  

A fragment that is included must not have frontmatter, frontmatter has no relevant use case there beyond being a comment explaining what the file is and where it is to be used.    The error you see is MDX trying to parse the three minuses as MDX because no preprocessor was in place to do all the magic.  

I cannot tell why adding it to the filesystem source fixed it to be honest.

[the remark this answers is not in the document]
gabriele-ct: It's a good guess but it doesn't seem to be the case. I've just tried removing the frontmatter header, and the parsing error is still there:
Screenshot 2023-01-23 at 14 10 02

So, the reason must be something else. Definitely related to mdx2 strictness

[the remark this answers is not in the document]
nkuehn: Not necesssarily MDX v2 itself. Could also be the webpack loader not understanding that `.mdx` imports have to be compiled as MDX, not as JS.    It's still a parsing error, just that my guess was wrong what kind of parsing error

packages/gatsby-theme-docs/gatsby-config.js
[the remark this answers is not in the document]
gabriele-ct: in terms of what's directly used in `gatsby-config.js`
Screenshot 2023-01-23 at 15 33 49
I managed to upgrade all the remark/rehype plugins out of the box, except for `remark-gfm` which seems to be using a different pattern regarding what the default function returns. It might be possible upgrate it, but it's not effortless. Probably it's worth to upgrate to gatsby 5 which should support ES6.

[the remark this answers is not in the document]
nkuehn: nice!  We'll get that renovate configuration slimmed down piece by piece.

gabriele-ct:
@nkuehn I'm trying to make a list of missing points here, then we can decide if we want to tackle them in this PR or we can wait (creating a separate ticket).

1. `topic` folder: I couldn't find a reason for the failed compilation if I remove the folder from the `gatsby-config.js`. Do we want to dig deeper into this or we can accept the workaround of having these additional lines https://github.com/commercetools/commercetools-docs-kit/pull/1459/files#diff-b782e34a5e520693c45063e541ec7ac164c1f308b11974118d24c61f05df44a8R122 in `gatsby-config.js` ?
2.  mdx linter. I gave a go to `eslint-mdx`, I couldn't get decent results: a lot of complains about missing imports (all the gastby injected components). I just invested 30 minutes on it. Shall I insist or create a separate issue? 
3. Migration guide. I've created it here https://wiki.commercetools.com/display/DOCS/Mdx+v2+migration+guide. I'm not sure about the correct location, feel free to move it. About the content, I'm planning to add whatever I'll find while appying the theme-docs to the docs repo
4. Dev server error recovery. Probably I need to understand better what are your expectations here.  I tried to reproduce the scenario. If I introduce an error, the page live-reloads showing  something like this:
Screenshot 2023-01-23 at 21 54 30
If I fix the error, the live-reload happens showing the correct content on the page.
5. `gatsby-transformer-mdx-introspection` deprecation: I've made the change to the `package.json` file of the library, added the deprecation message in readme, we need to remember to do the remaining steps once the docs-kit is released

Probably I'm missing something, feel free to add points.

nkuehn:
> 1. `topic` folder: I couldn't find a reason for the failed compilation if I remove the folder from the `gatsby-config.js`. Do we want to dig deeper into this or we can accept the workaround of having these additional lines https://github.com/commercetools/commercetools-docs-kit/pull/1459/files#diff-b782e34a5e520693c45063e541ec7ac164c1f308b11974118d24c61f05df44a8R122 in `gatsby-config.js` ?

Let's stick to autogenerating the folder at startup and adding it to gatsby's filesystem knowledge as you put it.  It's a hard convention anyways so autogenerating is actually in sync with our general practice of autogenerating the conventional filesystem API structure.   You can create a ticket for later research or leave it because it's not a practical problem any more.  As a side effect, it prohibits people from doing arbitrary MDX imports outside "topics" so they have to stick to the convention.  It's a feature, not a bug :-) 

> 2. mdx linter. I gave a go to `eslint-mdx`, I couldn't get decent results: a lot of complains about missing imports (all the gastby injected components). I just invested 30 minutes on it. Shall I insist or create a separate issue?

I suggest to leave it.  If it implicitly assumes that the file is a self contained react component in MDX syntax then I think it's not worth constantly tracking and updating some hack.  

> 3. Migration guide. I've created it in the internal wiki. I'm not sure about the correct location, feel free to move it. About the content, I'm planning to add whatever I'll find while appying the theme-docs to the docs repo

(I removed the link to the internal wiki, this is a public repo).  I found the location and will review it there. Thanks !

> 4. Dev server error recovery. Probably I need to understand better what are your expectations here.  I tried to reproduce the scenario. If I introduce an error, the page live-reloads showing  something like this:
> If I fix the error, the live-reload happens showing the correct content on the page. 

I tested it and it's really working :-) Worth mentioning in the release notes and the authors will send you love letters for it.  Even the "Open in Editor" button works! 

> 5. `gatsby-transformer-mdx-introspection` deprecation: I've made the change to the `package.json` file of the library, added the deprecation message in readme, we need to remember to do the remaining steps once the docs-kit is released

Create a ticket 1SP if you think it may get forgotten.

nkuehn (commented):
Getting good. While writing on the migration notes the following came to my mind:
 
HTML tags:  we do use manual <br />  in some places, notably inside table cells. It's not really optional, so we have to provide a custom name for that to circumvent the sandboxing.  Maybe as a convention PascalCase it like our components?  `<Br />` or better use a special name to not confuse the hell out of people, for example `<LineBreak />`? 

Also HTML:  Do Entity sequences like `&lt;`  still work?   For example, for proper number formatting the style guide calls for  `&#8239;`  as a thousands separator.

gabriele-ct:
> HTML tags: we do use manual in some places, notably inside table cells. It's not really optional, so we have to provide a custom name for that to circumvent the sandboxing. Maybe as a convention PascalCase it like our components? `<Br />` or better use a special name to not confuse the hell out of people, for example `<LineBreak />`?

I vote for Pascal case with standard html tags. It's a lot easier rule to remember for editors and it's less confusing IMO. I had a quick  look into the docs repo and I found a good example with a table full of lists. I added `<Ul>`,`<Li>` and `<Ol>` tags and the rendering seems to work good
Screenshot 2023-01-24 at 11 02 03
Shall I just create a PascalCase mapping for all the html components defined in `markdown-components` or do you think we should have a more refined strategy, restricting to only the one actually used in the docs repo? Maybe doing a regexp serch through all the mdx files? @nkuehn 

>
> Also HTML: Do Entity sequences like `&lt;` still work? For example, for proper number formatting the style guide calls for `&#8239;` as a thousands separator.

Those still work as before:
Screenshot 2023-01-24 at 11 07 55

nkuehn:
Nice, thank you for testing! 

> Shall I just create a PascalCase mapping for all the html components defined in markdown-components 

I don't think that is a good idea and would encourage using HTML just because of personal opinions.  I suggest to only expose PascalCase wrappers for elements that we know we need to use in places where markdown cannot be nested.  To my knowledge that is only the table.  Proposal:
- ul
- ol
- li
- dl  (replacing our renamed components - if you want to introduce a PascalCase convention in general I would rather stick to HTML compatibility here, too)
- dt
- dd 
- br
- hr

things like bold, italic, strikethrough, images, links etc should normally work inside tables. And headings, tables, blockquote should not be used inside tables (like we restrict them in the markdown fragment renderer, too).   So there's not a lot remaining TBH.

timonrey (approved):

PR #1461, opened 2022-12-16
feat(markdown): add support to nested ordered list with different ind…

gabriele-ct:
…exes


gabriele-ct:
This is how markdown nested ordered lists are rendered:
Screenshot 2022-12-16 at 15 14 30

nkuehn (approved):

PR #1462, opened 2022-12-18
[feats] Enhancements for the request body

### LINKED ISSUE #1324: Error rendering Request Body when uploading image to Products
I'm currently working on the [RAML for Products PR](https://github.com/commercetools/commercetools-docs/pull/1933) and noticed the following error under Request Body:

image

This is how it appears in RAML (`products.raml`):

image

This is how correctly in the flattened version of the endpoint (in `ByProjectKeyProductsByIdImages.raml`):

image

I'm curious if this is caused by a missing feature in Docs Kit.

### LINKED ISSUE #1426: Display text for responses description and hide "No body is returned"/ generic type name
We need to display descriptions for two responses.
Furthermore, we should not display the "No body is returned" sentence for these responses (see screenshot below). 
As covered in [this PR](https://github.com/commercetools/commercetools-docs/pull/2481/files#diff-c00079ffab3ceb01b04de1349c64dc366929afbdd0b46bee9ddd92962733ffd7), even when the specs will be defined better, we will not return a specific type for these responses.
*NOTE:* as you can see from the PR mentioned above, in the future we will also provide examples.

FYI - the published version of the Frontend APIs docs renders "hand written" specs, not RAML files. The screenshot was taken locally.

```yaml
/{namespace}/{action}:
  uriParameters:
    namespace:
      displayName: Namespace
      type: string
      description: Namespace
    action:
      displayName: Action
      type: string
      description: Action
  is:
    - right-header
    - frontastic-locale
  get:
    is:
      - action-errorable
    description: Use the GET method to allow the frontend to fetch data asynchronously.
    responses:
      200:
        description: We recommend to use standard HTTP response codes and `application/json` encoded content.
  post:
    is:
      - action-errorable
    description: Use the POST method to write data to a backend system.
    responses:
      200:
        description: We recommend to use standard HTTP response codes and `application/json` encoded content.
```
image

### LINKED ISSUE #1430: Render request body of web forms
We have the need to document POST request bodies with content type: `application/x-www-form-urlencoded`

Given the following RAML specification for a  POST request in which the `properties` are listed individually (_format 1_):
```yaml
post:
  displayName: Search Products by post
  description: |
    The query parameters are included in the body encoded in `application/x-www-form-urlencoded`.  
  body:
    application/x-www-form-urlencoded:
      properties:
        fuzzy?:
          type: boolean
          description: |
            Whether to apply [fuzzy search](#fuzzy-search) on the text to analyze.
  responses:
    200:
      body:
        application/json:
          example: !include ../examples/product-projections-search.example.json
          type: ProductProjectionPagedSearchResponse
```

The Docs Kit could also support the below format with specifying the `type` that contains the properties (_format 2_):
```yaml
  body:
    application/x-www-form-urlencoded:
      type: SearchQueryParameters
```
This is currently supported for  `application/json` bodies, but it does not render it by mentioning the type as Link what would be consistent with payloads of other POST requests:

render-type-with-link

Instead, it lists all the properties of the specified type:

render-type

The latter rendering I suggest to apply for the case of _format 1_.
I also suggest to mention in the label for the **Request Body** something like _URL-encrypted key-value-pairs_.

timonrey:
This pull request implements requested enhancements:
- the content type is now displayed next to `body`
- the request and response bodies are were adjusted to support images
- descriptions inside of the response body are now displayed next to the response code

closes #1426
closes #1430 
closes #1324 

Link to the examples:
https://docskit-pr-1462.commercetools.vercel.app/api-docs-smoke-test/endpoints/endpoints-for-resource

Design ideas:

image

image (2)

image (1)

### DIFF
--- a/packages/gatsby-theme-api-docs/src/components/resource/method/method.js
+++ b/packages/gatsby-theme-api-docs/src/components/resource/method/method.js
@@ -33,6 +33,23 @@ const Container = styled.div`
 
 const TitleWithAnchor = Markdown.withCopyToClipboard(Title);
 
+function convertContentType(type) {
+  switch (type) {
+    case 'applicationjson':
+      return 'application/json';
+    case 'applicationxwwwformurlencoded':
+      return 'application/x-www-form-urlencoded';
+    case 'imagejpeg':
+      return 'image/jpeg';
+    case 'imagepng':
+      return 'image/png';
+    case 'imagegif':
+      return 'image/gif';
+    default:
+      return '';
+  }
+}
+
 const Method = ({
   apiKey,
   uris,
@@ -49,6 +66,26 @@ const Method = ({
   if (method.uriParameters) {
     allUriParameters = allUriParameters.concat(method.uriParameters);
   }
+
+  let contentType;
+  if (method.body) {
+    const findOutContentTypes = Object.keys(method.body).reduce(
+      (list, value) => {
+        return method.body[value] !== null ? [...list, value] : [...list];
+      },
+      []
+    );
+    findOutContentTypes.forEach((type) => {
+      if (!contentType) {
+        contentType = convertContentType(type);
+      } else {
+        contentType = `${contentType} ${convertContentType(type)}`;
+      }
+    });
+  }
+
+  const isImage = contentType ? contentType.includes('image') : false;
+
   const methodColor = computeMethodColor(methodType.toLowerCase());
 
   const id = generateEndpointURN({
@@ -108,18 +145,22 @@ const Method = ({
                     method.body.applicationjson?.type ||
                     method.body.applicationxwwwformurlencoded?.type
                   }
+                  isImage={isImage}
+                  contentType={contentType}
                 />
               )}
 
-              {method.responses && (
+              {!isImage && method.responses && (
                 <Responses apiKey={apiKey} responses={method.responses} />
               )}
             </SpacingsStack>
-            <RequestResponseExamples
-              apiKey={apiKey}
-              requestCodeExamples={method.codeExamples}
-              responses={method.responses}
-            />
+            {!isImage && (
+              <RequestResponseExamples
+                apiKey={apiKey}
+                requestCodeExamples={method.codeExamples}
+                responses={method.responses}
+              />
+            )}
           </SideBySide>
         </Container>
       </SpacingsStack>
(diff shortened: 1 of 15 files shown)


nkuehn (commented):
thanks for adding significant test coverage and getting the logic right.  I have two comment on the implementation pattern (I know, the issue is a legacy but it would be an opportunity to support any mime  type for good not). 

Visually I am not 100% convinced yet. 
 - the min-width between the Title and the mime type feels weird and random.  A fixed spacing between them would be better
 - vertical alignment of the two kinds of text (the code formatted looks higher up)
 - a list needs to be understandable as a list, needs separators and maybe more explicitly "or".    For example  "`image/gif` or `image/jped` or `image/png`"  ? 

image

packages/gatsby-theme-api-docs/src/components/resource/method/method.js
```diff
@@ -49,6 +66,26 @@ const Method = ({
...
+      []
+    );
+    findOutContentTypes.forEach((type) => {
+      if (!contentType) {
+        contentType = convertContentType(type);
+      } else {
+        contentType = `${contentType} ${convertContentType(type)}`;
+      }
+    });
+  }
```
gabriele-ct, line 85: my feeling is that this whole contentType discovery could be somehow simplified but I'm not getting the full logic behind it.
If you have time I'd recomend you extract this to a separate utils function, document it with a couple of lines of descriptions and maybe add few unit test just to make sure that any future refactor will not break.

nkuehn: The root cause lies in the design of the data structure in the GraphQL surface (IMHO).

timonrey:
We decided that we want to change the data structure of the GraphQL surface to address the root issue that was discussed above to not overload this pull request. I created an issue for it #1475.

nkuehn:
@timonrey  can we have a 10 minute session speaking about the layout? Something feels inconsistent about having the mime type on the Request but not the response as well as the mime type being in the "Request Body:"  heading line vs next to the details of what the Request is.  Maybe we can find a more logical / consistent setup together (sorry, visual trial & error here)

(plus one detail: the "or" between multiple mime types is not part of the mime type string so it should not be formatted as part of the string literal)

timonrey:
We decided to display the content type in the request body, as well as in the response body if the type is defined.

nkuehn:
Maybe post the "design" ideas / hacks here for reference for reviewers?

nkuehn:
@timonrey wow, it was worth the extra iteration. It's pretty clear and consistent now. 

I just found one visual nitpick:  If the description of a response type is longer than one line, the response code badge is vertically centered and not, analogous to the HTTP Verb,  top aligned:

image

nkuehn (approved):
Very nice feature, thank you!

PR #1466, opened 2022-12-21
fix: card border being cut off

### LINKED ISSUE #1442: Card(s) border and drop shadow cut off in certain card grid situations
We (maybe I?) seem to have introduced a regression in the Cards Grid that has slipped through the VRTs unnoticed although being pretty obvious.  Check the Cards smoke test for multiple examples:

https://commercetools-docs-kit.vercel.app/docs-smoke-test/components/cards

Here's some examples from the smke test.

image

image

timonrey:
The issue was introduced with this change that was made for the sticky example feature:
https://github.com/commercetools/commercetools-docs-kit/commit/86feac81f1127d2b5c9645147463a8d77bef33c3#diff-f1996a28d6f3f74e10c2bd9bac8c7e46a2a566b5549b06065d9f39d4e00ec1b0L19

I checked the sticky behaviour..it is still working as expected.
Please also have a look into the snapshots. They will show that not only the card borders are fixed but also all bottom-borders that we use inside of the sections.
closes #1442

### DIFF
--- a/.changeset/giant-cougars-live.md
+++ b/.changeset/giant-cougars-live.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+---
+
+Fix issue with the clickable cards being cut off in certaion situations.
--- a/packages/gatsby-theme-docs/src/layouts/internals/page-content-inset.js
+++ b/packages/gatsby-theme-docs/src/layouts/internals/page-content-inset.js
@@ -22,9 +22,6 @@ import { designSystem, Markdown } from '@commercetools-docs/ui-kit';
 const PageContentInset = styled.div`
   max-width: ${(props) =>
     props.maxWidth || designSystem.dimensions.widths.pageContentWide};
-  section {
-    overflow: clip;
-  }
   section > * {
     max-width: ${(props) =>
       props.maxWidth || designSystem.dimensions.widths.pageContent};


timonrey:
The snapshot discovered another bug with the code examples for endpoints on the wide layout. This will be fixed separately. Here the issue:
https://github.com/commercetools/commercetools-docs-kit/issues/1468

nkuehn (changes requested):
Visually the result is correct but I do really wonder whether we're not running circles here.   @gabriele-ct  just added the clip overflow and I doubt that he did not have a reason.  

And even if you end up changing it, please update the comment explaining in length why it was initially added at all.  Good opportunity to check whether it really has no side effects.

timonrey:
Probably yes, but it doesn't really fix the problem of the code sample not fitting in the box. I'm also wondering if there was another reason why this was added to the code. Anyway, we will need to fix that issue separately, because the problem lays probably in the left column of this specific example.

gabriele-ct:
Probably the issue was introduced here: https://github.com/commercetools/commercetools-docs-kit/pull/1453
the `overflow: clip` change  is related with the sticky code functionality.
Please @timonrey have a look to the comment here: https://github.com/commercetools/commercetools-docs-kit/pull/1453/files#r1040073803

If you make any changes to that overflow property, make sure the sticky functionality still works.

timonrey:
@gabriele-ct and I discussed the `overflow` change and came to the conclusion that the property is not needed here at all. Before we changed it to `overflow: clip`, it was set to `auto` but It only had some bad side effects on the bottom borders of every component on the page. It's also not needed for the new sticky example feature that was introduced recently.
So the removal of that property is fine from our side.

nkuehn (approved):

PR #1470, opened 2023-01-02
refactor: use gtag.js and support multiple IDs

### LINKED ISSUE #1469: Switch to Gtag based analytics plugin to support GA4
July 1st is the cutoff date for Google Universal Analytics to be deprecated in favor of GA 4. 

There is discussion going on about using a completely different tool but to decouple and be prepared we should nevertheless switch the setup so that it supports GA 4, which is not a hard task to achieve. 

1. use [gatsby-plugin-google-gtag](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-google-gtag) instead of the currently used plugin.  This also involves changing the external link component in our "magic" link to <OutboundLink>
2. ensure the IP is anonymized
3. allow to pass an array of strings to the configuration that holds the property ID (with graceful fallback to the existing single string).

nkuehn:
Resolves #1469 

includes two small changes of related scope: pass identical config to hubspot tracker and do not block passing the referrer to external links so our own other sites can see what traffic comes from the docs.

### DIFF
--- a/.changeset/proud-adults-think.md
+++ b/.changeset/proud-adults-think.md
@@ -1,0 +1,9 @@
+---
+'@commercetools-docs/gatsby-theme-docs': minor
+---
+
+Allow multiple GA tracking IDs by using gtag.js
+
+To migrate, rename the `gaTrackingId` theme configuration to `gaTrackingIds` and pass an array of strings.
+The previous configuration is still supported for compatibility.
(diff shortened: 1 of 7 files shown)


nkuehn:
FYI the visual regressions were all related to the year change in the footer

FFawzy (approved):
LGTM 👍

PR #1485, opened 2023-01-19
Render all images within the content with proper spacing

### LINKED ISSUE #1460: Apply the same padding and spacing to SVG images as to pixel images processed by gatsby
SVG images can be embedded in markdown but do not go through the processing pipeline and do not receive the automatic picture element wrapper with spacing etc. 

To not have to hardwire the spacing and padding into all individual diagrams the learning team would like SVGs (and/Or any other images that are not inside the gatsby image wrapper) to have at least the 8px padding left/righ and 16px top/bottom margin.

timonrey:
closes #1460

### DIFF
--- a/.changeset/eleven-eggs-carry.md
+++ b/.changeset/eleven-eggs-carry.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/ui-kit': patch
+---
+
+All images within the content are now rendered with the correct spacing at the edges.
--- a/packages/ui-kit/src/components/markdown.tsx
+++ b/packages/ui-kit/src/components/markdown.tsx
@@ -270,6 +270,11 @@ const TypographyPage = styled.div`
   font-weight: ${typography.fontWeights.regular};
   line-height: 1.5;
 
+  section > p > img {
+    margin: ${dimensions.spacings.m} 0;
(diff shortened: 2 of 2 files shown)


nkuehn:
> The issue says the a spacing of `margin: top/bottom 16px` and `padding: left/right 8px`should be applied.
> But the processed images that have the image wrapper use a padding of `8px` on all sides and a margin of `16px` for the top and `40px` for the bottom side.

whether margin or padding on CSS level is your implementation choice.  interpret my text as "spacing". 

The 40üx at the bottom are specific to certain content situations as far as I know, in some cases it's 16px only. I think it depends on what element comes after the image.  So I wanted a simple default to start with.  But if you can emulate the complete layout behavior you can do that too for sure.

nkuehn (approved):
It's solving 90% of the problem this way.  Choose for yourself if you want to emulate the complete spacing behavior or not.

nkuehn:
@timonrey centering is correct, but @lukpan  and me had agreed that we want to replace the grey box around it with white background, so it was intentional that there is an 8px white spacing left and right if the image is wide.   Ideally add a test case SVG that is full width or wider to see the behavior in action

PR #1487, opened 2023-01-24
fix(api type): remove un-necessary side by side component when render…

gabriele-ct:
…ing a single child


nkuehn (approved):
Thank you and sorry for the premature review on the tests. I was too excited that you found the cause.

gabriele-ct:
> Thank you and sorry for the premature review on the tests. I was too excited that you found the cause.

My bad, I didn't notice the PR wasn't in draft mode 👍

PR #1488, opened 2023-01-24
(to ease the discussion): drop new MC "inter" font instead of Roboto

nkuehn:
- just here to ease the discussion and evaluation of a UX pro
- dropped in inter without any further adjustments anywhere.  Letter and Line spacing and the weights feel a bit imbalanced yet but fortunately inter has very good documentation and tooling to optimize that (and is a variable weight/width font so it can actually be adjusted in detail as we like) https://rsms.me/inter/dynmetrics/ 
- there is no monospace / code variant of Inter so Roboto Mono remains there
- I think it's much better than the experiment with the TT commons brand font but could get some tuning of weights / sizes to be used.


timonrey:
@nkuehn I created a fresh playground PR for the new font (with the latest version) here: https://github.com/commercetools/commercetools-docs-kit/pull/1727
I will use it to test the suggested changes from @zbalek so that we have a preview of the new font with the design changes.

nkuehn:
superseded by #1727

PR #1495, opened 2023-01-31
Ga fix type api filter

gabriele-ct:


packages/gatsby-theme-api-docs/src/hooks/use-type-locations.js
```diff
@@ -12,6 +12,9 @@ const convertComponentInMdxToTypeLocations = (data) => {
   if (!locationsAreIndexed) {
     data.allContentPage.nodes.forEach((node) => {
       node.shortcodeOccurrence.forEach((occurrence) => {
+        if (occurrence.component !== 'ApiType') {
```
gabriele-ct, line 15: I admit, I could not find a better way to express this filtering in the graphql syntax. There might be a way, but I can't find it.
Essentitally this filter:
```
allContentPage(
          filter: {
            shortcodeOccurrence: { elemMatch: { component: { eq: "ApiType" } } }
          }
        )
```
Is matching all the pages that have at least one ApiType component, but not returning only the ApiType components...

nkuehn: That looks like the filter is on the level of the page itself but not on the type elements but you cannot put a filter at the level of the query that is the nested components.   I think that is the expected behavior if the components occurrences are not graphQL nodes by themselves which are referenced by other nodes.  

If that remains the case I suggest to merge the two hooks so that you only have one static query pulling all the occurrences.   If you have two hooks that pull the same static query but then filter out only the one or the other part of the contents you bundle the complete occurrence data into the site javascript twice.

gabriele-ct: It's not the same static query, indeed the graphql filter is acting as "pre-filter" by excluding all the pages containing only  components that don't match the specified query. So, it's more of a 2 step filtering, one at gql level, one at js. It can probably be optimized but I'd like to try first if it's solving the issue we're seeing on MDX v2 upgrade

nkuehn: OK, please create a tech debt ticket to refactor this to not have the complete data twice in the bundles.

nkuehn (approved):

timonrey (approved):

PR #1499, opened 2023-02-02
fix(gatsby-node): create pages only for content and release notes

gabriele-ct:


nkuehn (approved):
I pushed better commenting of the filesystem plugin configuration.  "internalContent" is a bit misleading, it's theme-provided default content.

Thanks!

gabriele-ct:
@nkuehn The VRT diff should be the usual timing issue right? Ok to be merged?

PR #1502, opened 2023-02-03
Headings id and toc url slugs sanitization

gabriele-ct:


packages/gatsby-theme-docs/src/plugins/rehype-id-slug.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * TODO: ehnancement: add the possibility of passing in a 'options.postProcess' function to be able to sanitize
+ * the slugs with a custom function implementation
+ */
+module.exports = () => (tree) => {
+  slugs.reset();
+  visit(tree, 'element', (node) => {
+    if (headingRank(node) && node.properties && !hasProperty(node, 'id')) {
+      const nodeSlug = slugs.slug(toString(node));
+      node.properties.id = nodeSlug.replace(/[-_]+$/, '');
```
nkuehn, line 23: The intent is not to remove them completely, the problem we face in various occasions is that there are trailing or leading slashes / underscores.   the root-rootcause is that the slugger does have issues with MDX components in the content. 

Didn't you work with a function that gets the pure text content from a piece of MdX in some other place? 

Ideally this implementation is re-using a function that is used in the matching implementation that generates the ToC

packages/gatsby-theme-docs/utils/generate-toc.js
```diff
@@ -19,7 +19,9 @@ function processToC(node, current, toString) {
   switch (node.type) {
     case `paragraph`: {
       current.title = toString(node);
-      current.url = `#${slugger.slug(current.title)}`;
+      const nodeSlug = slugger.slug(current.title);
+      // remove the trailing dash(es) or underscore(s)
+      current.url = `#${nodeSlug.replace(/[-_]+$/, '')}`;
```
nkuehn, line 24: please re-use the logic of how this is exactly happening in a utility.  The root cause of our problems is that the implementations of generating the ID and the toc are not matching

packages/gatsby-theme-docs/src/plugins/rehype-id-slug.js
gabriele-ct: The AST node structure is something like this:
```
node {
  type: 'element',
  tagName: 'h1',
  properties: {},
  children: [
    {
      type: 'text',
      value: 'Get Customer in Store ',
      position: [Object]
    },
    {
      type: 'mdxJsxTextElement',
      name: 'Beta',
      attributes: [],
      children: [],
      position: [Object],
      data: [Object]
    }
  ],
  position: {
    start: { line: 7, column: 1, offset: 29 },
    end: { line: 7, column: 33, offset: 61 }
  }
}
```
and the `toString` implementation is basically concatenating all the text nodes in a single string that gets passed to the `slug` function.
As you can see, the empty space is in the text node, and it gets converted to `-`. So this is how it works, I'm  not sure what you mean with `the slugger does have issues with MDX components in the content`. The slugger doesn't even see the mdx components.

nkuehn: Yes, you're right.  I was still mentally on the MDX v1 implementation which did not have as proper an AST.

packages/gatsby-theme-docs/src/plugins/rehype-id-slug.js
```diff
@@ -0,0 +1,26 @@
...
+ * rest of the implementation works exactly like the original.
+ */
+module.exports =
+  (options = {}) =>
+  (tree) => {
+    slugs.reset();
+    const postProcess = options.postProcess || ((input) => input);
+    visit(tree, 'element', (node) => {
+      if (headingRank(node) && node.properties && !hasProperty(node, 'id')) {
+        const nodeSlug = slugs.slug(toString(node));
```
nkuehn, line 22:
```suggestion
        const nodeSlug = slugs.slug(toString(node).trim());
```
trim could actually be enoug. 

I suggest to move the slugs.slug()  call into the postprocessing, too to really make sure the slugging is the same.  more like our own slug function

packages/gatsby-theme-docs/utils/slug-pre-process.js
```diff
@@ -0,0 +1,5 @@
+function preProcessSlug(text) {
```
gabriele-ct, line 1: I'm not particularly opinionated with naming. @nkuehn you tell me what you prefer here

PR #1503, opened 2023-02-03
React 18 (without any further updates)

nkuehn:
Started as an experiment but seems to at least build at all. So we may want to have a look into it.


gabriele-ct (approved):
🎉 let's get docs-kit to the age of majority!

PR #1509, opened 2023-02-07
chore: force shared components out of intermediate per-page bundles

gabriele-ct:


nkuehn (commented):
Please see my extra commit after the one I sent via slack.  I found a much simpler and more effective way to achieve the same and already included the separate plugins there.

nkuehn (approved):
let's hope this at least gets us to a state that builds reliably so we can refocus on the dev server

PR #1523, opened 2023-02-13
fix: reset filter in release notes not working

gabriele-ct:


gabriele-ct:
@nkuehn this fixes the issue but I'm still not 100% happy about how it works, I think we'll need to revisit it at a later stage

PR #1524, opened 2023-02-14
Load Mermaid JS from CDN to reduce build load

nkuehn:
@emmenko  had proposed this already in the original implementation PR, at the time I had preferred having the mermaid version and exact code controlled through NPM and renovate like all other dependencies.  But I did underestimate the impact that such large dependencies can have on the overall gatsby build and develop server through memory pressure and repetitive bundling and compression. 

So here's an approach to load the mermaid javscript client side from the jsdelivr CDN.  It's using a script loader hook someone else wrote (linked in the code).   I have the feeling there might be a simpler way  (e.g. ESM importing the library from the client side?)  but at least this one changes as little as possible and works.  

There may be inefficiencies that lead to flicker, a review would be appreciated. 

The graph rendering changes very slightly but I think it's related to the mermaid version change (VRTs show a difference in how the elements are exactly aligned). In any case we have no control of the exact layout anyways and mermaid versions will change it too over time. 

TODO: 
 - Try to get the loading of the external script to suspend the client side component loading so the suspense is only resolving when the lib is loaded and not just when our client side component is loaded.    This may resolve the testing issue and would be cleaner too concerning the fallback UI.  
 - some unrelated e2e test related to pagination fails which is weird

If this is considered good we could consider doing the same with video.js  which is a similarly huge chunk that is used as a whole in client side component, too.  Together they would reduce the amount of code being processed that is not per-page by over 50%.

### DIFF
--- a/packages/ui-kit/src/components/mermaid-client-side.tsx
+++ b/packages/ui-kit/src/components/mermaid-client-side.tsx
@@ -3,9 +3,10 @@ import PropTypes from 'prop-types';
 import { designTokens } from '@commercetools-uikit/design-system';
 import styled from '@emotion/styled';
 import murmurhash from 'murmurhash';
-import { colors, typography, tokens, dimensions } from '../design-system';
+import { colors, typography } from '../design-system';
 import { cssVarToValue } from '../utils/css-variables';
 import useScript from '../hooks/use-script';
+import LoadingSpinner from '@commercetools-uikit/loading-spinner';
 
 // This is a client-side only component.
 // It loads the mermaid library externally from a CDN to prevent the big mermaid codbase
@@ -99,18 +100,10 @@ const config = {
 `,
 } as const;
 
-const Figure = styled.figure`
-  background-color: ${colors.light.surfaceSecondary1};
-  border-radius: ${tokens.borderRadiusForImageFrame};
-  margin: 0;
-  padding: ${dimensions.spacings.xs};
-  display: flex;
+const Wrapper = styled.div`
+  width: 100%;
+  display: inherit;
   justify-content: center;
-  line-height: normal;
-  a span.nodeLabel {
-    color: ${colors.light.link} !important;
-    text-decoration: underline;
-  }
 `;
 
 const idForGraph = (graph: string) => `mermaid-${murmurhash.v3(graph)}`;
@@ -136,10 +129,16 @@ const Mermaid = ({ graph }: MermaidProps) => {
   }, [graph, mermaidLoadStatus]);
 
   return (
-    <Figure
-      data-test-id="mermaid-diagram"
-      dangerouslySetInnerHTML={{ __html: svg }}
-    ></Figure>
+    <>
+      {mermaidLoadStatus == 'ready' ? (
+        <Wrapper
+          data-test-id="mermaid-diagram"
+          dangerouslySetInnerHTML={{ __html: svg }}
+        />
+      ) : (
+        <LoadingSpinner scale="l" maxDelayDuration={0} />
+      )}
+    </>
   );
 };
 Mermaid.propTypes = {
--- a/packages/ui-kit/src/components/mermaid-client-side.tsx
+++ b/packages/ui-kit/src/components/mermaid-client-side.tsx
@@ -1,13 +1,17 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-import mermaid from 'mermaid';
 import { designTokens } from '@commercetools-uikit/design-system';
 import styled from '@emotion/styled';
 import murmurhash from 'murmurhash';
 import { colors, typography, tokens, dimensions } from '../design-system';
 import { cssVarToValue } from '../utils/css-variables';
+import useScript from '../hooks/use-script';
 
 // This is a client-side only component.
+// It loads the mermaid library externally from a CDN to prevent the big mermaid codbase
+// from causing build performance issues although effectively just being passed through
+// to client side processing anyways.
+const mermaidVersion = '9.3';
 
 // styling happens through a mix of the generic "themeVariables", diagram
 // type specific settings and direct CSS classes for diagram types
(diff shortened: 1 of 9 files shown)


packages/ui-kit/src/components/mermaid-client-side.tsx
```diff
@@ -111,20 +115,25 @@ const Figure = styled.figure`
...
+  );
   useEffect(() => {
-    mermaid.mermaidAPI.render(idForGraph(graph), graph, (svg: string) => {
-      setSvg(svg);
-    });
-  }, [graph]);
+    if (mermaidLoadStatus === 'ready') {
+      // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+      const mermaid = (window as any).mermaid;
+      mermaid.initialize(config);
```
nkuehn, line 131: maybe there is optimization potential in trying to avoid re-initialization of mermaid in every render?

gabriele-ct:
@nkuehn I've given a try to the branch.
- I've a figured out something that was wrong in the e2e testing and it was not waiting for the mermaid graph to be loaded before taking the screenshot.
- I've attempted an alternative implementation of lazy loading to achieve what you mentioned above (show the loading spinner while the library is being downloaded). It's far less robust of the `useScript` hook you've found, but if you like it we can merge the hook code into it

packages/ui-kit/src/components/mermaid-lazy.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const Mermaid = lazy(() => {
+  return new Promise((resolve) => {
+    const script = document.createElement('script');
+    script.src = `https://cdn.jsdelivr.net/npm/mermaid@${mermaidVersion}/dist/mermaid.min.js`;
+    script.onload = resolve;
+    document.body.appendChild(script);
+  }).then(() => {
+    return import('./mermaid-client-side');
+  });
```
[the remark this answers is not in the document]
nkuehn: @emmenko  I did not use gatsbys `<Script>` because it would load that big chunk for everybody on the first page hit, not only when they hit one of the very few pages that have such a diagram.  I was looking for something that only loads it when needed like the current implementation does.

packages/ui-kit/src/components/mermaid-lazy.tsx
```diff
@@ -0,0 +1,16 @@
+import { lazy } from 'react';
+
+const mermaidVersion = '9.3';
+
+const Mermaid = lazy(() => {
+  return new Promise((resolve) => {
+    const script = document.createElement('script');
```
nkuehn, line 7: @gabriele-ct  that's similar to the idea I folllowed, too but my implementation is catering for not adding the same script multiple times and the other cases the useScript hook covered.   At least I think it's the same approach (it's practically nesting two Suspense's to make sure that the loading of the external script is not happening server-side.  One Suspense loads the client side component,  the next is loading the external script client-side. 

Here's the commit:   701421daebfa7a2348805e53543dcd346fcf35cd 

@emmenko too wild?  I think it's not using undocumented React APIs, the lazy(load) API is specified in the official React type declarations and the documentation is also explaining it in general as somthing that returns a promise that resolves to a component export.

nkuehn:
@gabriele-ct  concerning the build / memory pressure here's my peaks in dev mode with 150 extra pages in the site:
 - my last version of this change:  ext peaks at  12 GB,  rss at 2.7 GB
 - main: ext peaks at 12 GB, rss at 3 GB
 
 It's still horribly slow even on my M1 mac and the memory difference is at best an indirect effect.  
 
Overally I would nevertheless vote for doing this  if considered viable and robust enough to have less noise in the analysis, both mentally as well as in the bundle size and memory analysis tools. Also on the video component.
 
Which exact approach we use is up to you and nicola.

packages/ui-kit/src/components/mermaid-lazy.tsx
gabriele-ct: @nkuehn I agree, in your implementation the whole script tag generation is much more robust, I just felt something was not right conceptually in your version (the component part) as you're lazy loading a component `mermaid-client-side.tsx` that in fact is rendering immediately after downloading JUST the dependencies declared in the import statements (not waiting for the mermaid library), and it's not showning the loading spinner while downloading the mermaid library from the CDN. 
I created a video to show a sort of "slow motion" of the 2 versions
Your version:
https://drive.google.com/file/d/19X7Gy5EpDCitXvjl38GitDdkaZKFzhsW/view?usp=sharing
My version:
https://drive.google.com/file/d/1y9OGQTLaknYyiF77JIgU4GVyGQ7h5YBA/view?usp=sharing

We can discuss this tomorrow, in any case my implementation is just to try it out, it needs refactoring and better handling of script tag injection

nkuehn: Yes, there is something not entirely clear to me about the way the nested Suspense works in my version. 

Needs slow-mo testing on a full build though (`yarn build && yarn serve`), the dev server you recorded has a very different bundling behavior.

packages/ui-kit/src/components/mermaid-lazy.tsx
```diff
@@ -0,0 +1,16 @@
+import { lazy } from 'react';
+
+const mermaidVersion = '9.3';
+
+const Mermaid = lazy(() => {
```
emmenko, line 5: Shouldn't we simply load the `mermaid-client-side` module here?

```tsx
const Mermaid = lazy(() => import('./mermaid-client-side'));
```

Then in the `mermaid-client-side` component you use the `useScript` hook.

emmenko: Example usage: https://javascript.plainenglish.io/how-to-load-a-dynamic-script-in-react-2940d30998dd

nkuehn: We could replicate the same loading indicator in the component through checking the loading state vs. through suspense.  

I had issues getting the tests to properly wait until the script is loaded and the component rendered when doing it through useScript and condiditional rendering, so I tried this other approach that works through suspense.  But it may well be that it can be solved in the test code, too.

gabriele-ct: I reverted to your initial implementation.

nkuehn:
@gabriele-ct  thanks.  I have pushed one more commit that cleans up the loading behavior. I validated it slow motion.

 - the Figure wrapper is now in the server side wrapper to have the background already be present while hydrating.
 - the server side wrapper renders the loading spinner if it's not client side (it does not actually shop up yet, requires a small fix in the UI-kit: https://github.com/commercetools/ui-kit/issues/2425

Good to go for me!

gabriele-ct (approved):
nice job

PR #1530, opened 2023-02-17
Webpack experimental lazy compilation in low_mem mode

nkuehn:
This brings the development server down to very low memory and startup time but at the cost of dev server robustness. 
That's why it's hidden behind our LOW_MEM flag which has other negative effects too (no information about where an error is, uncached builds)


gabriele-ct (approved):
nice

PR #1504, opened 2023-02-03
Gatsby 5

nkuehn:
Upgrade to Gatsby 5 plus various related dependency updates and changes to get the linting process nearly warning-free to better see the actual issues. 

> **This contains a lot of general cleanup and bugfixes that only surfaced as part of the upgrades of gatsby and nodejs. We could pick only those changes and merge them independently**

An experiment with using the new Slices feature was not successful, not being able to pass a function to a slice-d component is a constraint and the slice components have problems with theming as it seems.    I suggest to investigate that separately and now upgrade gatsby at all to be on a supported version concerning our memory performance issues. 

The build memory issues are unchanged though (docs-smoke test peaks at ca. 3.8-4 GB RAM in the single threaded "Building HTML renderer" phase.)

Known issues to be looked at:
 - images broken (maybe we have to re-add the relative image path plugin or find out why / what the replacement for that deprecated plugin is)
 - Running the VRTs on CI takes many many more snapshots, need to research what has to be changed in the command to filter for the intended ones only.  Maybe a result of one of the only indirectly required dependency updates or differing path / filename structures? 
 - local link check fails (times out, claiming the local server is not available).  Maybe first roll back the dependency updates related to the local link check.

PR #1531, opened 2023-02-17
Prepare for Gatsby 5 (but don't do it): node 18, dependency updates and fixes

nkuehn:
This PR is a subset of #1504  which includes everything that was addressed there to make the kit run on the latest dependencies and node 18, but not the actual Gatsby 5 update any more. 

It also refactors the build workflow to match the new one we use elsewhere (old one did not really work on node  18)

Merging this early allows to make a later upgrade to v5 simpler and does not force us to maintain the gatsby 5 branch if we continue to be blocked by the MDX v1 plugin working better than the v2 plugin. 

For explanations, see the comments in my self-review in #1504


gabriele-ct:
I've merged current `main` branch into this and solved some conflicts. In particular I'd like to underline that before that merge, the `nk-gatsby-5-preparation` branch was using MDX v2 plugin for gatsby and some mdx v2 remark plugins (remark-gfm).
I had to downgrade those dependencies to be compatible to MDX v1 which is the version we currently offer on the `main` branch

gabriele-ct (approved):

PR #1535, opened 2023-02-20
Revert to mdx v1

gabriele-ct:


timonrey (approved):
Thank you for the effort!
LGTM

PR #1545, opened 2023-03-09
feat: upgrade to Gatsby v5 (5.6.1)

gabriele-ct:


packages/gatsby-theme-docs/package.json
```diff
@@ -41,29 +41,30 @@
...
+    "gatsby-plugin-feed": "5.6.0",
+    "gatsby-plugin-google-gtag": "5.6.0",
     "gatsby-plugin-hubspot": "2.0.0",
-    "gatsby-plugin-image": "2.25.0",
-    "gatsby-plugin-manifest": "4.25.0",
-    "gatsby-plugin-mdx": "3.20.0",
-    "gatsby-plugin-sharp": "4.25.0",
+    "gatsby-plugin-image": "3.6.0",
+    "gatsby-plugin-manifest": "5.6.0",
+    "gatsby-plugin-mdx": "patch:gatsby-plugin-mdx@3.20.0#./patches/gatsby-plugin-mdx.patch",
```
nkuehn, line 51: Oh, this is cool. It means we don't even have to touch the repos that use the kit :-).   I had only read the docs which were not very specific about how the patch is held and maintained.

websites/documentation/src/content/configuration/plugins.mdx
```diff
@@ -5,8 +5,6 @@ beta: false
 
 ## Gatsby Transformer MDX Introspection
 
-<Warning>This Functionality has been deprecated. MDX v2 provides a rich AST that allows to solve the problems in a cleaner way</Warning>
```
nkuehn, line 8: I would not make a release of this plugin again just because of that readme change.

gabriele-ct: I agree

packages/gatsby-theme-docs/package.json
gabriele-ct: It's a cool feature indeed 👍

nkuehn (commented):
Nice and small :-)   

in the meanwhile gatsby 5.7 was released coincidentally.  I did the update and pushed it so we don't have to run another cycle.  it has a performance improvement related to the head component but I think we had tested that already and it was not addressing our problem.

nkuehn (approved):

PR #1546, opened 2023-03-09
Migrate gatsby-theme-docs to ESM gatsby-node and gatsby-config

nkuehn:
Since Gatsby 5 is merged now I remembered that since gatsby 5.3 ESM support is stable and that this could allow us to get rid of those pesky locked old versions of plugins in the remark / rehype / MDX ecosystem which is important to us. 

Current state: Shows that it works and is not more than a stupid chore now. 

Still to do here:
- for consistency the other themes and plugins should also have their gatsby-node and gatsby-config as ESM
- upgrading the various remark and rehype plugins that we locked and removing the locking instructions from the renovate config.

Deliberately not done:
 - using typescript


gabriele-ct:
@nkuehn I migrated all the `gatsby-config.js` and `gatsby-node.js` of the packages, including the websites. I got to the point that I get a green build but I still don't understand if we should be worried about this warning:
https://github.com/commercetools/commercetools-docs-kit/actions/runs/4414325422/jobs/7736263306#step:11:167

I feel it's something that assumes that the filename should be `gatsby-node.js` instead of `gatsby-node.mjs`

What's your opinion?

nkuehn:
> if we should be worried about this warning: https://github.com/commercetools/commercetools-docs-kit/actions/runs/4414325422/jobs/7736263306#step:11:167

I found one discussion with someone having a similar issue (part of the main ESM discussion thread at gatsbyjs) https://github.com/gatsbyjs/gatsby/discussions/37069#discussioncomment-4371660 

> I feel it's something that assumes that the filename should be `gatsby-node.js` instead of `gatsby-node.mjs`

Literally read, yes.  It's also cache-related - did you try deleting the gatsby cache file for this branch before re-running?  This answer in the mentioned thread also recommends that: https://github.com/gatsbyjs/gatsby/discussions/37069#discussioncomment-5022158

gabriele-ct:
> 
> Literally read, yes. It's also cache-related - did you try deleting the gatsby cache file for this branch before re-running? 

@nkuehn  I tried locally by running a yarn clean before building, and I still get the warning, so I don't think cleaning the GH action cache would help. Anyhow, I tried to filter by branch but I don't see any cache for `nk-esm-modules`

FFawzy:
@gabriele-ct  the cache key is `Linux-yarn-v3-98aa58be25c2c3e60743c24e7f9208b33700815362278cbc6d1f00b7088dea11` not the branch key ..  you can see it [here](https://github.com/commercetools/commercetools-docs-kit/actions/runs/4414325422/jobs/7736263306#step:7:20)

nkuehn:
@FFawzy  the github action cache holds caches per branch separately even if the key is identical.  So you need to filter for the branch first and then look up the cache by key

FFawzy:
weird .. in this case Gabriele is correct .. if you filter in the cache tab by branch first its empty .. even tho the logs clearly specified a cache key that has been used

gabriele-ct:
In case you were wondering, locally the cache gets generated, at least there are some files in the `.cache` directory, so it's not that it's not generating any cache at all

nkuehn:
Others have the issue, too: https://github.com/gatsbyjs/gatsby/discussions/37069#discussioncomment-5302178

I suggest to move forward and decide whether to open a formal issue with gastbyjs (i could not find it yet but it seems a common observation)

gabriele-ct:
Sounds good, ready to be reviewed then. Maybe we can consider releasing a pre-release version of the packages affected so we can try a real `docs` build and see if there are performance issues that become visible only with "beefy" websites

gabriele-ct:
Following what suggested in the thread on gatsby github, I tried to use the `resolveJSFilepath` helper like this:

```
const cacheConfig = {
      type: `filesystem`,
      name: stage,
      cacheLocation,
      buildDependencies: {
        config: [__filename, ...(await Promise.all(store.getState().flattenedPlugins.filter(plugin => plugin.nodeAPIs.includes(`onCreateWebpackConfig`)).map(async plugin => {
          var _plugin$resolvedCompi;
          return (_plugin$resolvedCompi = plugin.resolvedCompiledGatsbyNode) !== null && _plugin$resolvedCompi !== void 0 ? _plugin$resolvedCompi : await resolveJSFilepath({filePath: `${plugin.pluginFilepath}/gatsby-node`});
        })))]
      }
    };
    config.cache = cacheConfig;
```

This seems indeed to address the issue.
So it looks it's something could be potentially addressed in one of the future gatsby releases

nkuehn:
I'm not sure whether prereleases are worth the effort but I may be misled becuase a manual one-off prerelease is simpler than the documentation suggests ( https://github.com/changesets/changesets/blob/main/docs/prereleases.md ) 

Do you think we should directly include the remark / rehype version upgrades in this or do that separately?

gabriele-ct:
> I'm not sure whether prereleases are worth the effort but I may be misled becuase a manual one-off prerelease is simpler than the documentation suggests ( https://github.com/changesets/changesets/blob/main/docs/prereleases.md )

when I say pre-release I mean totally manual release, meaning from the feature branch run npm publish by manually specifiying the version e.g. `21.2.0.alpha.0`.
Then, on `docs` repo, on a separate branch (not to be merged), test the new alpha version of the libs. If all works good, we just run the proper `changeset` release publishing the final `21.2.0` version and then we update the `docs` repo with the final libs version (this time merging it into main).
Am I missing anything? 

> Do you think we should directly include the remark / rehype version upgrades in this or do that separately?
I'd keep it separate as I don't want to replicate the mdx2 patterm. Let's go step by step.

I can dedicate tomorrow morning to create the pre-release (with the help of someone with publish rights) versions, and test them on `docs` repo. If all looks good, by the end of tomorrow we could have this branch merged into main and ready for a final release

gabriele-ct:
@nkuehn I've released a pre-release version of the modified packages and upgraded `docs` repository. From what I see, the warning message about caching is not creating any performance issue during build or in development mode.
Please go ahead with review/approve 👍

nkuehn:
@gabriele-ct  nice.  the pr was originally opened by me so you have to approve to merge

gabriele-ct:
> @gabriele-ct nice. the pr was originally opened by me so you have to approve to merge

🤦 deja vu

PR #1548, opened 2023-03-10
Allow newer changeset GH action versions

nkuehn:
following their documentation, this unlocks minor
versions of the changesets action.
The 1.2 we hardwired uses the deprecated way
to set outputs and would at some point fail


timonrey (approved):
Thanks

PR #1562, opened 2023-03-17
chore: upgrade locked dependencies to their latest esm version

gabriele-ct:


jest.test.config.js
```diff
@@ -1,3 +1,29 @@
...
+  'space-separated-tokens',
+  'trim-lines',
+  'unified',
+  'unist-util-filter',
+  'unist-util-generated',
+  'unist-util-is',
+  'unist-util-position',
+  'unist-util-stringify-position',
+  'unist-util-visit',
+  'vfile-message',
```
nkuehn, line 24: It's unfortunate that we have to maintain the list of ESM only modules in a different place because of jest but it's still better than to not be able to upgrade them at all.

packages/gatsby-theme-docs/utils/process-table-of-content-fields.js
nkuehn, line 1: did we keep this because we would need it again for MDX v2 or was it just a leftover.

nkuehn (approved):
@gabriele-ct  LGTM as a pattern, I pre-approve you this because I'm out in the afternoon and fady, too

packages/gatsby-theme-docs/utils/process-table-of-content-fields.js
gabriele-ct: Good question. It might be because of mdx v2, but to be honest it doesn't matter that much as we have the mdx v2 branch to be used as reference.

jest.test.config.js
gabriele-ct: There might be a better and more elegant solution, i didn't dig too much into jest. This solution seems to be working for now.

PR #1570, opened 2023-03-20
fix: use only subdomain for kit website

nkuehn:
This makes it consistent with all other URLs we use and the sentry and other
tooling configuration work by using only  *.commercetools.vercel.app


gabriele-ct (approved):
Thank you 👍

PR #1576, opened 2023-03-20
fix: explicitly include api-specs/test package

gabriele-ct:
Seems like the base configuration of renovate excludes directories named test:

https://github.com/renovatebot/renovate/blob/e5ccf2b85e290e45b20bbf36d5a529a5261d0c62/lib/config/presets/internal/default.ts#L42-L55

I try fixing it adding explicitly the path api-specs/test/** to the includePaths rule


nkuehn (approved):
let's try.  
And ideally give them feedback that it's a pretty wild assumption that  can fire back.

PR #1581, opened 2023-03-21
chore: dependency updates and fix renovate syntax

gabriele-ct:


nkuehn (approved):
Thank you for finding the bug in the renovate config! 

The VRTs were related to a simpler way of calling curl CLI in the example code, all good and approved.

PR #1583, opened 2023-03-21
chore: fix renovate grouping

gabriele-ct:
Following what described in this [issue](https://github.com/renovatebot/config-help/issues/339), I'm trying to re-sort the priority of the rules and match package names instead or URLs


gabriele-ct:
@nkuehn another try

PR #1586, opened 2023-03-21
chore: remove resoultions from package.json

gabriele-ct:


nkuehn (approved):
nice.  Just the inevitable lockfile conflict remaining

PR #1592, opened 2023-03-24
Ga page awareness course status

gabriele-ct:


.github/workflows/main.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           path: |
             .cache
             public
-          key: ${{ runner.os }}-gatsby-monolithic-v1-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-gatsby-monolithic-v2-${{ hashFiles('**/yarn.lock') }}
```
nkuehn, line 70: why does this make the previous caches incompatible?

packages/gatsby-theme-docs/gatsby-node.mjs
```diff
@@ -348,6 +349,9 @@ export const onCreateNode = async (
         ? Number(node.frontmatter.timeToRead)
         : 0,
       tableOfContents: await generateToC(nodeBodyAst),
+      courseId: node.frontmatter.courseId
+        ? Number(node.frontmatter.courseId)
+        : 0, // TODO: understand how to make an Number field nullable
```
nkuehn, line 354: Answer: remove the exclamation mark in the schema.   https://graphql.org/learn/schema/#object-types-and-fields

packages/gatsby-theme-docs/src/hooks/use-course-pages.js
```diff
@@ -0,0 +1,41 @@
+import { useStaticQuery, graphql } from 'gatsby';
+
+export const useCoursePages = () => {
+  const coursePagesData = useStaticQuery(
+    graphql`
+      {
+        allContentPage(filter: { courseId: { ne: 0 } }) {
```
nkuehn, line 7: reminder to change the filter if you switch to using null for not set

websites/docs-smoke-test/src/content/self-learning/customer-management/customer-groups.mdx
nkuehn, line 1: let's not use actual meaningful content on the smoke test. The smoke test is public content on the web which woul d be very  misleading. 

Please make up a minimal scenario or ideally re-use the existing smoke test pages that we have for quizzes (you can add a second page and move them into a chapter in the nav).    We need to keep an eye on the percy load (it's not cheap) and I think tests should be very targeted and describe a specific scenario in the content)

nkuehn (commented):
just comments.  curious to see how the next step (live status from the courses) turns out. 

having the full course here is OK as long as it's in development.  Speak to @FFawzy  next week about how to improve the smoke test course to cover the needs here.

packages/gatsby-theme-learning/src/hooks/use-course-status.ts
```diff
@@ -0,0 +1,59 @@
...
+  useEffect(() => {
+    const fetchToken = async () => {
+      const token = await getAuthToken();
+      setAuthToken(token);
+    };
+    fetchToken();
+  });
+
+  const { data, error, isLoading } = useSWR(
+    isAuthenticated && authToken ? apiEndpoint : null,
```
gabriele-ct: As you can see here we're using conditional fetching. Basically it will fetch the courses only when authenticated and with a valid authToken (probably we just need to check the authToken). The issue here is that all the renderings that happen while the authToken has been fetched, swr does not return the stale data but instead:
```
{
data: undefined
isLoading: false
error: undefined
```
This is a probelm because it causes a very quick flash in the rendering each time we navigate through pages.
I spent quite some time trying to figure out how to overcome this issue without having to use the imperative API for fetching the data. Maybe I'm missing the big picture here, any ideas?

packages/gatsby-theme-learning/src/hooks/use-course-status.ts
```diff
@@ -0,0 +1,59 @@
+import useSWR from 'swr';
```
gabriele-ct, line 1: Note about SWR: by default `revalidateIfStale = true` meaning that even if swr returns the stale data straight away, it will perform a fetch in the background to refresh the state, in case it changed. So, don't get fooled by the network activity (like me), stale data is actually being returned

packages/gatsby-theme-learning/src/hooks/use-course-status.ts
```diff
@@ -0,0 +1,59 @@
...
+import { useContext, useEffect, useState } from 'react';
+import ConfigContext from '../components/config-context';
+import { useAuth0 } from '@auth0/auth0-react';
+import { useAuthToken } from './use-auth-token';
+import type { Course } from '../external-types';
```
gabriele-ct: @FFawzy we need to review the Course type. The type is fine but the API return it wrapped in `result.enrolledCourses` nodes

websites/docs-smoke-test/src/data/navigation.yaml
```diff
@@ -57,3 +55,9 @@
     - title: BetaPageWithLongNameThatOverflowsBecauseItsOneWord
       path: /empty-pages/empty-page-four
       beta: true
+- chapter-title: Self-learning course
```
gabriele-ct, line 58: Temporarily added to check how it behaves with courses where the user is not enrolled

packages/gatsby-theme-docs/src/layouts/internals/page-course-status.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
```
gabriele-ct, line 1: Don't focus on this, it's just a throw-away component until we have designs

packages/gatsby-theme-learning/src/hooks/use-course-status.ts
gabriele-ct: ahah, thank you I'll update the types

packages/gatsby-theme-learning/src/hooks/use-course-status.ts
gabriele-ct: Ignore this comment. After reading a bit more in depth the documentation, I refactored the code and now it seems to be working as expected

websites/docs-smoke-test/src/icons/svg/badge-placeholder-cm.svg
nkuehn, line 1: I think the images are not necessary any more.

packages/gatsby-theme-docs/src/layouts/internals/page-course-status.js
```diff
@@ -0,0 +1,44 @@
...
+import { useCourseStatusByCouseId } from '@commercetools-docs/gatsby-theme-learning';
+import PropTypes from 'prop-types';
+
+const CourseStatus = (props) => {
+  if (props.error) {
+    return <span>unavailable</span>;
```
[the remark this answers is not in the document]
zbalek: From the screenshots that Gabriele sent, I'd agree to go with the first approach and not show anything to a user.

_(Please note that if this was a critical component for a user to progress, I'd suggest showing that there's an error displayed so that users get feedback re what's happening.)_

FFawzy:
@gabriele-ct 
a quick comment on the swr implementation so far (I know you are still working on this) 
- consider abstracting your `fetcherWithToken` function because you will most probably need it with every upcoming self-learning feature so we don't duplicate it in every hook
- another thing your cacheKey for invalidating could be as simple as `/courses` and the rest could be handled by your `fetcherWithToken` (baseUrl .. etc)

Well done so far this is looking good ! 🚀

gabriele-ct:
> * consider abstracting your `fetcherWithToken` function because you will most probably need it with every upcoming self-learning feature so we don't duplicate it in every hook

Yes, that's planned. I'll move it to a util as soon as we implement the next API call.

> * another thing your cacheKey for invalidating could be as simple as `/courses` and the rest could be handled by your `fetcherWithToken` (baseUrl .. etc)
> 

True, that will move some complexity to the fetcher function.

Thank you! 👍

packages/gatsby-theme-docs/src/hooks/use-course-pages.js
```diff
@@ -0,0 +1,41 @@
...
+
+const coursePageMap = new Map();
+var isIndexed = false;
+/**
+ * Creates a map between page slugs and course ids
+ */
+const courseMapToPages = (coursePagesData) => {
+  if (!isIndexed) {
+    coursePagesData.forEach((element) => {
+      coursePageMap.set(element.slug, element.courseId);
```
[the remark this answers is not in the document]
gabriele-ct: I agree

packages/gatsby-theme-learning/src/hooks/use-course-status.ts
nkuehn: Yes, you just "see" it as the status not disappearing but staying the old status while the new state is fetched.

packages/gatsby-theme-docs/gatsby-node.mjs
```diff
@@ -481,10 +486,14 @@ async function createContentPages(
...
 
       default:
-        actions.createPage({
+        let contentPageData = {
           ...pageData,
           component: require.resolve('./src/templates/page-content.js'),
-        });
+        }
+        if (courseId) {
+          contentPageData = {...contentPageData, context: {...contentPageData.context, courseId}}
```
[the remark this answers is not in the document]
nkuehn: I just wanted to say that I do not understand the code because I forgot what "context" is in this context :-)

FFawzy:
the `getAccessTokenSilently` or the `getAccessToken` 
we are asking for the token everywhere .. can't we centralize this with a useEffect or any other idea, store the value in a context and update it when needed and then you can use it right away, saving you the hassle of requesting it every time 

one more thing, don't forget changesets

gabriele-ct:
> the `getAccessTokenSilently` or the `getAccessToken` we are asking for the token everywhere .. can't we centralize this with a useEffect or any other idea, store the value in a context and update it when needed and then you can use it right away, saving you the hassle of requesting it every time
> 
Thank you @FFawzy for bringing this up. I actually spent quite some time trying to simplify this, now with fresh eyes I found that it was easier that I thought, I refactored a bit the code.
To clarify: `getAccessToken` is just a wrapper around `getAccessTokenSilently` which is auth0 API. The wrapper simply avoid code duplication and calls the `getAccessTokenSilently` API. Although this is marked as async, most of the times the token is cached on auth0 side, so there's no real overhead calling multiple times this hook. I'd avoid caching auth token in the context as it might expire and we can find ourselves in strange corner cases. Calling the `getAccessTokenSilently` is the safest and most efficient way to get a valid authToken.

FFawzy (approved):
🚀 🎊

PR #1595, opened 2023-03-31
feat: display course topics status

gabriele-ct:
https://github.com/commercetools/learning-tech/issues/376


packages/gatsby-theme-learning/src/components/config-context.ts
```diff
@@ -1,19 +1,24 @@
 import { createContext } from 'react';
 
+export enum EFeatureFlag {
+  CourseStatus = 'status-indicator',
```
nkuehn, line 4: I wonder if a data structure that  has defaults might be better suited.  I can't really wrap my head around why exactly to be honest, it's just a feeling.

packages/gatsby-theme-learning/src/components/sidebar-topic-status.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+type PageTopicStatusProps = {
+  courseId: number;
+  pageTitle: string;
+};
+
+const SidebarTopicStatus = (props: PageTopicStatusProps) => {
+  const { isAuthenticated } = useAuth0();
+  const { data } = useFetchCourseDetails(props.courseId);
+  const { features } = useContext(ConfigContext);
```
nkuehn, line 32: It somehow feels wrong to me that it's necessary to pull the features from a hook to pass them in although the Config Context got them as props when initialized in gatsby-ssr   - but I am really not sure and have not fully read the code in detail. 

BTW tobias maintains this helper lib which is how we do it in the MC:  https://github.com/tdeekens/flopflip  (backed by a cloud service to be able to dynamically toggle per user).  Its solving problems of big apps, not really ours yet but has a lot of helpers.

FFawzy:
@zbalek 
that definitely should be the case that you don't get a tick if failed.
from my side I'm getting the correct behaviour 
Screenshot 2023-04-04 at 2 20 42 pm

if you want we can huddle or you can write here the steps to reproduce what you saw so I can check it

nkuehn (commented):
I only found the time to scan it, not fully read it.  So some of what caught my eyes might be wrong. 

Please add a PR description or link to the ticket so that the definition is known what is in scope and what not.

zbalek:
@FFawzy happy to approve then - For me, once I completed the quiz (incorrectly) the 'Overview' page got a green tick

nkuehn:
I tried the UI now, too. 

@gabriele-ct  nice to see the SWR approach is working - no flicker, the status in the nav is updating with a little delay but smoothly.    What's slightly irritating though is that when the page is reloaded (e.g. after navigating to a different docs page and back) the icon is the same as if I did not do anything yet while it loads.   Maybe we should use the localstorage based SWR cache and clear it on login / logout. 

@FFawzy I forgot what the logic behind the "overview" page with no content becoming green automatically after being auto-enrolled into the course when first taking the quiz on the second page.  If that is a content-only page it would be strange to mark it done but only after you finished the next page.   It could either have a neutral look in general because it has nothing to be completed or we provide the API and background listener that marks the activity completed on that page when you scroll down to the end or someething.

FFawzy:
@nkuehn yea I'd skip the overview page since it is just an intro (at least in the MVP state) .. maybe later when we introduce reading-trackers we can apply it to those pages 

WDYT ? @zbalek
removing the circle from the overview pages as there is nothing to be tracked there (yet)

zbalek:
@FFawzy I think I just got confused, I think it makes sense to mark it as completed once user goes to the next page from the overview page - so let's keep it. I just got confused cause once I completed the quiz, it was the overview page that got the tick instead of the test one. Apologies for the confusion.

nkuehn:
> I just got confused cause once I completed the quiz, it was the overview page that got the tick instead of the test one.  

That's effectively the problem I meant.  If you complete the quiz on the second page but don't pass it,  the first page is completed implicitly (we only check when you submit a quiz) but the second is not. 

Learning for me: we may need to have 
 - an "in progress" state visually for e.g. partially done (failed the quiz)
 - and / or a special visual state for pages (topics) that do not have any activity that can have a status

zbalek:
From my perspective, I wouldn't add additional states as even though they may be factually correct, user does not need to have e.g. 4 different states. Their main goal here would be to understand what's done and what's not done.
My proposal: 
- if a page is in progress (.e.g. failed quiz) = not completed = empty circle
- topics that don't have activity = once user visits the page = learning is completed = green tick

packages/gatsby-theme-learning/src/components/config-context.ts
gabriele-ct: not sure what you mean here. but the `feature flag` feature is working and I would descope this discussion from this PR. We can have a discussion about it and then create a separate ticket

packages/gatsby-theme-learning/src/components/sidebar-topic-status.tsx
gabriele-ct: The `isFeatureEnabled` utility function is trivial and very possibly not needed, indeed I could simply `features.includes('feature-name')` and save the hassle of passing features to another helper function.

websites/docs-smoke-test/src/content/self-learning/2-quiz.mdx
```diff
@@ -1,8 +1,9 @@
 ---
 title: Quiz component
-courseId: 55
+courseId: 66
+topicName: 'Quiz 2'
```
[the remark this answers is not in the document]
gabriele-ct: Yeah, you're right. I think I keep getting confused with numbers as string where indeed quotes are needed

gabriele-ct:
> @gabriele-ct SWR is a cache system in its main functionality, so I don't really understand your concern to be honest. At the moment it's a memory cache with pageload lifetime (which is across page navigations after the react app hydrated). The idea is to extend the lifetime across page reloads. invalidating has to be done anyways after logout, even in the current state.

An additional caching layer I mean...
Anyhow, if you feel is needed, let's specify it properly and implement it. Definitely out of this ticket scope

nkuehn:
> An additional caching layer I mean...

Excuse the confusion @gabriele-ct  - I did not mean an additional layer, I only thought about configuring SWRs cache backend to use localstorage instead of the default memory store.   https://swr.vercel.app/docs/advanced/cache#localstorage-based-persistent-cache

gabriele-ct:
> I did not mean an additional layer, I only thought about configuring SWRs cache backend to use localstorage instead of the default memory store.

ahhh, I thought implementing something on our side. I'll give it a go

gabriele-ct:
@nkuehn I created a separate branch for testing swr with localstorage
https://github.com/commercetools/commercetools-docs-kit/pull/1604

I'd like you to have a look, maybe try it and let me know what you think. My feeling after spending some time on it, is that it's not that trivial to setup (definitely not just a matter of adding a cache handler) and opens several corner cases that might be hard to figure out. Let me know what you think

packages/gatsby-theme-docs/src/hooks/use-course-pages.js
```diff
@@ -25,17 +26,33 @@ var isIndexed = false;
...
+    (prev, curr) => ({ ...prev, [curr]: coursePageMap.get(curr) }),
+    {}
+  );
+  // sanity check: all the pages should belong to the same course, therefore have the same courseId
+  const isOk = Object.values(courseInfo)?.every(
+    (info, _, theArray) => info?.courseId === theArray[0]?.courseId
+  );
+  if (!isOk) {
+    // TODO: decide if we want to make this blocker
+    console.warn(
```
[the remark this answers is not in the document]
gabriele-ct: A build time check has been implemented with an explanatory error message

packages/gatsby-theme-learning/src/components/sidebar-topic-status.tsx
nkuehn: A helper function is mandatory in my opinion,  down the line feature flagging is per user or group of users (e.g. commercetools users only etc).   It's a black box for the individual component why a feature is off or on now.

gabriele-ct: if you feel we should look into the `flipflop` lib, we can do it but I don't think it's worth doing it now. We could create an enhancement ticket to make feature flagging more powerful in the future

nkuehn: I think your simple function call is enough for what we need now.  I just referenced flipflop to you as an inspiration of how a mature set of helpter API for feature flags looks like .

FFawzy:
Zoe gave the OK on the following usecases

- loading state = (after refreshing the page, .. etc, AKA no cache in memory ) show no circles, we reserve the space for it from the beginning to avoid layout shifting and only show the circles once we have a response
- Stale while revalidating use case = showing the old value until its updated
- Overview being checked after first quiz = OK for now, until we implement the reading tracker

----
which means we can apply the localstorage as a cache provider as an enhancement after MVP.
along with the reading tracker


CC: @zbalek @nkuehn @gabriele-ct

nkuehn:
OK

> which means we can apply the localstorage as a cache provider as an enhancement after MVP.
along with the reading tracker
Can be separate, this sounds like they need to go together.

nkuehn (changes requested):
There's still a lot of layout shift and multiple times.  See video, shows two consecutive page reloads.  The vertical scrolling is a different topic, but there are two additional steps of layout shift 
- once when the page hydrates, the indentation depth changes from the statically pregenerated version  (I guess the indentation is in a code path that only evaluates in the hydrated app - but the information that the page is a learning page is in the frontmatter so it can be applied statically, too) 
- then the chapter name also shifts a bit when the data is loaded  (I guess that is just polish work - the chapter level status icon seems to be bigger than the space reserved. Maybe rather constrain the icon into the box that reserves the space than letting the icon take whatever space it needs? )


https://user-images.githubusercontent.com/4006139/232491312-1d41554f-1a11-4598-9826-96886979b517.mov

gabriele-ct:
> * once when the page hydrates, the indentation depth changes from the statically pregenerated version  (I guess the indentation is in a code path that only evaluates in the hydrated app - but the information that the page is a learning page is in the frontmatter so it can be applied statically, too)

This happens because the page has no knowledge of the logged state of the user and we agreed not reserve the icon space if the user is not logged in. Once the logged state becomes available we can reserve the space while the actual data gets loaded

> * then the chapter name also shifts a bit when the data is loaded  (I guess that is just polish work - the chapter level status icon seems to be bigger than the space reserved. Maybe rather constrain the icon into the box that reserves the space than letting the icon take whatever space it needs? )

This can be easily fixed, I'll have a look into it

nkuehn:
> This happens because the page has no knowledge of the logged state of the user and we agreed not reserve the icon space if the user is not logged in. Once the logged state becomes available we can reserve the space while the actual data gets loaded 

Sorry, this does not make sense to me.  We now have a state where we don't show any icon but blank space while the app does not know anything about the user state yet (but knows that the page is a course page).  The user does not care about which exact technical phase the loading is is to make a visual difference  - we reserve a little less space while the app is hydrating and then a little more when the API is called but in any case there is no icon visible.   

@zbalek  maybe you're more talented in verbalizing the problem I see in the video (or tell me that I'm wrong)?

gabriele-ct:
> Currently the app reserves a little less space while the app is hydrating and then a little more when the API is called but in any case there is no icon visible. The user does not care about which exact technical phase the loading is is to make a visual difference.
> 

@nkuehn I have no problem at all in implementing the initial spacing, I just implemented what was initially in the requirements . I understand the current time is a bit hectic but It would be beneficial in the future, define how the UI should look like in different app states to avoid all these roundtrips around an icon position 👍

zbalek:
@gabriele-ct I understand your concerns. In this case believe that the outcome that we agreed on (Fady's comment above) was that we reserve the space for the icons from the beginning to avoid layout shifting and only show them once we have a response. @nkuehn So what seems to be happening in the video is that it looks like the layout did just that but then once it loads it actually jumps one more time - which is what we wanted to avoid by reserving the space.

FFawzy (approved):
🎊

nkuehn (approved):
:heart:

PR #1596, opened 2023-04-03
Remove open-source from banned words

melindafekete:
When used as an adjective preceding a noun, open-source should be hyphenated. 

Sources:
- https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/o/open-source
- https://www.oxfordlearnersdictionaries.com/definition/english/open-source

I deleted the check entirely, instead of swapping them around since there can be cases when the non-hyphenated version should be used (see in Microsoft example above).


naveenchkrvrthy (approved):

emmenko:
@melindafekete Please remember to add a changeset if the changes should be released.

Once the changeset is merged, it will be included in the next release

PR #1604, opened 2023-04-05
Ga topic status with localstorage

gabriele-ct:
Attempt to use localstorage cache for swr.

I invested some time on it and still has glitches:
- if a user logs out with, say, 1 quiz passed, then logs in again and passes the 2nd quiz, when navigates from one page to another, keeps getting the old 1 quiz passed response.
- little flash when navigating to a SSR page (as it doesn't have localstorage knowledge)

### DIFF
--- a/packages/gatsby-theme-docs/src/layouts/internals/layout-application.js
+++ b/packages/gatsby-theme-docs/src/layouts/internals/layout-application.js
@@ -1,7 +1,9 @@
 import React from 'react';
+import { SWRConfig } from 'swr';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import { designSystem } from '@commercetools-docs/ui-kit';
+import useIsClientSide from '../../hooks/use-is-client-side';
 
 /* NOTE: `overflow` shorthand is only supported is Chrome and FF */
 const Root = styled.div`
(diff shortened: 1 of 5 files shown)


packages/gatsby-theme-docs/src/layouts/internals/layout-application.js
```diff
@@ -41,18 +43,55 @@ const Container = styled.div`
...
+  const { isClientSide } = useIsClientSide();
+  return (
+    <>
+      <Root
+        role="application"
+        id="application"
+        isGlobalNotificationVisible={Boolean(props.globalNotification)}
+      >
+        {isClientSide ? (
+          <SWRConfig value={{ provider: localStorageProvider }}>
```
gabriele-ct, line 84: First doubt. If we want to have the course/topic status available in the sidebar and (potentially in the future) in the actual content page, the only place common between these 2 layouts is the application layout.
Adding cache provider here means that all the usages of swr in the app will be cached in local cache. This might be what we actually want, or might be creating side effects.

packages/gatsby-theme-docs/src/layouts/internals/layout-application.js
```diff
@@ -41,18 +43,55 @@ const Container = styled.div`
...
-    </Root>
-    <div id="modal-portal" />
-  </>
-);
+const localStorageProvider = () => {
+  // When initializing, we restore the data from `localStorage` into a map.
+  const map = new Map(JSON.parse(localStorage.getItem('app-cache') || '[]'));
+
+  const handleBeforUnload = (event) => {
+    const appCache = JSON.stringify(Array.from(map.entries()));
```
gabriele-ct, line 51: This handler is very aggressive as it will dump swr state to the cache each time the page is refreshed. But we refresh the page alsow each time we login and logout.
This is the reason of the 2 custom events that basically disable writing the cache upon login and logout.

packages/gatsby-theme-docs/src/layouts/internals/layout-application.js
nkuehn: #1  I wonder if something like the SWRConfig would not be better placed in gatsby-browser's "wrapXXXElement" hooks.   That implicitly makes them client side only and also makes sure they are in all layouts. 

#2  having one cache provider for any SWR even across microsites felt weird to me too in the beginning but I concluded that it's the whole point of it.   The cache keys are the URLs you use to fetch from SWR and the content is the API response.  And we want that information to be available at least in stale state for other pages / views / microsites.   So it must be one fixed key like "app-cache" and also at top level for the domain.  We could use a more specific localstorage key maybe that is more explanatory.

nkuehn (commented):
The behavior is nice :-) 

Still a slight delay but that is the general hydration of the static page after reloading (it just becomes more visible but is not new, the top meny does not work too in the first second).

nkuehn:
> If you have time I'd like you to have a look to more specific scenarios like, for example:

@FFawzy  could you do me a favor and test through the mentioned scenarios with a diligent look at the status indicators in the navigation?  Realistically I would not be able to follow up because I'll only be there for a day next week.

nkuehn:
>  the tick disappears for a sec and then appears again 
@FFawzy  to properly analyze you have to have the network tab open and filter the api requests.  There is a short delay on reloads (like microsite switch) that cannot be avoided because it's the time the browser takes to hydrate the static page with the app.  But there could also be a longer delay if it really re-fetches the data.    Slow-motion / slow network mode of the dev tools is you friend in this kind of analysis work.    
The second topic you brought up also sounds like you need to analyze it in depth with slow mo and the API calls logged while it loads

PR #1611, opened 2023-04-11
feat: modal complete profile

gabriele-ct:


packages/gatsby-theme-learning/src/components/profile-complete-guard.tsx
```diff
@@ -0,0 +1,55 @@
...
+        if (data.status !== 200) {
+          throw new Error(`Error fetching user profile for user ${user?.sub}`);
+        }
+        const userData = (await data.json()) as User;
+        updateProfile(userData);
+      } catch (error) {
+        console.error(error);
+      }
+    }
+    if (isModalFeatureEnabeld && isAuthenticated) {
```
nkuehn, line 49: I'm not sure if I fully understood the code path here. This looks like it would call the API every time the app is loaded, even if firstname, lastname and company are already properly set as made available by the Auth0  login.   
@FFawzy  maybe I'm misled here and the Auth0 session doesn't actually provide these values. 

Maybe I'm sticking my finger into a more general thing here - the user profile with first name, last name etc is application state whose master information is on the server side and it's being shown in many places in the UI (e.g. avatar, the "hey welcome back Gabriele" thing we have on the board etc).   In other places we used SWR to centrally cache it and deduplicate the requests.   Here it may be a bit more fiddly as the avatar is actually pulling the initial info from the Auth0 token's custom claims - that's btw also maybe why the avatar does not update if I change my name in the form.  Could it make sense to use the Auth0 provided data as some kind of initialization of the local cache (like SWR) but from then on have all components rendering it use that cache only? 

Sounds like something we better discuss together (or I really missed something)

nkuehn (commented):
Just a comment for now, I have only gotten half through and have to drop off now. Plus, unsure when / whether I will find time for feedback tomorrow, so I better leave my first impression than not.

packages/gatsby-theme-docs/gatsby-ssr.js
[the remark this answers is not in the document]
nkuehn: PS: I must have submitted during a change, maybe obsolete

packages/gatsby-theme-learning/gatsby-ssr.js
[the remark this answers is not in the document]
nkuehn: PS: I must have submitted during a change, maybe obsolete

packages/gatsby-theme-learning/src/components/profile-complete-guard.tsx
gabriele-ct: > Sounds like something we better discuss together (or I really missed something)

Let's discuss this together. Some inputs:
- At login, the `user` object that auth0 returns is missing the `company` field, that's why I need to always fetch the user profile to have a complete view of what fields might be missing
- Avatar component is getting most of the information displayed from custom claims such as `https://docs.commercetools.com/display_name` and `https://docs.commercetools.com/global_account_name` it's not directly fetching data from the auth0 user profile. 
- I consider the user profile as data with very low update rate: the user will typically update those information only when logs in the first time and probably never again, or something close to it. For this reason I believe that storing this data in context once at login is a better approach rather than using SWR.

nkuehn: > 
@FFawzy  it shuold be trivial to add the company here: https://github.com/commercetools/learning-tech/blob/main/auth0/add-user-data-to-token-action.js#L7

FFawzy: - we can definitely add the missing `company` field to the token.
- @nkuehn I'll add a task for me to document these somewhere.
- the million dollar question here is, does the `user` object from the `useAuth0` hook ... update by itself after updating a user through the api ? or is there a specific parameter that triggers this? @gabriele-ct did you try this ?
if yes, then it will be easier to leave out your fetch

gabriele-ct: I've implemented the following changes:
- the user profile is read from the token upon login and stored in the context.
- All the parts of the application where profile informations are needed are now using LearningContext (apart from avatar which I'll update once Timon's changes are merged)
- Upon user profile update, the updated profile returned from the API is used to update the app LearningContext

zbalek:
reviewed the implementation - all good 🟢

packages/gatsby-theme-learning/src/hooks/use-update-user.ts
```diff
@@ -0,0 +1,93 @@
...
+  const [data, setData] = useState<QuizAttempt | undefined>();
+  const [correlationId, setCorrelationId] = useState<string | undefined>();
+
+  const updateUser = useCallback(
+    async (newUser: User): Promise<Response> => {
+      const apiEndpoint = `${learnApiBaseUrl}/api/users/${userId}`;
+      const accessToken = await getAuthToken();
+      const data = await fetch(apiEndpoint, {
+        method: 'POST',
+        body: JSON.stringify(newUser),
```
FFawzy, line 27: I think you need to add the `credentials` property here. ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included))

It is responsible for sending the `user_session` cookie along with the request, without it we will be creating a new learn-api JWT with every request.

Its a bit tricky as the possible values are `include`, `same-origin` or `omit`
I guess we can make it `credentials: 'include'` on testing only. which will include the cookie cross-domain
but make it; `same-origin` on production. because at that point `docs.commercetools.com` and `api.learn.commercetools.com` are same domain 

WDYT?

gabriele-ct: I implemented what you suggested, maybe you can have a check and see if it's working as expected

packages/gatsby-theme-sso-ui-kit/src/sso.const.ts
```diff
@@ -0,0 +1,4 @@
+export const AUTH0_CUSTOM_CLAIM_NS = 'https://docs.commercetools.com/';
+export const AUTH0_CLAIM_DISPLAYNAME = `${AUTH0_CUSTOM_CLAIM_NS}display_name`;
+export const AUTH0_CLAIM_GLOBAL_ACCOUNT_NAME = `${AUTH0_CUSTOM_CLAIM_NS}global_account_name`;
```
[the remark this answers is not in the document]
FFawzy: @nkuehn we have both
`global_account_id` and `global_account_name`

FFawzy (approved):
🥇

packages/gatsby-theme-learning/src/hooks/use-update-user.ts
nkuehn: P.S:: from a security standpoint "include" is okay. The actual locking down happens in the CORS headers that the API server sends and that has an allowlist for our preview and production domains.

PR #1615, opened 2023-04-17
Adjust avatar component according to Zoe's design

### LINKED ISSUE #1620: [SelfLearning] adjusting the `login to see quiz` button
Zoe has adjusted the design of the button to login to take a quiz.

[Link to Figma board
](https://www.figma.com/file/GqhHFUcEaMxBXnMM9TXdaZ/%F0%9F%9F%A2-Feedback-Board?node-id=324-3368&t=8UhPubea93XA8b4y-0)

timonrey:
closes https://github.com/commercetools/learning-tech/issues/390
closes #1620 

If we want to use the dropdown function for the avatar again, we can simply look into the file history to get the code back. No need to keep all of that as comments in my opinion.

**This pull request**
- adds a new flag called `hideLogin` to the gatsby config that toggles the top bar login information for the microsite
- implements the design changes for the `login, logout, avatar and login button for the quiz`
- introduces the new `Sign up button`
- refactors the avatar component and the component in the `overrides` folder that is now called `topbar-login-info`


packages/gatsby-theme-sso-ui-kit/src/components/login-button.tsx
```diff
@@ -30,20 +31,37 @@ const LoginButton = (props: LoginButtonProps) => {
       : window.location.pathname;
   };
 
+  if (props.quizId) {
+    return (
+      <SecondaryButton
```
timonrey, line 36: I kept the login button for the quiz so far. I'm not sure if we still want to use it in some way or if we can simply remove it and show something else in the quiz box when the user is not logged in.

nkuehn:
@timonrey  I think we should only have the login on the smoke test site now - it's very misleading on the public kit documentation sit.  

FYI  @FFawzy  maybe we can additionally add a big fat warning to the signup screen of our testing auth project or even restrictt that to commercetools.com  addresses.

nkuehn:
@zbalek  could you review this please (you can find the preview link in the autogenerated comment). 

I am not really convinced of the alignment and spacing in the top bar vs. the grid of the content below.  It feels a bit randomly positioned)

image

FFawzy:
> @timonrey I think we should only have the login on the smoke test site now - it's very misleading on the public kit documentation sit.

@nkuehn, The sso package is added to the main site because when you login in to auth0 the redirection happens to the baseUrl (the main microsite) and then internally it routes to the internal callback route, if we removed the sso package from the main microsite the redirection after auth0 login won't work.

But we need to find a way to hide the buttons even tho the package is configured in the gatsby config, (eg a list of allowed microsites or something)


> FYI @FFawzy maybe we can additionally add a big fat warning to the signup screen of our testing auth project or even restrictt that to commercetools.com addresses.

the docs-kit was directly connected to the production auth0 app, (since the beginning of the project, we had learning-tech connected to testing, docs-kit and docs connected to production) 
we can change that definitely and disable signups for docs-kit then or only allow @commercetools emails
I'll create a ticket for that

nkuehn:
> because when you login in to auth0 the redirection happens to the baseUrl (the main microsite) and then internally it routes to the internal callback route

Argh, thanks for the reminder that this was a mess. Looks like we really need a toggle whether to show the user avatar and logins.   But it does not have to be a list of sites etc - could be a flag set on the sso plugin configuration only?

timonrey:
> Argh, thanks for the reminder that this was a mess. Looks like we really need a toggle whether to show the user avatar and logins. But it does not have to be a list of sites etc - could be a flag set on the sso plugin configuration only?

I can implement an optional flag for the sso configuration that will allow us to toggle the login/logout buttons for each microsite. I will create a separate PR for that...it's out of the scope of this pull request.

FFawzy:
@timonrey 
if you do it in a separate PR, this means that once you merge this PR the button will be visible publicly in the docs-kit site in the main landing page which is something we are trying to avoid.
and if its delayed even more and you did a kit release this feature will arrive to the docs site which will make the button suddenly appears in the main docs site too.

nkuehn:
Agree. It must be done here, it's a part of the intended behavior of this PR and not something else.

packages/gatsby-theme-sso-ui-kit/src/@commercetools-docs/gatsby-theme-docs/overrides/avatar.js
```diff
@@ -208,21 +139,19 @@ const UserProfile = () => {
       });
     }
 
-    return <UserInformation userData={userData} logout={logout} />;
+    return <LoggedInState userData={userData} logout={logout} />;
```
gabriele-ct, line 142: I think we are still pushing `association` in the `userData` object. That should be not needed anymore as we don't display the company information anymore

packages/gatsby-theme-sso-ui-kit/src/components/login-component.spec.tsx
```diff
@@ -26,17 +26,11 @@ describe('login-button', () => {
...
-    expect(defaultIcon).toBeInTheDocument();
-  });
-
-  it('should render the specified icon', () => {
+  it('should render the specified icon for the login button inside of the quiz component', () => {
     const CusotomIcon = () => {
       return <div data-testid="custom-icon"></div>;
     };
-    render(<LoginButton icon={<CusotomIcon />} />);
+    render(<LoginButton quizId="123" icon={<CusotomIcon />} />);
```
gabriele-ct, line 33: is this hardcoded value needed?

packages/gatsby-theme-sso-ui-kit/src/@commercetools-docs/gatsby-theme-docs/overrides/avatar.js
```diff
@@ -1,179 +1,110 @@
 // A React component to be rendered in the top bar next to the top menu toggle button
-import { useEffect, useState, useContext } from 'react';
+import { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-import md5 from 'md5';
-import { useSelect } from 'downshift';
```
gabriele-ct, line 5: I think `downshift` dependency can be removed at all from `package.json`

timonrey: Right! Good catch 👍🏼

packages/gatsby-theme-sso-ui-kit/src/components/login-button.tsx
FFawzy: there is an issue already for this https://github.com/commercetools/commercetools-docs-kit/issues/1620

zbalek:
reviewed the implementation - all good 🟢

packages/gatsby-theme-sso-ui-kit/src/components/avatar.js
```diff
@@ -0,0 +1,128 @@
...
+const LoggedOutState = () => {
+  const { loginWithRedirect } = useAuth0();
+
+  return (
+    <AvatarContainer>
+      <LoginButton data-test-id="login-button" label="Login" />
+      <PrimaryButton
+        onClick={() =>
+          loginWithRedirect({
+            authorizationParams: {
```
FFawzy, line 71: @timonrey 
I think you need here the appState and returnTo parameter as in the login button .. otherwise the user will not be returned to the right page after signup

packages/gatsby-theme-sso-ui-kit/src/components/avatar.js
```diff
@@ -0,0 +1,128 @@
...
+  );
+};
+
+LoggedOutState.displayName = 'LoggedOutState';
+
+const UserProfile = () => {
+  const { isAuthenticated, logout, getIdTokenClaims } = useAuth0();
+  const [customClaims, setCustomClaims] = useState({});
+  useEffect(() => {
+    const fetchCustomClaims = async () => {
```
gabriele-ct, line 90: Here we should be using the learning context to get `email` and `displayed name`. However, the learning context belongs to the to the theme learning plugin and we don't have the guarantee that both plugins are running together.

As we discussed it's worth probably move everything into the `gatsby-theme-docs` package and not have to worry too much about these issue

(nothing to do for now @timonrey ) but refactoring is needed

PR #1622, opened 2023-04-19
Toggle login/logout buttons over the `hideLogin` flag in the sso plugin config

timonrey:
part of #1615 
Sorry, I was too fast with the new PR 😅 
But we can merge it into the other pull request. It's also easier for you to review this way :)

### DIFF
--- a/websites/documentation/gatsby-config.mjs
+++ b/websites/documentation/gatsby-config.mjs
@@ -27,6 +27,7 @@ const config = {
             auth0Domain: 'auth.id.commercetools.com',
             auth0ClientId: 'xLk8EDUCc8PKqCbrSJCnuahvn86nEn4z',
             learnApiBaseUrl: 'https://api.learn.commercetools.com',
+            hideLogin: true,
           },
         },
       ],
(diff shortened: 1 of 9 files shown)


websites/documentation/gatsby-config.mjs
```diff
@@ -27,6 +27,7 @@ const config = {
             auth0Domain: 'auth.id.commercetools.com',
             auth0ClientId: 'xLk8EDUCc8PKqCbrSJCnuahvn86nEn4z',
             learnApiBaseUrl: 'https://api.learn.commercetools.com',
+            hideLogin: true,
```
FFawzy, line 30: I can see you have the flag here set as true, so in theory the buttons should be hidden .. but in the preview link they are still visible.

Also, I'm looking at the rest of the code, I can see you are declaring the flag but never used it somewhere or maybe I missed it somewhere

timonrey: Yes because I only implemented the flag in this PR, but I'm not using it yet...but I can do that here already 👍🏼

PR #1625, opened 2023-04-19
feat: confirmation modal upon course completion

gabriele-ct:


gabriele-ct:
FYI, this is how the modal looks like when the user completes the last course in the learning path
Screenshot 2023-04-21 at 11 03 25

nkuehn:
> FYI, this is how the modal looks like when the user completes the last course in the learning path
Nice, independently of what we think about where the button leads to the situational language is nice.

packages/gatsby-theme-docs/src/layouts/content.js
```diff
@@ -22,8 +22,14 @@ import LayoutPageNavigation from './internals/layout-page-navigation';
 import LayoutPageContent from './internals/layout-page-content';
 import PageContentInset from './internals/page-content-inset';
 import PageReadTime from './internals/page-read-time-estimation';
+import {
+  CourseCompleteModal,
+  useCourseInfoByPageSlugs,
+} from '@commercetools-docs/gatsby-theme-learning';
```
FFawzy, line 28: wouldn't this make learning package a dependency for theme docs?

FFawzy (approved):
Thank you 👏

packages/gatsby-theme-docs/src/layouts/content.js
gabriele-ct: I think this is what we had already discussed with @nkuehn.
Learning theme is going to be a dependency of theme docs as learning is not really an independent gatsby plugin. The final goal is to  move all the learning implementation into the docs-theme to avoid too many packages.
The other way is to create many overrides but as the application gets more and more complex we will end up with lots of meaningless overrides.

FFawzy: yes you are right, I just thought we missed adding it to the package.json since I didn't see it in the changed files but It is there from a different PR

All good then 🙌

Issue #1610, opened 2023-04-11
[Moodle Investigation] make self-learning e2e more deterministic


FFawzy:
Using a single test user is causing issues as we don't always begin the e2e test from a clean state. We need to think about a way to either reset a specific user's attempts, or create a new user each time (and then delete it at the end of the test). 

With coming new features this will become soon a blocker so better to start having a thought

FFawzy:
+ having a way to reset completion status if possible
https://github.com/commercetools/commercetools-docs-kit/pull/1632#pullrequestreview-1402657201

FFawzy:
consider deleting the user from moodle
providing an endpoint that takes care of deleting the test user only

gabriele-ct:
`core_user_delete_users ` seems to be working as expected also because the new user has a different userid so it cannot reference any existing tracking data.
Here I see 3 possible issues:
- Running e2e will "burn" a lot of user ids (it shouldn't be a big problem) but it's worth to mention
- We need to add an additional call to get the userid based on the user email
- There might be concurrency issues if 2 e2e are running at the same time the second might delete the user used by the first one

FFawzy:
```mermaid

sequenceDiagram
  participant A as Browser
  participant B as Auth0
  participant C as Learn API
  participant D as Moodle
  A->>B: Initiate Signup process
  B->>A: account created
  A->>B: Initiate login process
  B->>A: logged in + JWT
  alt whatever e2e test cases 
    A->>C: get attempt
    C->>C: do stuff
    C->>D: get attempt
    C->>A: attempt
 end
  A->>C: Delete user + session cookie + JWT (get user details from cookie)
  C->>D: delete user
  C->>B: delete user
  C->>A: 200

```

maybe we can do something like this WDYT? @nkuehn @gabriele-ct

FFawzy:
One disadvantage of the above approach (sequence diagram), it will increase the costs of Auth0 as we are charged for monthly active users .. and this way we will have a lot of test users instead of just one

gabriele-ct:
@FFawzy, thank you for taking the time to make the diagram.
my 2 cents. If we go this way:
- Signup and user deletion should happen via API endpoint. It would be great to have a `users/create-test-user` that returns the username(email) and password? of the freshly created test user. And a specular `users/delete-test-user/[user-email]` that does the clean up.
- If we implement the point above we could even go one step further and delete the user both form moodle and auth0 (don't know how complex this could be)
[...]

PR #1632, opened 2023-04-21
feat: implement course and learning path cards

gabriele-ct:


gabriele-ct:
Note that, having updated all the uikit dependencies to the 15.15.1 version (in order to get the latest `Stamp` component implementation), some ui element have slightly changed causing several VRT errors. In particular:
- stamp element font is larger than previous implementation
- copy icon has been replaced

gabriele-ct:
Logged out state:
Screenshot 2023-04-24 at 11 00 31

Logged in state (with all the possible course states):
Screenshot 2023-04-24 at 11 09 20

FFawzy (approved):
🥇

gabriele-ct:
Since the components were related and using shared code, I've added to the PR also the implementation of the LearningPathCard (see comments in the code)

websites/docs-smoke-test/src/content/index.mdx
```diff
@@ -29,3 +31,27 @@ title: Explore how to build a documentation website for commercetools
...
+  <CourseCard duration="30 min" title="Course name" href="/">Course description. Introduction to extensibility possibilities available in Composable Commerce. </CourseCard>
+</Cards>
+
+# Learning path card
+
+<Cards>
+  <LearningPathCard title="Overview" duration="1 hour 25 minutes | 7 courses" productName="Composable Commerce">
+    This learning path is great for Composable Commerce administrators who create/maintain e-commerce data points, primarily work with a user interface, and have some familiarity with APIs.
+  </LearningPathCard>
+  <ImageCard>
```
gabriele-ct, line 52: I thought to create a specific card just as image container, so that we can include it in the `Cards` component and get all the layout/responsivness features already implemented.

I haven't found a way to actually import the svg file into mdx so I tried to use the image syntax to load it as image and then wrap it into the `ImageCard` so that it can be positioned correctly.

gabriele-ct: I'm running out of ideas but I',m open to any other solution @nkuehn

nkuehn: Concerning the images: For right now whatever works is good.  

Take note of the problem separately, we'll need an author-friendly way to use SVGs in the MDX fiels as images that are not a content image.

websites/docs-smoke-test/src/content/index.mdx
```diff
@@ -29,3 +31,27 @@ title: Explore how to build a documentation website for commercetools
...
+# Course cards
+
+<Cards narrow clickable>
+  <CourseCard duration="30 min" title="Course name" href="self-learning/overview" courseId="66">Course description. Introduction to extensibility possibilities available in Composable Commerce. </CourseCard>
+  <CourseCard duration="30 min" title="Course name" href="/">Course description. Introduction to extensibility possibilities available in Composable Commerce. </CourseCard>
+  <CourseCard duration="30 min" title="Course name" href="/">Course description. Introduction to extensibility possibilities available in Composable Commerce. </CourseCard>
+  <CourseCard duration="30 min" title="Course name" href="/">Course description. Introduction to extensibility possibilities available in Composable Commerce. </CourseCard>
+  <CourseCard duration="30 min" title="Course name" href="/">Course description. Introduction to extensibility possibilities available in Composable Commerce. </CourseCard>
+  <CourseCard duration="30 min" title="Course name" href="/">Course description. Introduction to extensibility possibilities available in Composable Commerce. </CourseCard>
+</Cards>
```
nkuehn, line 44: This is really nice for authors :-) 

For smoke testing purposes you may want to include a card where the author chose to put extra blank lines and  ( or markdown into the card to see the layout effects.

nkuehn (commented):
### Visual

Visually the Stamp looks to me like it's aligned to the top of the row with the time, leaving more space to the bottom than it should.  I wasn't able to quickly spot how to do it (had hoped on `margin-top: auto`, too). 

While trying I found that I can remove the custom min-heights and heights in the bottom of the card and the layout being OK again if both are removed ([nr1](https://github.com/commercetools/commercetools-docs-kit/pull/1632/files#diff-16a62515b08838ea8baa95d93c1d211f9eb578f322062a372e584663d7833fa5R80)   [nr2](https://github.com/commercetools/commercetools-docs-kit/pull/1632/files#diff-16a62515b08838ea8baa95d93c1d211f9eb578f322062a372e584663d7833fa5R56) ). Maybe these are artifacts of earlier iterations? 

### Behavior @FFawzy 

The testing setup should maybe needs two or even three courses to be able to see the different states more clearly in VRTs with the test user. 

I tried to emulate seen the "not completed" state by re-taking a quiz with wrong answers, but  that led to the course still being completed (both in the card stamp and the side nav).  It's maybe logically correct (I did pass the course before) but UX  / visually inconsistent.   
It's not a big thing because external Users can't re-take quizzes after they passed but internals can and we need some way to test.

FFawzy:
> The testing setup should maybe needs two or even three courses to be able to see the different states more clearly in VRTs with the test user.
> 
> I tried to emulate seen the "not completed" state by re-taking a quiz with wrong answers, but that led to the course still being completed (both in the card stamp and the side nav). It's maybe logically correct (I did pass the course before) but UX / visually inconsistent.
> It's not a big thing because external Users can't re-take quizzes after they passed but internals can and we need some way to test.

retrying the quiz won't change the fact that you actually completed the course, (how moodle works)
you need to delete the completion data of the course in order for the course completion to actually change.

in any case We had a ticket for this and I added this usecase in it too ... I'll try to figure out something.

https://github.com/commercetools/commercetools-docs-kit/issues/1610

Issue #1617, opened 2023-04-17
[SelfLearning] Learning Path Card


FFawzy:
Goal
a card component with the following props:

- duration
- product name
- title

The following feature is going to be implemented in a separate PR
A wrapper in the mdx page that is responsible for displaying content if the user is not logged in only including
- The Learning path card 
- the signup CTA button

[Link to Figma Board](https://www.figma.com/file/GqhHFUcEaMxBXnMM9TXdaZ/%F0%9F%9F%A2-Feedback-Board?node-id=340-3361&t=8UhPubea93XA8b4y-0)

gabriele-ct:
@nkuehn about the logged in/logged out wrappers:
Do you have any idea about how to handle upon a full page refresh or minisite change the state when auth0 is still loading and the user doesn't appear as authenticated?

This is the current behaviour:


![Image](https://user-images.githubusercontent.com/106664971/233975438-301ff7c5-5e8f-4a6d-8662-d63d483fa2c6.gif)

nkuehn:
> Do you have any idea about how to handle upon a full page refresh or minisite change the state when auth0 is still loading and the user doesn't appear as authenticated?
> 
> As you can se there's a moment when the logged out content appears when we hard refresh the page, I think this is unavoidable

@gabriele-ct  This looks like having to retain cached local state like we intend to do with SWR's localstorage cache.  But in any case we'll have to live with the fact that we don't know anything until the static page is hydrated.  

An Idea:  Maybe give the component a boolean (set vs. not set) prop like "optimistic" or "assumeTrue" that allows the author to define whether to assume logged in or not until we know something.  if it's set the static pregeneration would render independently of what the hook delivers.  Otherwise we would not render anything until the page hydrates (we cannot prevent the layout shift in this case because the space cannot be reserved because it's unknown what it will contain)

gabriele-ct:
Good idea, I implemented that. Thank you

PR #1634, opened 2023-04-24
feat: add IsUserLoggedIn or IsUserLoggedOut component

gabriele-ct:
contributes to #1617


websites/docs-smoke-test/src/content/index.mdx
```diff
@@ -29,3 +30,13 @@ title: Explore how to build a documentation website for commercetools
...
+
+# Logged in or logged out content
+
+<IsUserLoggedIn>
+ ## Welcome back!
+</IsUserLoggedIn>
+
+<IsUserLoggedOut>
+ ## Please log in
+</IsUserLoggedOut>
```
nkuehn, line 42: Would it be more readable or "template language like"  if it were `<IfUserLoggedIn>`  vs. `<IsUserLoggedIn>` ?

websites/docs-smoke-test/src/content/index.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 title: Explore how to build a documentation website for commercetools
 ---
+import { IsUserLoggedIn, IsUserLoggedOut } from '@commercetools-docs/gatsby-theme-learning'
```
nkuehn, line 4: I think this component is small enough to be provided automatically through the shortcodes file.

websites/docs-smoke-test/src/content/index.mdx
gabriele-ct: Yes, agreed

websites/docs-smoke-test/src/content/index.mdx
gabriele-ct: Sure, makes sense

nkuehn:
P.S:  to reduce the flicker time on reloads we could set the Auth0 Provider cache to localstorage: https://auth0.github.io/auth0-react/interfaces/Auth0ProviderOptions.html#cacheLocation  

This would generally improve UX when switching between microsites

Issue #1633, opened 2023-04-24
Choose API domain based on the current deployment URL


gabriele-ct:
Client side implementation of the change implemented here https://github.com/commercetools/learning-tech/pull/409#pullrequestreview-1397543886

PR #1636, opened 2023-04-25
Change learning api host to and include credentials

nkuehn:
Addresses #1633 

(conditional switching exists only in the docs here - it's to be applied in the actual content repos)


gabriele-ct (approved):

nkuehn:
```
CypressError: Timed out retrying after 4000ms: 
The command was expected to run against origin https://docskit-sha-a7521d078dc62263fa709da1d0f41896d5afa9a5.commercetools.vercel.app 
but the application is at origin https://auth.id.commercetools.com

This commonly happens when you have either not navigated to the expected origin or have navigated away unexpectedly.
```
My first thought was that the login may be redirecting back but cypress still acts as if it's still working on the auth.id...  host ("or have navigated away unexpectedly" in the error message).  Or the other way around?  

What is really puzzling is how this relates to the change in this PR and why it runs fine in the browser if used manually.  

It is reproducible in that PR and other PRs seem to run the test fine at the moment. 

Issues I found:
 * https://github.com/cypress-io/cypress/issues/25911#issuecomment-1449284686  (not a convincing solution, looks just like a trial and error result. Manually does a navigation to the right page on the actual site after the login step but that would not test any more if the redirection is working)

gabriele-ct:
Very weird behaviour.
Running e2e headed I can see the login happening but the logout redirection fails displaying the auth0 error page.
Running local server I can't login, or, for better saying, I login but I immediately get logged out.
Better double check tomorrow with fresh eyes

nkuehn:
> Running e2e headed I can see the login happening but the logout redirection fails displaying the auth0 error page.

@FFawzy  that sounds quite a bit like we have to first fix the Auth0 origins configuration - the details of the three different fields can be confusing. I scheduled a remeet for the three of us concerning the issue.

FFawzy (approved):

PR #1637, opened 2023-04-25
chore: fix lint warnings for missing hook dependencies

gabriele-ct:


nkuehn (commented):
i would rather keep those functions that we know do not actually mutate inside the dependencies and not disable the ESLint rule.  

If the function (e.g. `closeModal`) is not actually changing it does not hurt to have it as a dependency because it will not trigger unnecessary re-renders anyways if it does not change.  But if we disable the rule altogether we will not see actual issues any more when the dependencies of the hooks change later when we evolve the code. 

I think the rule only needs to be disabled if there is a frequently changing dependency that you deliberately want to be excluded from the dependencies because it changes frequently but that change should not trigger a re-render.    In that case I would explain the situation in a comment.

gabriele-ct:
I don't like this rule very much as it leads devs to not really think about what actually should trigger the function to run (aka what are the actual changing dependencies). I've removed the disable lint where you suggested.

nkuehn:
>  I don't like this rule very much

I also first though it's just too mechanic and there should be thought behind it etc.   But while trying to find out about why and whether functions or not I saw that Dan Abramov made  boldfaced extrastrong statement that usually you should not turn off the rule because it will just bite you later ( https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528 ).  I then chose to not even try to fully understand it but follow that advice - the real boss spoke and I followed :-) 

Wording in that endless github thread sometimes ended with somthing like "if you want to live on the edge do it, but otherwise... ".

nkuehn (approved):

PR #1643, opened 2023-04-27
fix: useeffect dependency causes application crash

gabriele-ct:


nkuehn (commented):
I'm sorry. I had asked for this - I don't know what information to trust any more.  React docs say we should add all dependencees but it crashes nevertheless.

PR #1645, opened 2023-04-28
Ga learning index

gabriele-ct:


gabriele-ct:
@FFawzy 
PR is ready to be reviewed. When you have some spare time we can discuss this together. The layout is not 100% identical to the design but the mdx has some limitations

gabriele-ct:
Thanks to @timonrey ! I'm now using the same custom primary button used for the avatar area

websites/docs-smoke-test/src/content/index.mdx
```diff
@@ -61,18 +101,36 @@ This is a second **paragraph**
...
   </ImageCard>
 </Cards>
 
-# Logged in or logged out content
 
-<IfUserLoggedIn>
- ## Welcome back!
-</IfUserLoggedIn>
+<div style="display:flex;">
+  <PrimaryButton onClick={() => window.location.href = 'https://commercetools.eu.auth0.com/authorize?response_type=code&client_id=xLk8EDUCc8PKqCbrSJCnuahvn86nEn4z&redirect_uri=https://docs.commercetools.com&scope=openid%20profile'}>
```
FFawzy: it could be something like this (did not test it)

```suggestion
  <PrimaryButton onClick={() => window.location.href = `https://auth.id.commercetools.com/authorize?response_type=code&client_id=xLk8EDUCc8PKqCbrSJCnuahvn86nEn4z&redirect_uri=${window.location.origin}&scope=openid%20profile`}>
```

FFawzy: @gabriele-ct

FFawzy:
I think we are not going to able to use the button with the url, otherwise we need to handle the redirect ourselves.

best thing, I'd say use the button from the SSO kit directly in here if possible

gabriele-ct:
@FFawzy I've refactored a little the `LoginButton` and this works as expected.
Please @timonrey have a look to the refactoring and let me know if you see any issue

Issue #1642, opened 2023-04-27
SWR cache revalidation


gabriele-ct:
if taken literal none of this happens in the case I tested.
- `revalidateOnFocus` would have required me to switch between tabs or windows.  I did not actually do that when I tested.  I’ll try again and be mindful what exactly I do because I think focusing (clicking) something in the dev console counts as a tab / window switch
 - `refreshInterval` is off default and would have behaved differently
- `revalidateOnReconnect` can be ruled out here, network was stable and the timing of the calls was clearly associated to when I clicked the nav entries
 - `revalidateIfStale`  “The revalidateIfStale controls if SWR should revalidate when it mounts and there is stale data.” .  The question here is what is “when it mounts”.   Maybe a navigation between pages makes SWR unmount and remount?
 - in the options documentation (link) there is also revalidateOnMount - it sounds similar to me but rather on component level. If the specific component using the hook with that option is remounted the data would be considered stale.
 
A first thing I think we should do (and intend to do anyways to provide the localstorage) is to inject  the SWRconfig  at a high level in the app tree so that it is not remounted on page changes.  ( https://swr.vercel.app/docs/advanced/cache.en-US#create-cache-provider   )  That would give us explicit clarity to the scope and lifetime of a given SWR cache, even without providing a custom one like localstorage.    Ideally, we can stick to the defaults if it works.
“revalidateIfStale” may be a bit too much - e.g. it would revalidate on every microsite switch but better safe than sorry for now I would say.

gabriele-ct:
@nkuehn I've been digging a bit into this task and I've learning that by default gatsby unmounts and re-mounts the whole layout component at each [route change ](https://www.gatsbyjs.com/docs/how-to/routing/layout-components/#how-to-prevent-layout-components-from-unmounting)
That's why we get `revalidateIfStale` triggered each time we navigate the minisite.
There are some approaches to avoid this behaviour. Gatsby mentions [this plugin](https://www.gatsbyjs.com/plugins/gatsby-plugin-layout/) that apparently should prevent the un-mount/re-mount 

I'd like to know what are your thoughts about the possible solutions and if I should invest time in trying out the above mentioned plugin

nkuehn:
@gabriele-ct  I would not mess with gatsby's runtime model, that just causes technical dept.  Layouts are not the right place for such a use case anyways because it has nothing to do with the appearance of a certain page vs. other pages.  

[...]

PR #1652, opened 2023-05-03
fix: force swr cache invalidation

gabriele-ct:
This PR also addresses #1642


FFawzy:
@gabriele-ct apparently some of the new configs we are passing to SWR is breaking the `/courses` call.
it is not rendering properly now .. can you check please?

gabriele-ct:
@nkuehn and @gabriele-ct  I tried to implement the first solution described here: https://github.com/commercetools/commercetools-docs-kit/issues/1642#issuecomment-1534534495
It seems to be working fine but I'd like your opinion too

FFawzy:
Thanks @gabriele-ct for the research and explanations, I'd say the first option does the job for now too. 
I tried it and it is working as expected

FFawzy (approved):
Thank you 🙌

packages/gatsby-theme-learning/src/hooks/use-course-details.ts
```diff
@@ -39,7 +39,14 @@ export const useFetchCourseDetails = (
 
   const { data, error, isLoading } = useSWR(
     shouldFetchData ? apiEndpoint : null,
-    (url) => fetcherWithToken(url, getAuthToken, learnApiBaseUrl, 'GET')
+    (url) => fetcherWithToken(url, getAuthToken, learnApiBaseUrl, 'GET'),
+    {
+      revalidateOnReconnect: false,
+      revalidateIfStale: false,
+      revalidateOnMount: true,
+      dedupingInterval: 10000000,
```
gabriele-ct, line 47: @nkuehn regarding a reasonable deduping interval. This is clearly too aggressive: 160 hours meaning that the data will never be automatically revalidated. Shall we look more into minutes? or other ranges? Just to have a rough idea

packages/gatsby-theme-learning/src/hooks/use-attempt.ts
```diff
@@ -71,6 +77,10 @@ export const useAttempt = (fetchAttemptParams: FetchAttemptParams) => {
           }
           throw new Error();
         } else {
+          // we invalidate course status cache as the user might be just being
+          // enrolled into the course so we want fresh course status data to be fetched
+          // see https://github.com/commercetools/commercetools-docs-kit/issues/1644
+          invalidateCache();
```
nkuehn, line 83: @FFawzy  okay for now but my learning is that the API proxy (server) side "knows" much better when other resources might be stale or not. In this case, the API tier "knows" that it auto-enrolled and had to create an attempt for the first time, so it knows that the course is stale - but the client can only guess. 

So if this kind of situation becomes more commonplace we could consider rather letting the API hint to the client what other resources to mark stale now vs the client having to guess.

packages/gatsby-theme-learning/src/hooks/use-course-details.ts
```diff
@@ -39,7 +39,14 @@ export const useFetchCourseDetails = (
...
+      revalidateOnFocus: false,
```
nkuehn, line 48: @gabriele-ct  I would prefer to leave this set to "true" - it's a really good way to prevent user confusion when they work in multiple browser tabs when taking a course.  I had understood you only turned it on to have less noise when debugging.  

Similar but less clearly a use case for "revalidateOnReconnect" - that is rather a sign of the computer having gone to sleep in the meanwhile which is a good proxy for a user returning after a longer pause.

PR #1661, opened 2023-05-05
Ga json primitive types

gabriele-ct:


api-specs/test/api.raml
```diff
@@ -475,6 +475,21 @@ uses:
...
+        body:
+          application/json:
+            type: string
+            example: !include examples/action-payload-1.json
+        responses:
+          200:
+            description: We recommend to use standard HTTP response codes and `application/json` encoded content. The response will look like the response you have declared in your action. As an example we will fetch the cart.
+            body:
+              application/json:
+                type: object
```
nkuehn, line 490: The response type of the API described as "any JSON serializable payload"  would be `any`, not `object`.     E.g. `"just a string"` is valid JSON, or  `[1, 2, "string", false]` is, too.

api-specs/test/examples/action-success-1.json
nkuehn, line 1: As a test case, I would rather choose an edge case that might break (for Object just an empty object, or for `any` a raw primitive) to make sure that no other place in the codebase is assuming any JSON would be an object all the time.  It would also be less noisy in the testing data.    Finally, a meaningful file name would help maintainability of the test data.

packages/gatsby-theme-api-docs/src/components/type/type.js
```diff
@@ -12,13 +12,44 @@ import Enum from './enum';
...
+    number: '_Any JSON "number"',
+    any: '_Any valid JSON_',
+    object: '_Any JSON "object"_',
+    boolean: '_Any JSON "boolean"_',
+    string: '_Any JSON "string"_',
+    array: '_Any JSON "array"_',
+  },
+};
+
+export const getPrimitiveTypeByName = (contentType, type) =>
```
nkuehn, line 26: (just an idea because it was misleading me in my understanding of the code that uses the function. It's not returning a type or type name, it's returning the description for a type if it's by chance a primitive)

```suggestion
export const getDescriptionIfPrimitiveType = (contentType, type) =>
```

packages/gatsby-theme-api-docs/src/components/type/type.js
```diff
@@ -12,13 +12,44 @@ import Enum from './enum';
...
 import Examples from './examples';
 
+const contentTypeToPrimitiveMap = {
+  'application/json': {
+    number: '_Any JSON "number"',
+    any: '_Any valid JSON_',
+    object: '_Any JSON "object"_',
+    boolean: '_Any JSON "boolean"_',
+    string: '_Any JSON "string"_',
+    array: '_Any JSON "array"_',
```
nkuehn, line 22: Since the only formatting we chose to use here is making the whole thing italic and the values are hardwired in the codebase I suggest to leave out the (relatively expensive) markdownFragmentToReact  step and format the output natively using React.  The components representing the various markdown features are all available in the docs-ui-kit.    
If you find it more expressive, the map here could alternatively return a component so you can use JSX for the values directly there.  But I would rather not overdo it if there is no need for more formatting.

packages/gatsby-theme-api-docs/src/utils/render-type-as-link.js
```diff
@@ -2,19 +2,35 @@ import React from 'react';
...
+    Array.isArray(contentType) &&
+    contentType.includes('application/json')
+  ) {
+    primitiveJsonType = getPrimitiveTypeByName('application/json', type);
+  }
 
+  if (originalTypeLocation) {
+    return <Link href={originalTypeLocation}>{type}</Link>;
+  } else if (primitiveJsonType) {
+    return markdownFragmentToReact(primitiveJsonType);
```
nkuehn, line 30: (concerning above suggestion to illustrate the idea)
```suggestion
    return <Markdown.Em>{primitiveJsonType}</Mardown.Em>;
```

nkuehn (commented):
Nice, I have some detail feedback but the general approach looks good.  Could maybe even serve as the basis for being able to model image uploads and other atypcial mime types more flexible if the next use case in that area comes up.

nkuehn (approved):
It's really nice in the VRT :-) 

And also generally nice to see how the API docs kit is maturing for more generic use beyond the pattern of the composable commerce APIs.

PR #1669, opened 2023-05-09
chore: display empty circle indicator if unenrolled

gabriele-ct:


nkuehn:
nice to see this starting!  
I'll keep out of the details, can you maybe clarify with Zoe whether it's intentional that a state like this (see screen) can happen? The page has a status but the section does not.   It's my status in the test site at the moment.   If the original intent is that there is some sort of symbol in any situation it may have been just underspecified what that means. 

image

gabriele-ct:
@nkuehn The topics status and the course (section) status come from different endpoint. This means they're not in sync (one call can take longer than the other) and in case of errors with the API, we currently don't display anything.
IMHO we need to clearly define the possible statuses for both indicators (course and topic):
- Logged out
- Logged in but not enrolled
- In progress
- Completed
Then we should make sure the API is clearly returning the status. In particular I discussed with @FFawzy that the `not enrolled` status is not that easy to figure out since the API is returning a rather generic error.

Finally we should agree on which states should have a specific indicator and have maybe a fallback in case of error.

FFawzy:
**we can confirm with @zbalek  on the different states**

Course level
1. logged out
2. logged in enrolled (AKA in progress)
3. logged in not enrolled
4. logged in loading
5. logged in error
6. logged in completed

Page level
1. logged out
2. logged in enrolled (AKA in progress)
3. logged in not enrolled
4. logged in loading
5. logged in error
6. logged in completed

**the current implementation is**  (@gabriele-ct please confirm if something is different, or any state I missed)
logged out = reserving space
any other state is empty circle unless it is completed then it is a filled circle

zbalek:
In cases where there's a discrepancy between page and course loading, please keep an empty circle. I.e. if page loads but a course doesn't = keep empty circles for the other pages in the course, and if a course loads but a page doesn't = keep an empty circle for the page that didn't load.

FFawzy:
in other words @zbalek 
logged out = reserving space
any other state is empty circle unless it is completed then it is a filled circle

correct ?

nkuehn:
> logged out = reserving space
> any other state is empty circle unless it is completed then it is a filled circle

I just tested the implementation and after logging in, it still "waits" with blank (reserving space) until it loads the actual course status for the first time.   But if I get your (quoted above) summary right @FFawzy  at that time there should already be a grey circle because I am logged in and in doubt if we don't explicitly know that I have already done the topic or course it would be grey circle no matter what.   

Not sure if I got the logic right TBH, just asking stupid questions.

gabriele-ct:
@nkuehn Not stupid at all.
It was missing the loading state handling, which should as well display the empty circle.
I just pushed the change, I don't know why with all the various changes of requirements I had completely forgotten about it :)

nkuehn (approved):
(to unblock, I only did a rough test now)

FFawzy (approved):
Thank you

FFawzy (approved):
Thank you

PR #1671, opened 2023-05-11
[SL] Don't show login buttons in loading state

### LINKED ISSUE #1670: Login/Sign up buttons .. loading state
> If I’m logged in, and I reload the page, it shows “Login /sign up” buttons again until it detects the user is already logged in.
> It’s not a nice UX and might be confusing to users.
> I would suggest to think about the loading state. Maybe even don’t show anything until the info has been fetched (if possible).

issue reported by Nicola

Image

timonrey:
closes #1670

### DIFF
--- a/.changeset/large-tigers-hope.md
+++ b/.changeset/large-tigers-hope.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-docs/gatsby-theme-sso-ui-kit": patch
+---
+
+[SL] Don't show login buttons in loading state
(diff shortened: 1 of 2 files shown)


gabriele-ct (approved):

FFawzy:
Let me cross check this with @zbalek  too 

because now it takes a second to show the login button for the first time 

so its a trade off

FFawzy:
🟢 🟢 from @zbalek 
lets goo

PR #1675, opened 2023-05-12
Reduce home page flickering show relevant content only

gabriele-ct:
You probably have noticed that when a logged in user refreshes the learning minisite homepage sees logged out content first, then logged in content and, if he finished the learning path has a 3rd content change on the page as you can see in the video below:
https://vimeo.com/826184271/db6e4faced

This happens since some API calls need to happen in order to determine what content should be displayed to the user.
Although it doesn't cause a massive amount of layout shift I find it quite misleading for the user and rather unprofessional.

I'd like you to have a look to this different approach which is to display nothing until all the API calls have completed and we're sure about what content to display:
https://vimeo.com/826187147/082555513f

Obviously none of the above are ideal since there are technical limitations that we all know. However, I feel like the second approach feels somehow less amateur-ish than the first.

I'd like to know what's your opinion about it @zbalek @nkuehn @FFawzy @timonrey


gabriele-ct:
As note  I don't think the [`assumeTrue`](https://docs-kit.commercetools.vercel.app/documentation/writing/quizzes#ifuserloggedin--ifuserloggedout) helps much as the logged out user will see logged in content which is IMHO even worse

zbalek:
@gabriele-ct I think that the proposed approach makes sense. I was wondering though - would it be possible to implement something like this whilst we wait for the page to load?: https://uikit.commercetools.com/?path=/story/components-loading--loadingspinner @FFawzy @timonrey

PR #1676, opened 2023-05-16
chore: visual tweaks for mega menu with multiple products

gabriele-ct:


packages/gatsby-theme-docs/src/data/top-menu.yaml
```diff
@@ -9,3 +9,17 @@
       href: /../docs-smoke-test
     - label: API Docs Smoke Test
       href: /../api-docs-smoke-test
+
+- menu-title: Another product with long title
```
gabriele-ct, line 13: Fake menu items, just to give a taste of how the menu will work when having more than 2 items

timonrey (approved):
Looks like the suggested design form Zoe 👍🏼 
There is just one breakpoint where the column headline breaks again and the underline is not aligned with the other underlines anymore. I know it's an edge case but maybe there is a simple fix for it :)
Bildschirmfoto 2023-05-22 um 10 22 03



But other than that, it looks good enough to me.

gabriele-ct:
Thank you for you feedback @timonrey 
I think the fact that the title must be on maximum two line it's, at least for now, a hard requirement. Good that you notice that the mock data was actually spanning on 3 lines on a specific breakpoint. It will be editor's responsibility to ensure that the title fits all the breakpoints

PR #1683, opened 2023-05-19
feat: create a self-learning dedicated minisite

gabriele-ct:


package.json
```diff
@@ -24,11 +24,9 @@
     "labels:sync": "github-labels sync",
     "test": "jest --config jest.test.config.js",
     "test:watch": "yarn test --watch",
-    "test:e2e": "yarn test:e2e:docs-smoke-test && yarn test:e2e:api-docs-smoke-test",
```
timonrey, line 27: Why can't we have a script that runs all e2e tests for the three microsites? 🤔

.changeset/thin-foxes-ring.md
```diff
@@ -0,0 +1,10 @@
+---
+'@commercetools-website/self-learning-smoke-test': patch
+'@commercetools-docs/gatsby-theme-api-docs': patch
+'@commercetools-docs/gatsby-theme-docs': patch
+'@commercetools-website/docs-smoke-test': patch
+'@commercetools-website/documentation': patch
+'@commercetools-website/site-template': patch
+---
+
+Created a dedicated smoke test site for self-learning
```
timonrey, line 10: I think we don't need a changeset here...this is just for us internally.

timonrey (commented):
Nice! Just a few comments but overall it looks good 🚀 

One thing you could also do is to remove (or hide) the Login buttons from the docs-smoke-test. Do we even need the learning plugin there anymore?

FFawzy (commented):
couple of things for cleanup but the rest looks good already. Thank you 👏 

1- lets move the e2e to a separate job already .. in preparation of [this ticket](https://github.com/commercetools/commercetools-docs-kit/issues/1610) (if you want to do it in another PR .. fine too)

2- lets remove the learning and sso from docs-smoke-test at all .. so it should only be present in the self-learning-smoke-test and sso in the main site too.

3- maybe create another e2e course in Moodle, so your tests can run properly across multiple courses if needed  (duplicate the existing one so you get all the settings across, if you need help doing so LMK)

.changeset/thin-foxes-ring.md
gabriele-ct: You're right, but I don't want to create possible issues with other changesets, I'll leave it just in case

package.json
gabriele-ct: It's not used by automation and I never used it, but I can definitely leave it

gabriele-ct:
> One thing you could also do is to remove (or hide) the Login buttons from the docs-smoke-test. Do we even need the learning plugin there anymore?

You're right @timonrey , the plugins are not needed anymore, I'm removing them completely from the `docs-smoke-test`

gabriele-ct:
> 3- maybe create another e2e course in Moodle, so your tests can run properly across multiple courses if needed (duplicate the existing one so you get all the settings across, if you need help doing so LMK)

@FFawzy I've done that, I actually duplicated the existing e2e course at it is without changing the content. I think for the purpose of the e2e it's good enough and it's easier to detect the correct answers since the logic has been already implemented for the first course

gabriele-ct:
This is ready to be re-reviewed when you have time.
I'll also create a separate task to add VRT on the new site. At the moment we only run screenshots on the other smoke test websites but I think this PR is big enough

FFawzy (approved):
👏 Thank youu

one last thing, check that the content files has the right course id .. I was not seeing it in the files changed section so I couldn't verify it

gabriele-ct:
> one last thing, check that the content files has the right course id .. I was not seeing it in the files changed section so I couldn't verify it

It should be correct:
https://github.com/commercetools/commercetools-docs-kit/pull/1683/files#diff-4eca6a7b64d8cb05f285fefc7bfb23c44c9bf6bbb1ab93dd7ac9c8811e13cdff

The first course points to id 66, the other to 69 which is the newly cloned course

PR #1685, opened 2023-05-23
chore: login should be spelled as log in

gabriele-ct:


FFawzy:
learning package too ? for the quiz button ?

gabriele-ct:
> learning package too ? for the quiz button ?

True, also let's remember to change the login button in the minisite homepage, it still uses `login`

FFawzy (approved):
👍

PR #1688, opened 2023-05-23
Instrument Analytics: Learning interactions, microsite name, login/logout/update user

nkuehn:
A relatively simple first start to track our user interactions better into google analytics.   

Also addresses an issue with the new GA4 - we don't have traffic aggregation by path drilldown any more, so this is at least sending the microsite key as a value for the standard "content_group" feature which can then be used in the content reports. 

How would we test this normally?   The kit preview pages do not track (for good reason)

FYI / P.S.:  the content group _could_ alternatively be set "server side" in GA through parsing the path and setting the variable that way.  I'd nevertheless set it here to get better "out of the box" data, but we can still override / modify the field in the stream settings


timonrey (approved):
Nice, LGTM 👍🏼

packages/gatsby-theme-sso-ui-kit/src/utils/analytics.utils.ts
```diff
@@ -0,0 +1,87 @@
+// utility to track GA events client-side in a robust ways
+// lives in the SSO theme because typescript support is helpful in this case and the event tracking use cases are
+// (until the packages are merged anyways) all learning and authentication related
+// Recommended events reference: https://developers.google.com/tag-platform/gtagjs/reference/events
+
+export const gtagEvent = (
+  eventName: EventNames | (string & {}),
```
gabriele-ct, line 7: EventNames type doesn't include our own event names such as `submit_quiz`. I feel we should either add them to the `EventNames` type or create a `SelfLearningEventNames`.

nkuehn: good idea, done !

PR #1701, opened 2023-06-06
Ga svgr upgrade

gabriele-ct:


timonrey:
A changeset for the dependency changes in the packages would be good :)

PR #1711, opened 2023-06-09
Inline examples in generated raml files

### LINKED ISSUE #1349: Modifications to examples are not being cached properly
We've noticed that updates to content inside of examples are not being cached properly.

@nkuehn Input would be appreciated :)

timonrey:
closes #1349


packages/gatsby-transformer-raml/src/utils/resource/examples-to-array.mjs
```diff
@@ -1,16 +1,7 @@
...
-  const exampleAbsolutePath = path.resolve(fileNodeDir, filePath);
-  return fs.readFileSync(exampleAbsolutePath, 'utf8');
-}
-
-export function examplesToArray(examples, fileNodeDir, resolveExampleFile) {
+export function examplesToArray(examples) {
   if (examples) {
     return Object.entries(examples).map(([name, value]) => {
-      const jsonString = resolveExampleFile(fileNodeDir, value.value);
+      const jsonString = JSON.stringify(value.value, null, 2);
```
timonrey, line 4: The current indentation of the json examples is not really ideal yet. I'm open to suggestions how we can improve that. Ideally it should not change the indentation. It should simply use the same indentation as in the generated raml file.

nkuehn: @timonrey  the file based examples are formatted using prettier so you can use prettier to format the content if you want to ensure full consistency.     

I like moving to 2 space indentation to make it better readable. 

https://prettier.io/docs/en/api.html  and https://prettier.io/docs/en/options.html  -->  you can use prettier standards and only change the indentation depth if you want.

gabriele-ct:
This is an example that should render arrays on a single line:

```
const array = [1, 2, 3, 4, 5];
const jsonString = JSON.stringify(array, replacer, 2);

console.log(jsonString);

function replacer(key, value) {
  // Check if the value is an array
  if (Array.isArray(value)) {
    return '[' + value.join(',') + ']';
  }

  // For all other values, use the default behavior
  return value;
}
```

however it would be much better if we could not convert example string -> JSON -> string
I need to look a bit better into it

nkuehn:
> however it would be much better if we could not convert example string -> JSON -> string

If you imagine a future where there can also be examples that are not json but e.g. SDK code examples we could gradually switch to treating the examples as an opaque string with some meta-information about the language and reformat them through prettier for guaranteed consistency before rendering.

PR #1715, opened 2023-06-12
fix: render enums regardless of the discriminatorValue attribute value

gabriele-ct:


timonrey (approved):
Always nice to fix things by removing unnecessary code :)

PR #1724, opened 2023-06-20
Fix image caching issue

### LINKED ISSUE #1542: Moving an image unchanged  causes cached builds to fail
Moving an image file inside the images folder of the content to a different location (can only be a rename) without changing its contents causes cached builds to fail.  
Likely has to do with the fact that images are referred to by content hash in the gatsby cache only?

timonrey:
closes #1542 

I wasn't able to reproduce the issue in this gatsby version. Tested all three scenarios in the commit and the cached build never threw an error.


websites/docs-smoke-test/src/releases/excerpt-rendering-test.mdx
```diff
@@ -11,15 +11,15 @@ topics:
 
 <Subtitle>Subtitle Text</Subtitle>
 
-A plain paragraph `with inline code` in it. And **bold** and *italic*.
```
gabriele-ct, line 14: I'm wondering if you had to make these markdown changes for any specific reason. If these are "mandatory" changes because of some compatibility issue, we should keep that in mind in case some content in the docs repo needs to be updated

timonrey: Oh no that was just prettier I think...nothing I did on purpose. But it didn't change anything according to the VRTs

gabriele-ct: Ok, I'm not sure to be honest if we should include this change. The smoke test is supposed to cover specific usage of mardown that is widely used in the docs repo. What I'd like to avoid is to switch to a more "pretty" markup and not test corner cases that in the future could bite us back.
What do you think?

timonrey: Yes, let's keep this PR clean...I will remove these changes 👍🏼

gabriele-ct (approved):
Thank you 👍

PR #1722, opened 2023-06-16
Ga e2e page ready

gabriele-ct:

PR #1733, opened 2023-06-27
feat: wrap auth0 into custom hook for session initialization

gabriele-ct:


timonrey (approved):
LGTM 👍🏼

packages/gatsby-theme-docs/src/modules/sso/hooks/use-authentication.ts
```diff
@@ -0,0 +1,46 @@
...
+  const { isAuthenticated: isAuth0Authenticated, ...rest } = useAuth0();
+  const [isAuthenticated, setIsAuthenticated] = useState(false);
+  const { learnApiBaseUrl, selfLearningFeatures } = useContext(ConfigContext);
+  const { getAuthToken } = useAuthToken();
+
+  const shouldFetchData =
+    isFeatureEnabled(EFeatureFlag.CourseStatus, selfLearningFeatures) &&
+    isAuth0Authenticated;
+
+  const apiEndpoint = `/api/courses`;
```
nkuehn, line 21: Was `/api/courses` chosen to be the one blocking sync call because it's relatively likely that it will be called anyways?  Or because it's not very heavy on the backend (which I don't know)? Or because the data is changing frequently anyways so the info would need to be updated anyways on every other call we do? 

It would be good to document the choice in the code, it's not very self-explanatory and the implication of revalidating it every three minutes for every user could be significant (the hook is called very often so you can assume this code running more or less nonstop).   Originally, it felt more natural to me to use the user profile as the default info we make available before anything else happens  but from a UI loading perspective the courses list may be a good choice, too.

packages/gatsby-theme-docs/src/modules/sso/hooks/use-authentication.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const { data, error, isLoading } = useSWR(
+    shouldFetchData ? apiEndpoint : null,
+    (url) => fetcherWithToken(url, getAuthToken, learnApiBaseUrl, 'GET'),
+    {
+      revalidateOnReconnect: false,
+      revalidateIfStale: false,
+      revalidateOnMount: true,
+      dedupingInterval: 3 * 60 * 1000, // allow revalidation each 3 minutes
+      revalidateOnFocus: true,
```
nkuehn, line 31: Maybe document these choices in terms of what you want to achieve?  I assume they are specific to the situation inside the blocking hook and a tradeoff that tries to catch situations in which a user has no moodle token yet but is already logged in.   

by the way on that occasion: Did you consider making the logic more explicit in being conditional on whether a proxy API moodle token already exists?    The `user_session` cookie the API sets is HTTP only so it can't be read from JS but a common pattern seems to be that the API sets (and removes) another cookie that is JS readable but empty and only indicates the existence of the session but not the contents.

packages/gatsby-theme-docs/src/modules/sso/hooks/use-authentication.ts
gabriele-ct: > Was `/api/courses` chosen to be the one blocking sync call because it's relatively likely that it will be called anyways? Or because it's not very heavy on the backend (which I don't know)? Or because the data is changing frequently anyways so the info would need to be updated anyways on every other call we do?

Various reasons. Firstly that endpoint doesn't get any parameters, so it can be called straight away without any additional logic. Another reason is that it's a sort of _staple_ endpoint for our use cases: you always need to know at least in which courses you're enrolled in before doing anything else.

> 
> It would be good to document the choice in the code, it's not very self-explanatory and the implication of revalidating it every three minutes for every user could be significant (the hook is called very often so you can assume this code running more or less nonstop). 

document it in the code: yes. for the rest, I'm not sure to understand. The call goes through swr which is ensuring no actual call is sent to the server as long as there's a valid cache. And for the additional flags. These are the standard set of flags we agreed (after discussing and testing it) to use in our application. In details:
```
revalidateOnReconnect: false; you suggested to leave it disabled as it's not luckly to happen as we don't have many mobile users
revalidateIfStale: false; it must be left false otherwise it get's triggered at any user navigation (due to "gatsby SPA" limitation that unmounts and remounts all the components at each navigation)
revalidateOnMount: true; this must left true for the initial load to happen (it doesn't trigger if a cache is present)
dedupingInterval: 3 minutes; we decided to add this arbitrary timeout as fallback. If not set or set too high, the only way we have to refresh the cache is to manually `mutate` it. If in any corner case scenario we don't refresh the cache, at least it gets refreshed after 3 mins.
```

Originally, it felt more natural to me to use the user profile as the default info we make available before anything else happens but from a UI loading perspective the courses list may be a good choice, too.

not sure what "profile" endpoint you're referring to. These are the currently implemented endpoints our API exposes:
Screenshot 2023-06-29 at 12 49 36

packages/gatsby-theme-docs/src/modules/sso/hooks/use-authentication.ts
gabriele-ct: > Maybe document these choices in terms of what you want to achieve? I assume they are specific to the situation inside the blocking hook and a tradeoff that tries to catch situations in which a user has no moodle token yet but is already logged in.

See above, I can add the comments inline in the code so it stays documented.

> 
> by the way on that occasion: Did you consider making the logic more explicit in being conditional on whether a proxy API moodle token already exists? The `user_session` cookie the API sets is HTTP only so it can't be read from JS but a common pattern seems to be that the API sets (and removes) another cookie that is JS readable but empty and only indicates the existence of the session but not the contents.

What if the cookie exists but it has another user information? Shall I perform a sort of client side pre-validation? It feels un-necessary, at least until we see a real performance problem.

nkuehn: > What if the cookie exists but it has another user information? Shall I perform a sort of client side pre-validation? It feels un-necessary, at least until we see a real performance problem. 

The idea was meant only for the case you think it's worth the effort at all.  But it would not have to handle edge cases because the purpose is performance, not precision  (there are other scenarios remaining where parallel requests can have no session yet, e.g. when the moodle token expired in the middle of a session).   

I don't think explicitly handling whether there is a session or not is giving a relevant server load gain because the SWR based approach would nevertheless "ping" the courses list every few minutes.  The advantage would be that the implementation would make the purpose of why it exists very explicit which may be good in the long run.  

What I'm not sure about is the impact on the visual loading experience though if it's done without SWR caching across page loads and without checking the session.   For example, navigating to a page with login / logout specific UI elements would have the log in / out specific elements have to "wait" until the courses are loaded even if there is no learning specific information on that page.    I can't really tell you exactly, but my gut feeling says that the feature would be cleaner if it explicitly knew whether there's a session and would then explicitly do nothing but calling a specialized API that does nothing but initializing the session if there is none yet (no SWR etc, directly calling it).

packages/gatsby-theme-docs/src/modules/sso/hooks/use-authentication.ts
nkuehn: > not sure what "profile" endpoint you're referring to. 

`POST /users/{user-id}` or a newly to be created `GET /users/{user-id}`.    Alternatively a newly to be created `POST /init-session`

gabriele-ct: I think that calling `POST /users/{user-id}` which is an endpoint dedicated to update the user profile is even more misleading than calling the `/courses`. Let's put it like this: we need a new endpoint dedicated to session creation. Therefore I'd tag this to ticket as on hold until we have that endpoint. (I can work on that endpoint now, it should not be too hard to implement)

gabriele-ct: > Sorry, I did not remember them. Maybe we should expose them somewhere centrally as a constant and document them once there (separately from this). It would make clear that it's a standard config.

True, I've done that in the latest commit 👍

gabriele-ct:
@FFawzy and @nkuehn 
Having deployed the init-session endpoint to prod I've been able to test this branch which looks to perform much better in terms of e2e failures. This is encouraging and I'd ask you both to have a final look to the PR before we merge it. 
P.S. you'll notice a bunch of e2e changes, these are a port of [this](https://github.com/commercetools/commercetools-docs-kit/pull/1722) PR which will be eventually merged.

Thank you

packages/gatsby-theme-docs/src/modules/sso/hooks/use-authentication.ts
```diff
@@ -0,0 +1,70 @@
...
+  );
+
+  useEffect(() => {
+    const checkUserSessionCookie = () => {
+      const cookieName = 'user_session_for_client';
+      const cookieValue = getCookieValue(cookieName);
+      if (!doesSessionExist(cookieValue)) {
+        setShouldInitSession(true);
+      } else {
+        setIsAuthenticated(true);
```
nkuehn, line 50: it may be a nitpick or I am not fully understanding the logic here.  The fact that a session from the learn API exists does not imply that the user is also authenticated auth0-side.   We try to clear the session when logging out but Auth0 can also log out or expire a session in other ways.  

or is this just the wrapper logic that "delays" the signal to the rest of the app that the user is authenticated until the moodle API session is there, too?

nkuehn (commented):
Just one question I felt unsure about.  I will not find the time to thoroughly test the error cases like "not logged in any more from another tab" or session expiration realistically.

packages/gatsby-theme-docs/src/modules/sso/hooks/use-authentication.ts
gabriele-ct: Hi @nkuehn , 
I think you might be mislead by the function closure. `checkUserSessionCookie` gets invoked only if `isAuth0Authenticated === true` (see below). The `useEffect` is triggered only on `isAuth0Authenticated` change AND, only when it's true, the check session happens.

PR #1735, opened 2023-06-28
fix: strip pathPrefix only if it's matched exactly in the href

gabriele-ct:


packages/gatsby-theme-docs/src/components/link.js
```diff
@@ -17,7 +17,7 @@ const dummyHostname = 'dummy.com';
 // In order to resolve relative paths that do no start with a `/`, we need to first
 // trim the possible `pathPrefix` included in the `location` of `@react/router`.
 const withoutPrefix = (value, pathPrefix) =>
-  value.replace(new RegExp(`^${pathPrefix}`), '');
+  value.replace(new RegExp(`^${pathPrefix}\\b`), '');
```
gabriele-ct, line 20: pathPrefix is now matched as an exact word meaning that partial matches are not stripped anymore.
This mitigate the original issue, however it's still not possible to create a page with the same name as the pathPrefix, this rule should be kept in mind by editors team

timonrey (approved):

PR #1737, opened 2023-07-04
Support new feeds with only one release note in rss-feed component

### LINKED ISSUE #1429: Release Notes on the homepage merging feeds from different microsites
Only the latest four release notes are currently displayed on the [homepage](https://docs.commercetools.com/docs/). Currently only coFe release notes are displayed. This makes it hard to access the CoCo release notes. 
Please show the usual long list again.

timonrey:
closes #1429 

Preview link:
https://docskit-tt-fix-rss-feed-with-brandnew-feeds.commercetools.vercel.app/docs-smoke-test/components/rss-feeds


packages/ui-kit/src/components/rss/rss-feeds.tsx
```diff
@@ -101,6 +102,7 @@ const transformData = (data: FlatRssEntry[][]) => {
   const tableData: FlatRssEntry[] = data
     .flat()
     .reduce<FlatRssEntry[]>((list, entry) => {
+      if (!lastEntryOfList.pubDate) return [...list, entry];
```
timonrey, line 105: In case we don't have feeds with enough entries (like in the smoke-test atm). Not really important for the docs use-case.

websites/docs-smoke-test/src/content/components/rss-feeds.mdx
```diff
@@ -11,12 +11,13 @@ import { RssFeeds } from '@commercetools-docs/ui-kit'
...
 ## Table with multiple feeds
 
 <RssFeeds dataSources={[
-    "/docs-smoke-test/releases/feed.xml",
-    "/api-docs-smoke-test/releases/feed.xml"
-]}/>
\ No newline at end of file
+"/docs-smoke-test/releases/feed.xml",
+"/api-docs-smoke-test/releases/feed.xml",
+"/self-learning-smoke-test/releases/feed.xml"
```
timonrey, line 22: Adding a new feed with only one release note (with a new date) to create the use case in the smoke test.

packages/ui-kit/src/components/rss/rss-feeds.tsx
```diff
@@ -72,11 +72,12 @@ const fetcher = async (args: string[]) => {
 };
 
 const transformData = (data: FlatRssEntry[][]) => {
-  // First, we need to get the oldest release note from each feed,
+  // First, we need to get the oldest release note from feeds with at least 10 entries,
```
nkuehn, line 75: Please point to the place in the code where the 10 is defined and make a strong statement that it must by synced if that changes.   Same in the other direction in the place where the 10 is defined.   
I think we can't make it a central constant because it's in ui-kit vs. docs-kit but a comment should be good enough.

nkuehn (commented):
Nice, was really as easy as hoped for.  
Just a documentation wish

packages/ui-kit/src/components/rss/rss-feeds.tsx
timonrey: That is the only place related to the release notes with a hardwired 10 I was able to find...but I guess that's not the place you had in mind?
https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/gatsby-theme-docs/src/components/release-notes-filter-dates.js#L56

nkuehn: Hier ist der feed kodiert mit 10 hardwired:  https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/gatsby-theme-docs/gatsby-config.mjs#L298  

Die 10 könntest du sogar als template reinnehmen aber wie gesagt wäre das nervig weil ui-kit<->docs-kit abhängigkeit.

timonrey: Ah in der query, verstehe 👍🏼 dann verlinke ich die beiden Stellen noch.

PR #1744, opened 2023-07-07
feat: clickable chapter title

gabriele-ct:
The change enable the support for clickable chapter titles. We still support non-clickable chapters. No feature flags are needed.
Once the docs migration completes, we should refactor the code to remove backwards compatibility leftovers


gabriele-ct:
hi @zbalek, if you want to have a look to what I've implemented, just to give a UI validation, please use this link https://docskit-ga-clickable-chapters.commercetools.vercel.app/documentation/
Thank you

zbalek:
@gabriele-ct nice - looks good!

nkuehn:
@gabriele-ct  would it fail if a certain chapter lacks a link or what's the behavior?   I think it would be good if the smoke test is not only covering the good case but also bad cases like no URL being set or other cases that should normally not happen.

gabriele-ct:
> @gabriele-ct would it fail if a certain chapter lacks a link or what's the behavior? I think it would be good if the smoke test is not only covering the good case but also bad cases like no URL being set or other cases that should normally not happen.

https://github.com/commercetools/commercetools-docs-kit/pull/1744#issue-1793504924

Currently it doesn't fail, it fallsback to the original behaviour

gabriele-ct:
I will describe to the editors in details tomorrow in the daily but the idea is that they are going to have the maximum flexibility, they will be able per chapter to decide if the title is clickable or not (by adding the `path` attribute).
Eventually, if I understand it correctly, all the chapters will be clickable, but in the current implementation there's no hard requirement per site or any breaking change introduced. I think this is the smoothest path possible for editors

timonrey (approved):
Thank you for the effort making it robust for clickable and non-clickable chapter titles!
Code looks good 👍🏼

PR #1752, opened 2023-07-11
[Design adjustments] Enable new `Inter` font

### LINKED ISSUE #1749: [Design adjustments] Apply Inter V font
We decided to change the font of the docs website to Inter V. We also want to adjust the font to our needs through the font features listed here: https://rsms.me/inter/lab

Suggested font adjustments: https://github.com/commercetools/commercetools-docs-kit/pull/1727#discussion_r1237229163
Link to the test PR: https://github.com/commercetools/commercetools-docs-kit/pull/1727
Link to the Figma board: https://www.figma.com/file/GqhHFUcEaMxBXnMM9TXdaZ/%F0%9F%9F%A2-Feedback-Board?type=design&node-id=69-2615&mode=design&t=q9cpubSWtenhRzPD-0

timonrey:
closes #1749 

Hi @zbalek 
As discussed already, this is my proposal for the font and the feature settings we can use for the font. If you still want to check out other feature settings, here is the link:
http://rsms.me/inter/lab/


packages/gatsby-theme-docs/globals.css
```diff
@@ -1,6 +1,8 @@
 @import '@commercetools-uikit/design-system/materials/custom-properties.css';
 
 :root {
-  --font-family-default: 'Roboto', sans-serif;
+  --font-family-default: 'Inter', sans-serif;
+  --font-feature-settings: 'ss02';
```
timonrey, line 5: Setting the feature setting globally to ensure that it's applied everywhere.

packages/gatsby-theme-docs/package.json
```diff
@@ -43,8 +43,8 @@
     "@emotion/react": "^11.10.5",
     "@emotion/server": "^11.10.0",
     "@emotion/styled": "^11.10.5",
-    "@fontsource/roboto": "5.0.2",
-    "@fontsource/roboto-mono": "5.0.2",
+    "@fontsource/inter": "^5.0.5",
+    "@fontsource/roboto-mono": "^5.0.5",
```
gabriele-ct, line 47: I'd keep the version locked, don't forget the font might be used from `docs` repo too and we should enforce the same version for both repos to avoid bundling the font twice

packages/ui-kit/src/components/markdown.tsx
```diff
@@ -255,7 +256,8 @@ const InlineCode = styled.code`
   border: 1px solid ${colors.light.borderSecondary};
   border-radius: ${dimensions.spacings.xs};
   color: ${colors.light.textCode};
-  font-family: ${typography.fontFamilies.code};
+  font-family: ${typography.fontFamilies.nonMonospaceCode};
```
gabriele-ct, line 259: is it ok to use a non monospace font for code? If I understand correctly it's only for inline code. 
Also, we still use `roboto-mono` for code snipplets right? Shouldn't we use the same monospace font for all code (being it inline or not)?

gabriele-ct (commented):
Overall looks good, I just want to clarify a couple of points, see comments

packages/ui-kit/src/components/markdown.tsx
nkuehn: I think the name "code" should be refactored in this case.  The use case are inline string literals, i.e. nothing where you have multiple indented lines or lists that need to be aligned.   So it can use a non-monospace font that has good letter disambiguation but does not have to be strictly monospace.

packages/gatsby-theme-docs/globals.css
nkuehn: This is not correct, sorry.   Disambiguation needs to be applied selectively to string literals and other places where it is important that the exact sequence of characters is seen.   In general body and UI  text you should use the font as designed by the font designer. 

Same for the line below,  you want to let the font do its letter spacing magic and only apply it where you want a bit wider spacing for disambiguation.  Except if @zbalek  really wants custom letter spacing everywhere including body text.

packages/ui-kit/src/components/markdown.tsx
```diff
@@ -255,7 +256,8 @@ const InlineCode = styled.code`
...
+  font-feature-settings: 'salt';
```
nkuehn, line 260: at least my opinion:  apply `ss02` and letter-spacing=0 here vs globally.   But Zoe's view wins if she thinks we should have wider spacing or disambiguation everywhere .

timonrey: I think @zbalek should have a final call on it, because I'm not sure anymore which approach she wants to do here.
Either we have a consistent font configuration for all text components or we have a special one for string literals and other places where it makes sense as Nikolaus suggested.

nkuehn (changes requested):
Seems to not have been tested at all.   Disambiguation is not happening in the string literal (and no test case content was added for the topic. 

image

Guessed root cause:  re-declaring the font feature settings property instead of setting it with all values that should be applied in the context.  CSS is not cumulative on properties that allow lists of values

Visuall I would allow even a little more letter spacing on the string literals, e.g. 0.03em or something like that.  It would make it a bit more monospace-ish (just a personal preference when trying to read things letter by letter)

gabriele-ct (approved):
code looks 👍

packages/gatsby-theme-docs/gatsby-browser.js
```diff
@@ -13,9 +13,9 @@ import { SWRConfig } from 'swr';
 import { docsCache } from './utils/create-emotion-cache';
 import { PortalsContainer } from '@commercetools-docs/ui-kit';
 import * as Sentry from '@sentry/browser';
-import '@fontsource/roboto/latin-400.css';
-import '@fontsource/roboto/latin-500.css';
-import '@fontsource/roboto/latin-700.css';
+import '@fontsource/inter/latin-400.css';
+import '@fontsource/inter/latin-500.css';
+import '@fontsource/inter/latin-700.css';
```
nkuehn, line 18: maybe try  `latin-ext-400.css` for the disambiguation issue.   (also in the 500 and 700 weights if needed there)

nkuehn (commented):
Sorry, still a bunch of details now that I did a full review.  But overall I really like the new look. 

----

_Smoke test  still refers to the inline code as "monospace"_ 

image

----

_Applying disambiguation to the constants component could be a good choice (at least for the numbers):_ 

image

----

_Disambiguation still does not work although there is a test now - adding a test only makes sense if we look at it, too. My manual testing led to the conclusing that "tnum" is available in the font we load but the other feature settings seem to have no effect. I have two ideas what the problem could be.  1) we have to load the "ext" not the "normal" font file. ext has more characters included.  2) we have to load "Inter Variable" and not "Inter" (both available on fontsource NPM.  The inter lab tool uses the variable width font. _

image

----

I think it looks inconsistent that the OAuth scopes and mime types in the API docs use the monospace for inline string literals vs.  the other inline string literals now being disambiguated  normal fonts. 

image

@zbalek  what do you think ?  e.g.  here https://docskit-tt-design-adjustments-apply-inter-font.commercetools.vercel.app/api-docs-smoke-test/layouts/endpoint-large-desktop  

----

(could have seen that earlier):  The path of an API is an inline string literal, too.  We could apply the same settings (disambiguation, letter spacing) here, too now that we have that possibility in the typesetting tech. 

image

packages/gatsby-theme-docs/gatsby-browser.js
```diff
@@ -13,9 +13,7 @@ import { SWRConfig } from 'swr';
 import { docsCache } from './utils/create-emotion-cache';
 import { PortalsContainer } from '@commercetools-docs/ui-kit';
 import * as Sentry from '@sentry/browser';
-import '@fontsource/roboto/latin-400.css';
-import '@fontsource/roboto/latin-500.css';
-import '@fontsource/roboto/latin-700.css';
+import '@fontsource-variable/inter';
```
timonrey, line 16: @nkuehn I know that the font feature settings still don't have an effect on the string literals. I was also looking for a fix. I tried out the `fontsource-variable` package now, but the effect is still the same.

packages/gatsby-theme-docs/gatsby-browser.js
timonrey: Didn't work unfortunately.

packages/ui-kit/src/components/globals.tsx
```diff
@@ -81,8 +81,9 @@ const Globals = (props: TGlobalsProps) => (
       kbd,
       samp,
       pre {
-        font-family: ${typography.fontFamilies.code}, 'Menlo', 'Monaco',
+        font-family: ${typography.fontFamilies.nonMonospace}, 'Menlo', 'Monaco',
           'Consolas', 'Liberation Mono', 'Courier New', monospace;
+        font-feature-settings: 'salt', 'ss02';
         font-size: ${typography.fontSizes.small};
```
timonrey, line 87: We could also think about adjusting the font-size here...the mime types look a bit small to me.

packages/gatsby-theme-docs/src/fonts/Inter/inter.css
```diff
@@ -0,0 +1,178 @@
...
+  font-family: 'Inter';
+  font-style:  italic;
+  font-weight: 900;
+  font-display: swap;
+  src: url("Inter-BlackItalic.woff2?v=3.19") format("woff2"),
+       url("Inter-BlackItalic.woff?v=3.19") format("woff");
+}
+
+/* -------------------------------------------------------
+Variable font.
```
timonrey, line 155: The font features only work if we use the variable font, but together with the defined font faces above, or if we use Inter directly...in this case, we can remove the variable fonts here. Still thinking about the package size and how to maybe optimise it.

packages/ui-kit/src/design-system.ts
```diff
@@ -288,8 +288,9 @@ export const dimensions = {
 
 export const typography = {
   fontFamilies: {
-    primary: "'Roboto', sans-serif",
+    primary: "'Inter var', sans-serif",
     code: "'Roboto Mono', monospace",
+    nonMonospace: "'Inter var'",
```
nkuehn, line 293: to be honest I find the name of the token not ideal.  It's not actually a statement about what it is, it's just excluding monospace but leaving anything else undescribed

packages/gatsby-theme-docs/globals.css
```diff
@@ -1,6 +1,7 @@
 @import '@commercetools-uikit/design-system/materials/custom-properties.css';
+@import url('./src/fonts/Inter/inter.css');
```
nkuehn, line 2: Why did you choose to move this into globals.css but leave the roboto mono import in gatsby-browser?   I can understand that globals could actually be the better place but why not move roboto mono here, too?   -> consistent would be better in any case. 
Also add a comment explaining why the fonts are handled differently to prevent a later ambitious maintainer from "cleaning up" and making it fontsource based again.  Thanks !

packages/gatsby-theme-docs/src/fonts/Inter/inter.css
nkuehn: Yes, I suggest to make it variable-only and include a  README in the inter folder explaining how to update it and where to get it from.   (or, for that purpose, a comment in the remaining CSS piece).   

The version number in the url can be removed in our stack, the build process is hashing it anyways.  

Also, I would remove the unused font files to not bloat the repository and confuse people.

packages/ui-kit/src/design-system.ts
timonrey: Any suggestions?

packages/ui-kit/src/design-system.ts
```diff
@@ -288,8 +288,9 @@ export const dimensions = {
 
 export const typography = {
   fontFamilies: {
-    primary: "'Roboto', sans-serif",
+    primary: "'Inter var', sans-serif",
     code: "'Roboto Mono', monospace",
+    stringLiteral: "'Inter var'",
```
nkuehn, line 293: you may choose monospace as the generic family type fallback here.  In doubt that matches the use case better.

packages/ui-kit/src/components/globals.tsx
nkuehn: obsolete after the new component for string literals without a box is there.

nkuehn (approved):
Thank you!  Fine from the "craft" side of things for me. 

Final approval needs to come from @zbalek to make sure the effective overall look is OK now.  Esp. the API docs could deserve a second look.

zbalek:
Completed the final checks with Timon's help - everything looks good - let's merge🎉

zbalek (approved):

PR #1755, opened 2023-07-13
[Design adjustments] Reduce spacing between sections in the content page

### LINKED ISSUE #1748: [Design adjustments] Adjust spacing between sections
Adjust the current spacings between the mdx content sections according to the Figma board.

Link to Figma board:
https://www.figma.com/file/GqhHFUcEaMxBXnMM9TXdaZ/%F0%9F%9F%A2-Feedback-Board?type=design&node-id=69-2615&mode=design&t=q9cpubSWtenhRzPD-0

timonrey:
closes #1748 

Hi @zbalek 
This is how the content page looks like if we reduce the spacing between the sections to 8px as suggested in the Figma board. I personally think that it looks a bit too squeezed now. Maybe 16px between the h3 sections would be better, but that's just my opinion :)

### DIFF
--- a/.changeset/tough-walls-laugh.md
+++ b/.changeset/tough-walls-laugh.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/ui-kit': patch
+---
+
+Adjust spacing between sections in the page content.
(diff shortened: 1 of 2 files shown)


zbalek:
@timonrey yep, it definitely is - let's try to go with 16, potentially even 24px

gabriele-ct (approved):

PR #1767, opened 2023-07-20
chore: revert ssr bundling hack

gabriele-ct:


nkuehn (approved):
trusting your testing.  You may want to refer to the gatsby version that supposedly fixed it in the merge commit (or the ticket where it's linked) so that later we can identify better what broke stuff or not.

PR #1777, opened 2023-07-26
B2B prominence and PlanTag component

gabriele-ct:


gabriele-ct:
> * is the regression in the beta flag height intentional ( see VRTs)?
The whole component structure changed so, in order to keep consistent the height for tag with or without icons, the beta tag has changed
> * It would be very very good if the new feature were covered by the VRTs.  It's the kind of stuff that is really easy to accidentally resize / move / shift a few pixels through some other change
It will be, I guess we're still not at a stage of adding VRT coverage (see discussion in the ticket)

gabriele-ct:
@nkuehn maybe you can point me in the right direction here.
I've noticed that the SSR version of the page, when using inline `<Beta />` or `<B2B />` tags differs from the hydrated one. See [here](https://docskit-ga-b2b-prominence.commercetools.vercel.app/docs-smoke-test/views/markdown)
I'm unsure about the actual cause since I believe styles should be calculated and applied server side, so it shouldn't be related to it.
What I see is that the actual markup differs like there was some mismatch in the way the actual html markup gets generated when SSR.
Do you have any idea?

packages/gatsby-theme-docs/src/components/b2b-tag.js
```diff
@@ -0,0 +1,27 @@
+import PropTypes from 'prop-types';
+import { PlanTag } from './plan-tag';
+import { Icons } from '@commercetools-docs/ui-kit';
+
+/**
+ * This component stays as shorthand of the <PlanTag /> component. The configuration
+ * is hardcoded to display B2B tags.
+ */
+export const B2B_HOVER_TEXT = 'TBD';
```
gabriele-ct, line 9: the B2B hover text needs to be provided. @melindafekete do you have some content to add here?
We're talking about the text that the user will see when hovering over a B2B tag.
For beta tag the text is: _This feature is marked as beta and is subject to change. Use with caution_.

melindafekete: @nirusuu do you have any suggestion for what we should display for the B2B tag hover text? thanks!

gabriele-ct:
Hi @zbalek,
when you have time could you please have a final look and review to the B2B tag?
Here's a list of pages worth to check:
https://docskit-ga-b2b-prominence.commercetools.vercel.app/docs-smoke-test/views/b2b
https://docskit-ga-b2b-prominence.commercetools.vercel.app/docs-smoke-test/views/beta (beta tag has been refactored as part of b2b implementation)
https://docskit-ga-b2b-prominence.commercetools.vercel.app/api-docs-smoke-test/types/general (again, beta tag)

Thank you

packages/gatsby-theme-docs/src/components/b2b-tag.js
nirusuu: I suggest we reuse the same copypasta we have used before in the docs: 
> This feature is only available as part of Composable Commerce for B2B and will be subject to additional terms and pricing.

nkuehn: @gabriele-ct  I'm a bit confused.  We chose to implement a generic "plan tag" that can then be configured in the content repo with what plans with what descriptions are available. This should not happen here because this is open source, we would not be able to launch any business model changes without silently going public here.

gabriele-ct: > I'm a bit confused. We chose to implement a generic "plan tag" that can then be configured in the content repo with what plans with what descriptions are available. This should not happen here because this is open source, we would not be able to launch any business model changes without silently going public here.

@nkuehn 
`<Beta>` and `<B2B>` components are shorthands wrapping the PlanTag component. We don't want the editors to configure the same Beta or B2B tag over an over again. If hover text will be the same, as it is for Beta, I think this is the pattern we should follow

nkuehn: "over and over again" is putting the configuration once into the base theme in the docs repo.  One place to put it.

zbalek:
Hi @gabriele-ct - thanks for attaching the links! 

Re: https://docskit-ga-b2b-prominence.commercetools.vercel.app/docs-smoke-test/views/beta
- at the moment we only show Beta tag at the top of the page and right navbar.  I didn't realise that by adding a B2B text next to a heading we will also add a BETA tag there - is this actually correct?

packages/gatsby-theme-docs/src/components/b2b-tag.js
gabriele-ct: thank you @nkuehn for clarifying this additional requirement. Do we want this pattern to be implemented for Beta tag too?
Meaning that the siteMetadata will look like this:
```
siteMetadata: {
    betaHoverText: 'This feature is marked as beta and is subject to change. Use with caution.'
    betaLink: '/../offering/api-contract#public-beta',
    b2bHoverText: 'This feature is only available as part of Composable Commerce for B2B and will be subject to additional terms and pricing.'
    b2bLink: 'some link to the beta page'
  },
```

packages/gatsby-theme-docs/src/overrides/plan-tags-config.js
```diff
@@ -0,0 +1,27 @@
+import { Icons } from '@commercetools-docs/ui-kit';
+
+const planTagsConfig = {
```
nkuehn, line 3: nice, this is easy to edit even for not-as-technical authors!

websites/docs-smoke-test/src/content/plan.mdx
```diff
@@ -0,0 +1,7 @@
+---
+title: Plan Legal Agreements
+---
+
+# Terms and conditions
+
+TBD
```
nkuehn, line 7: I'm not sure what test case this page represents.  "TBD" souhds like it's not ready or a leftover

nkuehn (approved):
Thanks for including documentation!  I have not done an in-detail review, just took a look at the docs and how it's configurable / extensible.

websites/docs-smoke-test/src/content/plan.mdx
gabriele-ct: The idea here is to have a page similar to https://docs-kit.commercetools.vercel.app/docs-smoke-test/beta . It's used as sort of  dummy landing page for terms and conditions of the dummy plans defined in the smoke test site. Essentially when the user clicks on the `plan1` or `plan2` icons at the top of the pages, he will land on this page. TBD could be replaced with lorem ipsum or maybe I can just rephrase what I wrote here so that is self explanatory

PR #1780, opened 2023-08-01
feat: lordicon wrapper component

gabriele-ct:


packages/ui-kit/package.json
```diff
@@ -17,7 +17,8 @@
   "module": "dist/commercetools-docs-ui-kit.esm.js",
   "files": ["dist", "package.json", "LICENSE", "README.md"],
   "scripts": {
-    "generate-icons": "svgr --out-dir src/icons/generated --typescript -- src/icons/svg"
+    "generate-icons": "svgr --out-dir src/icons/generated --typescript --icon \"src/icons/svg/*.svg\" \"src/icons/lord-icon/*.svg\" && yarn copy-lord-icons-json",
+    "copy-lord-icons-json": "cpy --cwd=src 'icons/lord-icon/*.json' ../dist/icons/lord-icon"
```
nkuehn, line 21: Is the custom CLI used for windows compatibility?  I would be OK having the kit not support windows based developmemt without a Unix shell.  We have quite a lot of unix shell scripting, even in the docs repo.

packages/ui-kit/src/components/lord-icon-client-side.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const LordIcon = (props: LordIconProps) => {
+  const [animationData, setAnimationData] = useState();
+  const lottieRef = useRef<LottieRefCurrentProps>(null);
+  const { iconName, ...rest } = props;
+
+  useEffect(() => {
+    const loadAnimationData = async () => {
+      try {
+        const data = await import(
```
nkuehn, line 24: Did you consider importing and holding the fixed set of icons that we know are used anyways statically in the module?    At least for the top nav content we do know that every single website will need and load it anyways.   Could keep the dynamic import for other icons that are in the distribution.     Basically like you do it for the preview SVG, they are statically in the server side build and fall back to nothing. 

(Maybe that would resolve the flicker, too? )

packages/ui-kit/src/components/lord-icon-client-side.tsx
```diff
@@ -0,0 +1,55 @@
...
+      }
+    };
+    loadAnimationData();
+  }, [iconName]);
+
+  const handleMouseEnter = () => {
+    lottieRef.current?.play();
+  };
+
+  const handleMouseLeave = () => {
```
nkuehn, line 39: TBH I'm not conviced the animation should abruptly stop when leaving the mouse.  Normally they continue the initialized loop.  
The lordicon custom element has the triggers preconfigured, there's "hover",  "loop-on-hover", "loop", "morph" etc.  ( https://github.com/tomwilusz/lord-icon-element/blob/master/examples/triggers.html ) 

Their "hover" trigger only acts on mouse enter ( https://github.com/tomwilusz/lord-icon-element/blob/master/src/triggers/hover.ts ) 

Interaction with content and design people could also be easier if we use the lordicon custom element because the exact behavior of the triggers it provides are built into the lordicons website and they can try out there.  ( e.g. click "HTML" on one of the icons here: https://lordicon.com/icons/wired/outline )

packages/ui-kit/src/utils/lord-icon.ts
```diff
@@ -0,0 +1,10 @@
+import Suitcase from '../icons/generated/Suitcase';
+
+export const getStaticSvgComponent = (iconName: string) => {
+  switch (iconName) {
```
nkuehn, line 4: could be an object with props that are the name keys.  Would maybe have better Typing support by TS (the object would have an implicit type) and easier to maintain maybe.

nkuehn (commented):
+ documentation on how to add icons (how to exactly download them)

gabriele-ct:
@nkuehn I've integrated the `lord-icon-element` library as you recommended. The library itself seems to be working fine and, indeed, removes the issue with the little flickering.
However I'm finding some issue with the `svgr` icon generator. there are a couple of svg icons (namely login and document) that get converted to jsx with some vectors missing. if you see the icons here: https://docskit-sha-8c0d3b6340588fd24f2779b9efc89257702329b6.commercetools.vercel.app/docs-smoke-test/views/markdown

you'll notice that the static version of those 2 icons are not identical to the initial state of the lottie. I've tried several svgr options without much luck. Is there anything that you would recommend in this case? I'm running out of ideas. Maybe using the  tag instead but that would mean to host the svg somewhere.

.changeset/lazy-lemons-move.md
```diff
@@ -0,0 +1,7 @@
+---
+'@commercetools-docs/gatsby-theme-docs': minor
+'@commercetools-website/documentation': minor
+'@commercetools-docs/ui-kit': minor
+---
+
+Created a wrapper component for using lord-icon icons
```
timonrey, line 7: I see that there are two changesets here. Can you please remove one?

gabriele-ct: done. Thank you @timonrey

nkuehn (approved):
approving in the name of @timonrey  who is unavailable today but did the review already

PR #1790, opened 2023-08-03
Display content types for responses correctly

### LINKED ISSUE #1762: Content-type for request and response body should be treated independently
### Context
I have the following POST method specified in RAML that has an URL encoded body for the request, but a JSON body for the response:

```yaml
post:
  body:
    application/x-www-form-urlencoded:
      type: ProductSearchParameter
      (builtinType): object

  responses:
    200:
      body:
        application/json:
          type: ProductProjectionPagedSearchResponse
          (builtinType): object

```
### Current behavior

The Kit renders the content type for this endpoint correctly for the request as  `application/x-www-form-urlencoded`, but not for the response:
wrong response content type
### Expected behavior
The response should show `application/json` instead of  `application/x-www-form-urlencoded`.

timonrey:
closes #1762

### DIFF
--- a/packages/gatsby-theme-api-docs/src/components/resource/method/request-response-examples.js
+++ b/packages/gatsby-theme-api-docs/src/components/resource/method/request-response-examples.js
@@ -9,9 +9,9 @@ const RequestResponseExamples = (props) => {
   const apiTypes = useApiTypes();
   const responsesCodeExamples = [];
   if (props.responses) {
-    props.responses.forEach((response) => {
+    props.responses.forEach((response, index) => {
       const typeDisplayName =
-        response.body && response.body.applicationjson.type;
+        response.body && response.body[props.contentType[index]].type;
       const { code } = response;
       if (typeDisplayName) {
         const apiType = apiTypes.find((type) => {
@@ -20,7 +20,8 @@ const RequestResponseExamples = (props) => {
           );
         });
 
-        const examplesNode = response.body?.applicationjson?.examples;
+        const examplesNode =
+          response.body?.[props.contentType[index]]?.examples;
 
         if (examplesNode && Array.isArray(examplesNode)) {
           examplesNode.forEach((example) => {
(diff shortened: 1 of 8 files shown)


packages/gatsby-theme-api-docs/src/components/resource/method/request-response-examples.js
```diff
@@ -9,9 +9,9 @@ const RequestResponseExamples = (props) => {
   const apiTypes = useApiTypes();
   const responsesCodeExamples = [];
   if (props.responses) {
-    props.responses.forEach((response) => {
+    props.responses.forEach((response, index) => {
       const typeDisplayName =
-        response.body && response.body.applicationjson.type;
+        response.body && response.body[props.contentType[index]].type;
```
gabriele-ct, line 14: what if this expression is undefined? `body[props.contentType[index]]`
Also, you can use the [conditional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) instead of repeating `response.body && ....`

So it would become `response?.body?.[props.contentType[index]]?.type`

packages/gatsby-theme-api-docs/src/components/resource/method/request-response-examples.js
```diff
@@ -33,7 +34,7 @@ const RequestResponseExamples = (props) => {
               value: example.value,
             });
           });
-        } else if (apiType.examples && apiType.examples.length > 0) {
+        } else if (apiType?.examples && apiType?.examples.length > 0) {
```
gabriele-ct, line 37: here too, it could be simply `apiType?.examples?.length > 0`

packages/gatsby-theme-api-docs/src/components/resource/method/responses.js
```diff
@@ -48,16 +50,16 @@ const Responses = ({ apiKey, responses, contentType }) => {
                   <SpacingsInline alignItems="center">
                     {renderTypeAsLink(
                       apiKey,
-                      response.body.applicationjson.type,
+                      response.body[contentType[index]].type,
```
gabriele-ct, line 53: some conditional check here are probably also good to have

gabriele-ct (changes requested):
Just few safety check to be added

Issue #1797, opened 2023-08-14
CodeBlock markdown component not rendered correctly


gabriele-ct:
when using `markdownFragmentToReact` function overriding `pre` rendering with `CodeBlock` component. The rendering fails due to the fact that the content is expected to be passed as prop but the wrapper component doesn't seem to forward all the props down to the actual `CodeBlock` implementation.

The objective is to be able to use `markdownFragmentToReact` function to render multiline code blocks properly

nkuehn:
Solution is to use the `CodeBlockMarkdownWrapper`, not the pure `CodeBlock`

what's blocking it is that the UI Kit is only published in compiled minified form and that the markdown wrapper is not exported. 

Please patch release UI kit and export  https://github.com/commercetools/commercetools-docs-kit/blob/main/packages/ui-kit/src/components/multi-code-block.tsx#L223C14-L223C38  CodeBlockMarkdownWrapper  and also the Multi** variant of it.  

The wrapper can then be passed as the renderer for "pre" in markdown fragment renderer

nkuehn:
P.S:  as a default, it's intended that the markdown fragment renderer is not rendering full blocks, for example if embedded int the description field of an API field.    It shuold be a situational override to allow rendering "bigger" components

FFawzy:
the wrappers can be added as options here https://github.com/commercetools/commercetools-docs/blob/82ddc291924bd4e954a0acbfc7a2c3ca37a04ad6/websites/docs/src/components/chat-messages.jsx#L45

timonrey:
Solved by this PR:
https://github.com/commercetools/commercetools-docs/pull/3852

PR #1811, opened 2023-08-29
Export `CodeBlockMarkdownWrapper` component from the kit

timonrey:
part of #https://github.com/commercetools/commercetools-docs-kit/issues/1797

You can see the result here: https://docskit-tt-export-codeblockmarkdownwrapper-from-ui-kit.commercetools.vercel.app/docs-smoke-test/


packages/ui-kit/src/index.ts
```diff
@@ -11,6 +11,7 @@ export { default as LogoButton } from './components/logo-button';
 export { default as MediaQuery } from './components/media-query';
 export {
   default as MultiCodeBlock,
+  CodeBlockMarkdownWrapper,
   MultiCodeBlockMarkdownWrapper,
```
timonrey, line 15: Looks like only the markdown wrapper for single code blocks was not exported.

packages/gatsby-theme-docs/src/components/card.js
```diff
@@ -47,61 +48,83 @@ WrapWith.propTypes = {
...
+const age = 25;
+console.log('You are ' + age + ' years old.');
+\`\`\`
+
+This code declares a variable \`name\` with the value "John" and logs a greeting message to the console. It also declares a variable \`age\` with the value 25 and logs a message about the age.
+`;
+
+const TestMultiMarkdown = () => (
+  <div>
+    {markdownFragmentToReact(testValue, { pre: MultiCodeBlockMarkdownWrapper })}
```
timonrey, line 66: FYI I used the cards component to test the use case...once approved, I will revert all the changes in this file.

packages/ui-kit/src/components/multi-code-block.tsx
```diff
@@ -193,23 +193,25 @@ function MultiCodeBlock(props: MultiCodeBlockProps) {
...
   const [, languageCode] = languageToken.split('language-');
   const parsedOptions = parseCodeBlockOptions(childProps);
+
   const content =
-    childProps && childProps.children ? childProps.children : childProps;
+    childProps && childProps.children
+      ? Array.isArray(childProps.children)
+        ? childProps.children[0]
+        : childProps.children
+      : childProps;
```
FFawzy, line 214: I didn't go through the whole thing, but is there a better way than the fixed index here?

FFawzy (approved):
LGTM
just a small question in the comments

packages/ui-kit/src/components/multi-code-block.tsx
```diff
@@ -193,23 +193,25 @@ function MultiCodeBlock(props: MultiCodeBlockProps) {
...
+};
 
 const getCodeBlockPropsFromMdxPreNodeProps = (props: {
   children?: React.ReactNode;
 }) => {
-  const childElement = reactIs.isElement(props.children)
-    ? props.children
-    : null;
+  const childElement =
+    props.children && reactIs.isElement(props.children) ? props.children : null;
```
gabriele-ct, line 202: This can also be simplified:
```
const childElement = reactIs.isElement(props?.children) ? props.children : null;
```
As the `reactIs.isElement(undefined)` I assume will return false

PR #1816, opened 2023-09-06
fix: adjust support link

FFawzy:
support team just did a change that affects us and it requires this link change


FFawzy:
@timonrey if this can be taken in with your release that would be good too

gabriele-ct (approved):

PR #1817, opened 2023-09-07
Fix spacing in request and response details

timonrey:
A small fix for the spacing in request and response details between the displayed type and the "as". 
(In this example: "Paged as")

https://docs.commercetools.com/connect/deployments#query-deployments
Bildschirmfoto 2023-09-07 um 10 58 31


packages/gatsby-theme-api-docs/src/components/resource/method/responses.js
```diff
@@ -61,9 +61,14 @@ const Responses = ({ apiKey, responses, contentType }) => {
...
-                    {showDescription
-                      ? markdownFragmentToReact(response.description)
-                      : responseDetails}
+                    {showDescription ? (
+                      markdownFragmentToReact(response.description)
+                    ) : responseDetails ===
+                      response.body[contentType[index]].type ? (
+                      <p>{responseDetails}</p>
+                    ) : (
+                      responseDetails
```
timonrey, line 70: @gabriele-ct How would you write this piece of code? For me it feels like the nested conditions are still the best way to implement it. But I'm open for suggestions ;)

gabriele-ct: In JSX it's really controversial. I still prefer to be explicit in the conditions to I'd split this in 3 different conditionals. However I had some pushback from Nikolaus about this approach, so I'm ok to keep nested ternary operators in JSX

PR #1820, opened 2023-09-11
feat: expandable mega menu

gabriele-ct:


gabriele-ct:
Link to verify the menu: https://docskit-ga-megamenu-scaffholding.commercetools.vercel.app/

The ticket is ready for:
- Code review: I don't expect huge changes in the code, at least in terms of logic.
- Visual review: @zbalek I tried to follow as close as possible your designs, I'd like to have your visual approval.

Notes:
- The menu structure is not the smoke test one, I preferred to use the one we're going to use for the go live  so that we can have a more close to reality look of the menu. All the menu links are linking to the same page.
- E2E is failing due to the previous point, disregard it for now

FYI @timonrey @nkuehn @FFawzy

zbalek:
Hi @gabriele-ct   
- Re icon colors: please update the colors to: 00B39E and 078CDF
- Re font size: what is the font size that is currently implemented?
- Re padding: grey area - could we please reduce the padding on top and bottom to 16px? _(The overall menu size issue when it comes to a longer column is still somewhat there)_

image

gabriele-ct:
Hi @zbalek 

> * Re icon colors: please update the colors to: 00B39E and 078CDF

I got the assets from https://lordicon.com/collections?projectId=29623 if those are not the right colors, please create a new collection with the correct colors (or update the existing one) and I'll re-download them into the kit

> * Re font size: what is the font size that is currently implemented?

It's 14px as per your latest (line wrapping) design

> * Re padding: grey area - could we please reduce the padding on top and bottom to 16px? _(The overall menu size issue when it comes to a longer column is still somewhat there)_

Sure, I'll do that

> * Also, can we potentially still exchange the no-code tooling icon for this one: https://lordicon.com/icons/wired/gradient/995-table-sheet-spreadsheet ?

As all the icons will need to be re-generated, we can do that. As mentioned before, please create a new collection with the right colors and the correct icons

zbalek:
Thanks @gabriele-ct! I only defined the colors in the design ticket and wasn't aware it is possible to save in the collection. I have now updated the colors and exchanged the spreadsheet icon for the suitcase one.

gabriele-ct:
@zbalek 
I replaced the icons and reduced the padding.
https://docskit-ga-megamenu-scaffholding.commercetools.vercel.app/

nkuehn:
@zbalek & @gabriele-ct  - Instead of writing hard to understand stuff I recorded my review and testing of the megamenu as a video with talking.  “look over the shoulder” test so to say.  See here  https://drive.google.com/file/d/1-uEH81EiDhe-Fvy0N7nAKIFUSn1qvFcr/view?usp=drive_link 

Just tweaks and details, seemed easier to point out in a video.  Overall I do like it a lot.

gabriele-ct:
> Just tweaks and details, seemed easier to point out in a video. Overall I do like it a lot.

@nkuehn a picture is worth a thousand words, and a video probably more 👍 Thank you for taking the time of making it.

- The first column unexpected expand is indeed a bug. I'll look into it
- The Go back missing pointer since it's a tablet/mobile viewport, it's not strictly necessary but I'll add it anyway as it's low effort
- Regarding the first column not displaying the text when 3rd level is expanded, that's happens only on specific breakpoints, if the space is enough, the text is always visible.

Any spacings/padding/margin I'll wait for your input @zbalek

zbalek:
Hi @gabriele-ct, please see my comment below:

Mobile view
- 16px between the top area (go back and X) - same spacing as on top of it (at the moment there is only padding on the listed item on top)
- can we please also align the X with the arrows?
image


CAPS LOCK TITLES
- Can you please add padding at the bottom of the caps lock titles? (let’s try 4px first, not to expand it too much)

Offering icon
- Please update the icon to a handshake: https://lordicon.com/icons/wired/gradient/456-handshake-deal 

Animation
- Agree with Nikolaus, if possible, let’s stop the animation when user doesn’t hover over it anymore

gabriele-ct:
@zbalek ready for review.
I personally prefer the previous animation transition

nkuehn:
@gabriele-ct  yes, the animation starts harsh this way.  We could try the "morph" variant of the built-in ones (in case we use the library that the lordicons provide).  It does not stop when moving away but reverses at the point where it is.  So when hovering short it reverses short . 
@zbalek  the variants can be tried out here: https://stackblitz.com/edit/lord-icon-triggers?file=index.html  

Now one more thing (the better it gets the more can be seen:  I think the border around the selected item is changing the box size, selecting one shifts the ones above and below around a pixel or two.  Unaffected stuff should not move if possible.   Maybe related to CSS `box-sizing` choice (or, make no border a white border as a hack)

nkuehn:
PS: not sure about the alignment of "COMPOSABLE COMMERCE" at the top in this situation here

image

Would those uppercase headings maybe feel more consistent if they had the same  box size and spacing as the other menu entries?

zbalek:
Hi @nkuehn - morph seems to be a good option, from its description in the Lordicon docs I'm unsure if once cursor moves away and the icon returns to its original position if the animation stops. If it does it might be a less harsh transition @gabriele-ct

gabriele-ct:
@zbalek and @nkuehn 
[here](https://docskit-ga-megamenu-scaffholding.commercetools.vercel.app/)  you can see the morph transition. it seems to work better when user quickly hover over the icons. I also addressed the little shift in the menu items when applying the border. 
Regarding the last
Screenshot 2023-09-25 at 11 53 55
 point Nikolaus mentioned, I'm waiting for your input @zbalek . This would be the result

zbalek:
The new version feels a lot smoother - thank you @gabriele-ct! 

One (hopefully) last thing - I shared this iteration with Sven and he suggested to use slightly bigger/more prominent font for the main categories. If possible, could we please try to increase the font size in the 1st column (main categories) to 16px (currently 14px) and the font weight to medium (currently regular). I tested it in Figma (see the screenshot attached)

image

nkuehn:
@gabriele-ct  it's really smooth now  :-)  The "morph" transition is perfect for the use case

@zbalek  regarding the spacing of the leaf level of the tree that has sections I had a second look.  Thank you for the explanations.   I think what makes it look inconsistent to me is that the spacing between the sub-entries is larger than the spacing between the section heading and the sub-entries.  It looks like something's wrong with the first entry.   

If you agree, either the spacing between the "leaves" (the actual links) could be reduced to match the spacing between section heading and leaf or more space added somewhere else.    A little more density could be a good investment into the future content growth anyways, so I would prefer a version that makes the actual entries denser -  last word on you in any case, @zbalek

zbalek:
Hi @nkuehn In terms of reducing the spacing for individual entries, I think we’re a bit limited by the text wrapping in case an entry title goes over two lines. If we could make this dynamic (increase the spacing to fit two lines), we can slightly decrease the padding for individual entries. I’d be careful about reducing the space too much as this may make the menu feel too cramped as a result of insufficient white space. At the same time, adding equal (to the other entries) white white space around the subtitles may cause that users won’t realise that a subtitle is related to the subsequent entries.  
As a compromise I’d explore the following: 
- **decrease the padding around the sub-entries by 6px to 44p**x (assuming we can dynamically increase this size to keep the same top/bottom padding if we need to wrap the text to 2 lines). 
 - This change would **only be applied to direct links**
   - It would not be applied to the first column (would look too squished for the main categories) and neither to other entries which still have other sub-content.

I explored this alternative in Figma and believe it should work, despite the fact that the alignment would not match the categories with additional sub-content. I attached a screenshot of this for illustration.

cc/ FYI @Shecki 

image

nkuehn:
@zbalek  thanks for you efforts - the screen looks good and yes, you're right concerning text wrapping, I had not considered that (but would be OK if a two-line entry is higher - that portion of the UI looks a bit more like flow content anyways. 

The screenshot you posted from Figma looks intuitively good - "COMPOSABLE COMMERCE" looks like it's a bit differntly spaced than the others but I assume that's just because it's manual labor in Figma. 

Please move on without my feedback from here, all points have been made and you'll find a compromise

gabriele-ct:
@zbalek would you mind having a look here:
https://docskit-ga-megamenu-scaffholding.commercetools.vercel.app/
and confirm this is what you are expecting?
Then I'll share the link with docs team

zbalek:
@gabriele-ct Looking good! No changes from my side - there are just some direct links that are in the second column that sometime that do not behave in the same 'squished way' (e.g. UI Kit vs the other direct links are not consistent, offering links). Can we please squish all the direct links to have the same padding?

image


image

image

image

gabriele-ct:
@zbalek 
Just wanted to clarify the behaviour. 
The current logic is:
if the item is under a label -> display "squished"
if the item is not under a label -> display standard padding

This is what I had understood from your previous comment with the purpose of making the sub label items more grouped together.

If I understood that wrong, then the idea is to use reduced padding around ALL the items in the 2nd and 3rd column, regardless of them being under a label or not?

zbalek:
@gabriele-ct - apologies for not clarifying this well -> all direct links would be 'squished' = smaller padding - this would be applied to all direct links incl. those that are under a label but also those w/o a label (just to make it consistent for all direct links). _(this is what I was double checking with Sven this afternoon)_

gabriele-ct:
> @gabriele-ct - apologies for not clarifying this well -> all direct links would be 'squished' = smaller padding - this would be applied to all direct links incl. those that are under a label but also those w/o a label (just to make it consistent for all direct links). _(this is what I was double checking with Sven this afternoon)_

I see @zbalek ,  I'm wondering if you and Sven considered the fact that direct links or expandable items can be mixed in the 2nd column. You might have for example 2 direct links at the beginning, then some expadible items, then another link etc... Unless we explicitly instruct the editors to always add the expandable items first, then the direct links.
This might create a strange layout with different item paddings.

nkuehn:
> Unless we explicitly instruct the editors to always add the expandable items first, then the direct links.
This might create a strange layout with different item paddings. 

Throw a hard error as early as possible I would say (ideally let the build break or whatever else prevents it from going live).  Or, re-order them at runtime, putting the nested ones to the top so that whatever they put will end up the right way in any case.

zbalek:
Hi @gabriele-ct , if possible I'd try the solution suggested by Nikolaus and not allow the editors to mix the entries and instead always add the expandable items first, then the direct links.

gabriele-ct:
@zbalek writing test cases for the proposed solution I found a possible corner case that we didn't consider.
If in the second column there are items grouped into labels, and one of these items is expandable what shall we do?
Screenshot 2023-09-28 at 10 35 37

The current proposed implementation would remove this item from the label grouping moving it to the top of the column like this:

Screenshot 2023-09-28 at 10 44 17

So the questions are:
- do we want to allow to have an expandable item within a label grouping?
- if yes, then how do we want to render such item?

zbalek:
Hi @gabriele-ct - such corner case cannot exist in the current implementation. If an item is not a direct link it would not have a title. If an item is a subcategory, it would be placed on top as per your screenshot. And so:

- do we want to allow to have an expandable item within a label grouping? No
- if yes, then how do we want to render such item? N/A

gabriele-ct:
Thank you @zbalek, 
one more visual check https://docskit-ga-megamenu-scaffholding.commercetools.vercel.app/documentation/ before passing it to docs team.

zbalek:
Thank you @gabriele-ct - all good! 

One small thing would be the icon for the system status should be this one but the ready to be shipped: https://lordicon.com/icons/wired/gradient/1249-heart-beat?colors=primary-00b39e-secondary-078cdf

[also - the main category 'release notes' won't be implemented in the menu until we go live with the combined release notes page. probs q1 next year, depending on the tech team's capacity]

nkuehn:
@zbalek  FYI here in the kit the menu cotnent will be reverted to test content and the docs of the kit itself.  The actual menu configuration will be done in the docs repository (and MC app kit repo)

gabriele-ct:
@zbalek thank you for spotting the wrong icon. I'm passing the preview link to the docs team so they can have a check before we proceed with release.

cypress/e2e/docs-smoke-test/top-menu.cy.ts
```diff
@@ -23,11 +25,36 @@ describe('Top menu', () => {
...
+    cy.get('div[data-testid="desktop-top-menu"]')
+      .filter(':visible')
+      .within(() => {
+        cy.findByText('Product with long title which spans on multiple lines')
+          .filter(':visible')
+          .click();
+        cy.findByText('Getting started resources').filter(':visible').click();
+        cy.findByText('GETTING STARTED PAGES').should('be.visible');
+
+        cy.findByText('Product with long title which spans on multiple lines')
```
timonrey, line 54:
```suggestion
        cy.findByText('Product with long title with spans on multiple lines')
```

cypress/e2e/docs-smoke-test/top-menu.cy.ts
```diff
@@ -23,11 +25,36 @@ describe('Top menu', () => {
...
+    cy.visit(URL_DOCS_SMOKE_TEST);
+    cy.findByText('Components').click();
+    cy.findByText('Mermaid Diagrams').click({ force: true });
+    cy.get('div[data-testid="mermaid-diagram"]');
+    cy.findByLabelText('Open Top Menu').click();
+    cy.get('div[data-testid="desktop-top-menu"]').should('be.visible');
+    cy.get('div[data-testid="desktop-top-menu"]')
+      .filter(':visible')
+      .within(() => {
+        cy.findByText('Product with long title which spans on multiple lines')
```
timonrey, line 48:
```suggestion
        cy.findByText('Product with long title with spans on multiple lines')
```

packages/gatsby-theme-docs/src/components/top-menu-item.js
```diff
@@ -0,0 +1,185 @@
...
+import { LordIcon, designSystem } from '@commercetools-docs/ui-kit';
+import styled from '@emotion/styled';
+import { css, keyframes } from '@emotion/react';
+import { AngleRightIcon } from '@commercetools-uikit/icons';
+
+const MenuItemWrapper = styled.div`
+  display: flex;
+  padding: ${({ isSmall }) =>
+    isSmall ? '4px 12px 4px 8px' : '8px 12px 8px 8px'};
+  height: 32px;
```
timonrey, line 12: Would be nice if the designSystem is being used here instead of defining the pixels here.

packages/gatsby-theme-docs/src/components/top-menu-item.js
```diff
@@ -0,0 +1,185 @@
...
+
+  &:hover {
+    background-color: ${designSystem.colors.light.selectedItemBackground};
+    border-radius: 4px;
```
timonrey, line 16: Same here

packages/gatsby-theme-docs/src/components/top-menu-item.js
```diff
@@ -0,0 +1,185 @@
...
+  width: 30px;
+  border: 1px solid ${designSystem.colors.light.surfaceSecondary2};
+  border-radius: 4px;
+  box-shadow: 0px 1px 2px 0px hsla(0, 0%, 41%, 0.15);
+  margin-right: 16px;
+  background-color: ${designSystem.colors.light.surfacePrimary};
+
+  & svg {
+    height: 30px;
+    width: 30px;
```
timonrey, line 56: And here

packages/gatsby-theme-docs/src/components/top-menu-item.js
```diff
@@ -0,0 +1,185 @@
...
+  // eslint-disable-next-line react/no-unused-prop-types
+  level: PropTypes.number,
+  // eslint-disable-next-line react/no-unused-prop-types
+  shouldShrink: PropTypes.bool,
+};
+
+export const TopMenuLabelItem = styled.div`
+  font-weight: ${designSystem.typography.fontWeights['light-bold']};
+  padding: 0 8px 4px 8px;
+  margin-top: 20px;
```
timonrey, line 180: And here

packages/gatsby-theme-docs/src/components/top-menu-mobile.js
```diff
@@ -0,0 +1,188 @@
...
+import useTopMenuItems from '../hooks/use-top-menu-items';
+import { designSystem } from '@commercetools-docs/ui-kit';
+import { AngleLeftIcon, CloseBoldIcon } from '@commercetools-uikit/icons';
+import { useEffect, useState } from 'react';
+import { preProcessColumnItems } from './top-menu-components';
+import { TopMenuItem, TopMenuLabelItem } from './top-menu-item';
+
+const Container = styled.div`
+  background-color: ${designSystem.colors.light.surfacePrimary};
+  padding: 16px 24px;
```
timonrey, line 12: And here

packages/gatsby-theme-docs/src/components/top-menu-mobile.js
```diff
@@ -0,0 +1,188 @@
...
+`;
+
+const MenuNavigationArea = styled.div`
+  display: flex;
+  justify-content: space-between;
+  padding: 0 8px 16px 2px;
```
timonrey, line 18: And here, sorry to be that annoying 😄

packages/gatsby-theme-docs/src/components/top-menu.js
```diff
@@ -237,6 +252,60 @@ const TopMenu = (props) => {
...
 
+const MenuTop = styled.div`
+  display: inline-flex;
+`;
+
+const MenuBottom = styled.div`
+  background-color: ${designSystem.colors.light.surfaceSecondaryTopMenu};
+  display: inline-flex;
+  border-bottom-left-radius: 10px;
+  border-bottom-right-radius: 10px;
```
timonrey, line 266: Same as above, there are several places in this file where we could use the design system

packages/gatsby-theme-docs/src/hooks/use-top-menu-items.js
```diff
@@ -1,32 +1,74 @@
...
+          newArr.push(removeNullUndefined(obj[i])); // Recursively check objects in arrays
+        } else {
+          newArr.push(obj[i]);
+        }
+      }
+    }
+    return newArr;
+  } else if (typeof obj === 'object') {
+    var newObj = {};
+    for (var prop in obj) {
```
timonrey, line 18: Good old for loops, I like it 😁 💯

packages/gatsby-theme-docs/src/layouts/internals/footer.js
```diff
@@ -121,17 +121,6 @@ const AlignedRight = styled.div`
...
-        nodes {
-          id
-          menuTitle
-          items {
-            label
-            href
-            beta
-          }
-        }
-      }
```
timonrey, line 134: This is also being changed in the footer PR. So it would be best if we merge this PR first because it's probably easier to resolve the conflicts in my PR.

timonrey (approved):
Code looks good overall 🚀 
just left some improvement suggestions 🔧

gabriele-ct:
@timonrey Thank you for your feedbacks. Regarding the design system comments:
I believe that a design system makes sense if designers are taking in account the existing definitions (border size, margins, colors etc...) when creating a design. Simply by moving all the sizes we use to the design system we are not making our job easier.
The feeling I had developing the mega menu is that all the size/paddings/margins are subject to changes and they're not following any sort of pattern or preset or style guide (apart partially the colors). So I tend to believe that for such specific part of the site (megamenu) is fine to keep explicit values in the styled components, that makes our life a lot easier when asked to change something very specific.

As always, I'm open to discuss my view with the rest of the team.

PR #1821, opened 2023-09-12
fix: issue with scroll to item during page navigation

gabriele-ct:


timonrey (approved):
Code looks good & I tried it out in the smoke tests...looks like it's working now 👍🏼

PR #1836, opened 2023-09-20
Fix beta tag spacing in lists with link entries

### LINKED ISSUE #1813: Spacing between list items when Beta tag is used in lists
Looks like when a Beta tag is used in a list, the span margin is causing significant spacing between list items? Issue can be found [here](https://docs.commercetools.com/api/projects/types#list-of-customizable-data-types).

Screenshot

timonrey:
closes #1813

### DIFF
--- a/packages/ui-kit/src/components/markdown.tsx
+++ b/packages/ui-kit/src/components/markdown.tsx
@@ -135,6 +135,9 @@ const Li = styled.li`
   > ul {
     list-style-type: circle;
   }
+  > * + :not(span) {
+    margin-top: ${dimensions.spacings.m};
+  }
   > * + ${Ul}, > * + ${Ol}, > * + ${Dl} {
     margin-top: ${dimensions.spacings.s};
   }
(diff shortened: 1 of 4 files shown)


packages/ui-kit/src/components/markdown.tsx
```diff
@@ -136,6 +135,12 @@ const Li = styled.li`
   > ul {
     list-style-type: circle;
   }
+  > * + :not(span) {
```
gabriele-ct, line 138: This style change will affect any span tag and it might create undesired side effects
What you could to is specify a className in the PlanTag component and then change the negative rule to target just that specific span 

e.g.
```
> * + :not(span.planTag) {
```

gabriele-ct (changes requested):

packages/ui-kit/src/components/markdown.tsx
timonrey: Good idea, thank you :)

gabriele-ct (approved):
👍

PR #1842, opened 2023-09-28
chore: add displayPrimaryButton prop

gabriele-ct:


nkuehn (approved):
I have no context, approve to unblock since Fady and Timon are OOO.   Build is broken, I assume you fix that before merging.

PR #1843, opened 2023-10-02
Document new site metadata `products` and `contentType`

timonrey:
follow up: https://github.com/commercetools/commercetools-docs/pull/3906#discussion_r1340609499


gabriele-ct (approved):
LGTM, let's remember to apply this change to docs too when releasing

websites/documentation/src/content/configuration/theme.mdx
```diff
@@ -48,6 +48,8 @@ The Docs-Kit theme also receives its plugin configuration through this standard
 
 - `title` (**required**, string): Defines the title of the website.
 - `description` (**required**, string): Defines the description of the website.
+- `products` (**required**, array of strings): Defines which products the website relates to. Available options are: [`Composable Commerce`, `Frontend`, `Checkout`, `Connect`].
+- `contentType` (**required**, string): Defines the type of content in the website. Available options are: `referenceDocs`, `guidedDocs`, `userDocs`.
```
nkuehn, line 52: @timonrey  both are not required in the technical sense and there are cases where they would voluntarily be set to null.   the products should then be an empty array.

timonrey: @nkuehn But we want/need to add them to all new microsites in the docs. That's why I marked them as required. Not in a technical sense but in practice, we basically want to always add them, right?

nkuehn: There are "special purpose" sites that do not specifically belong to a product, e.g. "offering" or "docs" or stuff like that and they should then better not have a product set.  

If "contentType" is not set that's OK, too - it's better to not set it than to set it to a value that would be wrong for half of the site.    

(this from the AI search perspective)

gabriele-ct: My view:
does the build or site breaks if the parameters are not refined? (in that case I would mark them as required, specifying that they allow `null` and `[]` as "empty" values)
if the build doesn't break, then they should be marked as optional.

PR #1854, opened 2023-10-17
fix: adapt expandable menu for large content

gabriele-ct:


packages/gatsby-theme-docs/src/layouts/internals/sidebar.js
```diff
@@ -308,7 +308,7 @@ const ChapterPagesWrapper = styled.div`
   ${({ isExpanded }) =>
     isExpanded &&
     css`
-      max-height: 1000px;
+      max-height: 10000px;
```
gabriele-ct, line 311: This is not ideal but `height: auto` cannot be used in a transition. So, either we keep the current transition (expanding menu vertically) keeping in mind that there's an hard limit (10000px) or we need to give up the transition and make the content drive the div container height.

timonrey (approved):

PR #1857, opened 2023-10-18
Fix comments color when highlighted

### LINKED ISSUE #1332: Highlighted lines in code snippets don't have enough color contrast
Currently code snippets support the ability to highlight some of their lines when the writer wants to drive readers attention to an specific part of the example.
The issue we have is that, if the highlighted lines include code comments, the contrast between the code comment color and the highlighted background is not good enough so it is quite hard to actually read it.

An example can be found in [this documentation page](https://docs.commercetools.com/custom-applications/api-reference/commercetools-frontend-application-shell#createapollocontextforproxyforwardto).

image

We might need to adjust the color of the comments when they are highlighted to have a better contrast so they can be easily read.

CarlosCortizasCT:
closes #1332

This PR is a proposal to fix a problem we have with comments in code examples when they are part of a highlighted section since currently they are not visible in that situation.

Check the first line of the following example:

**BEFORE**
Screenshot 2023-10-18 at 10 29 45

**AFTER**
Screenshot 2023-10-18 at 10 28 58

### DIFF
--- a/.changeset/chilly-pillows-remain.md
+++ b/.changeset/chilly-pillows-remain.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-docs/ui-kit': patch
+---
+
+Fix color issue with comments in code examples when they are highlighted.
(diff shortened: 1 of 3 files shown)


gabriele-ct (approved):
Thank you @CarlosCortizasCT

emmenko (approved):
Love it, thanks!

PR #1865, opened 2023-10-25
feat: always showing the avatar when logged in

FFawzy:


packages/gatsby-theme-docs/src/modules/self-learning/components/profile-modal.tsx
```diff
@@ -228,17 +228,21 @@ const ProfileModal = () => {
...
-              <b>{profile.global_account_name}</b> with id{' '}
-              <b>{profile.global_account_id}</b> <br />
-              <Link
-                isExternal
-                to={`mailto:${mailToData.email}?subject=${mailToData.company_association_verify_request.subject}&body=${mailToData.company_association_verify_request.body}`}
-              >
-                Please contact us to change it.
-              </Link>
-            </Label>
+            <SpacingsStack>
```
nkuehn, line 231: Is this an unrelated fix?

packages/gatsby-theme-docs/src/overrides/topbar-login-info.js
```diff
@@ -1,10 +1,16 @@
...
 const LoginInfo = () => {
   const { hideLogin } = useContext(ConfigContext);
-  return hideLogin ? null : <UserProfile />;
+  const {
+    user: { profile },
+  } = useContext(LearningContextState);
+  if (hideLogin && profile) return <UserProfile />;
+  if (hideLogin && !profile) return null;
+
+  return <UserProfile />;
```
nkuehn, line 13: can't help but thinking this could be done as a simpler coditional.  showing the user profile is the default and the only exception is to not show it if not logged and and hiding it. 
```suggestion
  return (hideLogin && !profile) ? null: <UserProfile />;
```

websites/documentation/src/content/index.mdx
```diff
@@ -2,6 +2,14 @@
...
 ---
 
+import { LoginButton, IfUserLoggedOut, IfUserLoggedIn } from '@commercetools-docs/gatsby-theme-docs'
+
+<div style="display:flex;">
+
+  <LoginButton theme="primary" label="Log in"/>
+
+</div>
+
```
nkuehn, line 12: Looks rather like a smoke test than a part of the kit documentation

FFawzy: true .. I was just testing with that .. I'll change it

packages/gatsby-theme-docs/src/modules/self-learning/components/profile-modal.tsx
FFawzy: unrelated to the ticket purpose but relevant

timonrey (approved):

Issue #1858, opened 2023-10-18
404 page fails to load javascript


industrian:
I've noticed what appears to be a bug on 404 pages. Tested in both Chrome and Safari.

- Search cannot be used (clicking the search field does nothing)
- Left-side navigation cannot be expanded.

[Link](https://docs.commercetools.com/import-export/dfnsdfhns) to 404 page.

industrian:
The megamenu also does not seem to open on 404 pages as well.

gabriele-ct:
It might be related to [this issue](https://github.com/gatsbyjs/gatsby/issues/19618).
Needs further investigation

gabriele-ct:
After some tests here is what I've found out:

- When opening a microsite non existing page I see something like this:

![Image](https://github.com/commercetools/commercetools-docs-kit/assets/106664971/7a3cd7bd-ed97-4f19-b50f-76802a7c728c)

Essentially the first 404 is returned by the microsite 404 rule (which works as expected):
```
{
      src: `/${workingDirectoryName}/(.*)`,
      status: 404,
      headers: { "x-custom-404": "microsite404" }, // custom header added for debug
      dest: `/${workingDirectoryName}/404`,
    },
```

When the 404 page loads, the gatsby code  tries to load various resources. In the screenshot you can see the code tries to load `
https://docs.commercetools.com/getting-started/page-data/ss/page-data.json` which returns a 404, that's fine too and it's handled by gatsby correcly.

What's wrong is the fact that the code tries also to load `https://docs.commercetools.com/ss` (note the HEAD request) which matches the catchall rule BUT doesn't return a 404 but a 200. It seems like this rule:
```
{
      // any other 404 goes to the "docs" homepage 404
      src: '/(.*)',
      status: 404,
      dest: `https://docs.commercetools.com/docs/404`,
    },
```
gets executed but it doesn't send 404 error.

That somehow causes the issue that surfaces with an internal Gatsby error.

- Removing the catchall rule resolve the problem with the microsites but any 404 page in the proxy will cause the vercel 404 page to be displayed

gabriele-ct:
This is how the request would look like without the catchall:


![Image](https://github.com/commercetools/commercetools-docs-kit/assets/106664971/fd55c23c-d66c-4a3a-85c6-baf5777dc8bc)

as you can see, the fact that the `https://docs-proxy-ga-test-404.commercetools.vercel.app/sssd` actually returns a 404 error, seems to fix the issue.

[...]

PR #1872, opened 2023-11-10
feat: add PageRedirection component

gabriele-ct:
Following what suggested in the [comment](https://github.com/commercetools/commercetools-docs-kit/issues/1858#issuecomment-1803923107), I created a component `<PageRedirection>` that can be used in mdx files and essentially performs a configurable redirection after a configurable amount of time


timonrey (approved):
Nice :) Thank you 💯

PR #1888, opened 2023-12-19
feat: save chat state into localstorage

gabriele-ct:
The current implementation covers the main use case: chat is persisted in case of chat modal being closed.
I did an attempt (spent several hours on it) of chat live synchronisation between tabs but it turned out to be not working as expected (I have a local branch with a WIP implementation). I decided to use the [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore) react hook which seems to cover our use case. I roughly followed the architecture described [here](https://oakhtar147.medium.com/sync-local-storage-state-across-tabs-in-react-using-usesyncexternalstore-613d2c22819e) with no luck.

I believe live sync between tabs is not that crucial at least for go live and we can come back to it later if needed with more time.
let me know @nkuehn if you believe I should keep investing time on it or if the current implementation is good enough


packages/gatsby-theme-docs/src/modules/ai-assistant/components/chat-modal.jsx
```diff
@@ -40,9 +45,9 @@ const LEGAL_DISCLAIMER = (
   <>
     Results are generated using generative AI and should be verified prior to
     use. By interacting with the feature, you agree to our “
-    <Link href="/../offering/ai-disclaimer" target="_blank">
+    <a href="/../offering/ai-disclaimer" target="_blank">
```
timonrey, line 48: Why don't you want to use gatsby's Link component here?

websites/docs-smoke-test/src/content/components/child-section-navigation.mdx
```diff
@@ -3,6 +3,8 @@ title: Child section navigation
 navLevels: 2
 ---
 
+import { AiAssistantButton } from '@commercetools-docs/gatsby-theme-docs'
```
timonrey, line 6: Why do you import the Button here?

websites/documentation/src/content/beta.mdx
```diff
@@ -2,4 +2,7 @@
 title: Example Terms and Conditions
 ---
 
+import { AiAssistantButton } from '@commercetools-docs/gatsby-theme-docs'
```
timonrey, line 5: Same here.

packages/gatsby-theme-docs/src/modules/ai-assistant/components/chat-modal.jsx
gabriele-ct: > Why don't you want to use gatsby's Link component here?

[Gatsby Link API](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/)
It has no `href` and `target` props. It still works probably because internally it falls back to `a` but it creates a bunch of warnings in the console.

websites/docs-smoke-test/src/content/components/child-section-navigation.mdx
gabriele-ct: leftover 👍

websites/documentation/src/content/beta.mdx
gabriele-ct: same leftover 👍

packages/gatsby-theme-docs/src/modules/ai-assistant/components/chat-modal.jsx
nkuehn: @timonrey  We're either using our own Link wrapper component or pure <a> tags , we never use the gatsby link directly.   If possible, use our link wrapper (it's not flexible enough at the moment, I know - very monolithic and hard to adapt to differing styling and such).   Here it's not a big deal because it's a new window and a hardwired microsite.

timonrey (approved):

PR #1892, opened 2023-12-20
[AI] Make auth check part of the chat dialog

timonrey:
closes https://github.com/commercetools/commercetools-docs/issues/4372
closes https://github.com/commercetools/learning-tech/issues/592


packages/gatsby-theme-docs/src/modules/ai-assistant/components/chat-modal.jsx
```diff
@@ -529,6 +598,105 @@ const ChatModal = () => {
...
+          </ChatMainArea>
+          <RightBlank />
+        </ChatContainer>
+      </FormDialog>
+    );
+  }
+
+  if (isNotValidatedUser(user)) {
+    return (
+      <FormDialog
```
gabriele-ct, line 650: Do we really need to duplicate the whole `<FormDialog>` for not authenticated and not validated users? Isn't possible to use some conditionals within the existing `<FormDialog>` ?

packages/gatsby-theme-docs/src/modules/ai-assistant/components/chat.utils.js
```diff
@@ -20,10 +20,6 @@ const isAllowedEmailDomain = (email, domainList) => {
   return allowedEmailDomainRegExp.test(email);
 };
 
-export const isAllowedUser = (user) => {
```
gabriele-ct, line 23: Not sure if this has been discussed. But `isAllowedUser` had been added to give the possibility to whitelist only specific email domains (customers). I guess this is not necessary anymore as the go live will be for all customers. But worth to check @nkuehn

gabriele-ct (changes requested):

packages/gatsby-theme-docs/src/modules/ai-assistant/components/chat.utils.js
nkuehn: Yes, but not yet.  We'll do that only shortly before go-live. 

But we'll likely need a denylist on the API server to handle known misuse attempts.

gabriele-ct: I'll add a ticket in the go-live, for now, let's leave it @timonrey

packages/gatsby-theme-docs/src/modules/ai-assistant/components/chat-modal.jsx
timonrey: We could also do conditionals here but the code would be much harder to read in my opinion. However, we can move the two cases into separate components to reduce the code in the chat-modal component.

packages/gatsby-theme-docs/src/modules/ai-assistant/components/chat.utils.js
timonrey: I removed the `isAllowedUser` because it was not used anywhere anymore and there is also an identical negated value for isNotAllowedUser already.

gabriele-ct: I think you might got confused with `isNotValidatedUser` which has a different meaning.
`isAllowedUser` IMHO should stay preventing the chat to be launched if the user is not in the whitelist, up to you where you want to use it (either in the launch button or in the modal), it's temporary anyway we'll remove it when we go finally live

timonrey: Maybe it has a different meaning, but the function is identical to `isAllowedUser`, just returning false of user is allowed. So I used it instead of `isAllowedUser`.
https://github.com/commercetools/commercetools-docs-kit/pull/1892/files#diff-e625f24fbc71d5d652eca48d7d23c61d1d019faf2a0ddb4667bd82395c7954f0L27

packages/gatsby-theme-docs/src/modules/ai-assistant/components/chat-modal.jsx
timonrey: --> https://github.com/commercetools/commercetools-docs-kit/pull/1892/commits/7451e4e2e94ca502b757e0d89b1cd4fa6379cba5

gabriele-ct (approved):

PR #1893, opened 2023-12-21
fix: null when root breadcumb doesn't exist

gabriele-ct:


gabriele-ct:
An example of how the breadcrumbs are now displayed
Screenshot 2023-12-21 at 09 46 31

timonrey (approved):
Thanks! 💯

PR #1901, opened 2024-01-04
Ga UI kit upgrade

gabriele-ct:


FFawzy (approved):
Thanks for taking care of it

PR #1916, opened 2024-01-16
feat: add additional meta tags for algolia indexing

gabriele-ct:


algolia-crawler.js
```diff
@@ -39,6 +39,16 @@ new Crawler({
             lvl2: 'article h2',
             lvl3: ['article h3', 'article h4'], // an array is a list of fallbacks. Authors sometimes omit h3 to skip the side nav.
             lvl4: ['article h4', 'article h5'],
+            products: {
```
FFawzy, line 42: @gabriele-ct 
this structure is called custom fields in algolia which then allows you to make Filter queries, which can work. 
another approach can also be the Tags feature which is used by the reserved word `_tags:['product1','product2']`

we will need to compare which of them are better in terms of our requirements and the requirement that might happen in the future (having multiple levels of filtering)
would it better be served by custom Filters or Tags

nkuehn: "levels" as in hierarchy are not something the kind of engine algolia is does OOTB,  it can be emulated through multiple fields each of which represents a level.   See how the heading levels are modeled in docsearch.  

But I would not focus on it yet, the hieararchy style filtering Ux has many issues and I am not convinced it is sustainable.   Focus on properly modeling the multi-value product attribute first, that's more important (e.g. yo may have to use tags for that and a field for the breadcrumb and content type if there are no multi-value fields available.

gabriele-ct: @FFawzy 
certainly it makes more sense to use tags. I didn't know the API, I'll dig a bit more into it

gabriele-ct:
@FFawzy 
I followed the `_tags` approach, I'm wondering if we want to prefix the tags in some way to differenciate and avoid conflicts between product and contentType and make it more clear.
In terms of "preferred" approach I had found a doc page on algolia site that described the difference between tag and attribute search. One of the main thing is that the tags are not text searchable so they will not affect in any way the current text search functionality (which is what we want). 
Overall seems like tags are more appropriate for out usecase.

FFawzy:
sounds like the better approach to me too. Thanks for digging into it.
we don't need then the `product` property in the record anymore.

regarding prefixing, the terms currently do not overlap so I guess we are ok for now could be harder in the client side to figure out which prefix to use if the UI itself is not having the same distinction. 
in any case if we get this problem in the future it is as simple as introducing it here and running the crawler one more time.

algolia-crawler.js
```diff
@@ -39,6 +57,11 @@ new Crawler({
...
```
FFawzy, line 60: this is the one I'm talking about in my previous comment .. do we still need it ? since now we have `_tags`

PR #1921, opened 2024-01-18
feat: add tag search to algolia search functionality

gabriele-ct:


gabriele-ct:
@zbalek would you be able to have a look to the implementation I got so far and give me some feedbacks? https://docskit-ga-algolia-search-with-tags.commercetools.vercel.app/docs-smoke-test/

Thank you

packages/gatsby-theme-docs/src/components/advanced-search-filter.tsx
```diff
@@ -0,0 +1,115 @@
...
+    props.setFilters(tagsFilter);
+  };
+
+  const onClearFilterClick = () => {
+    setTagsFilter([]);
+    props.setFilters([]);
+  };
+
+  const toggleTagFilter = (tag: string) => {
+    console.log(tag);
```
timonrey, line 71: console.log left over

packages/gatsby-theme-docs/src/components/algolia-search.js
```diff
@@ -233,17 +234,22 @@ const AlgoliaSearch = React.forwardRef((props, ref) => {
...
         .then(({ default: docsearch }) => {
-          docsearch({
+          const docSearchClientInstance = docsearch({
             apiKey: '058c5342ed03a928f2dde0142bee8db0',
             indexName: 'commercetools',
             appId: '6NZ83LOI5M',
             inputSelector: `#${props.searchInputId}`,
             bindKeyboardShortcuts: false,
-            debug: process.env.NODE_ENV !== 'production',
+            debug: true,
```
timonrey, line 243: Do we really want to always run the client in debug mode?

zbalek:
Hi @gabriele-ct - looks good. However, in the designs I was counting with also extending the search box to 40px. We can keep the search box as is to potentially save some space but in that case, can we please adjust the height of the filter button to 32px to match the search box?

FFawzy:
1 small thing, can we have the drop down closable when clicking away please ?

packages/ui-kit/src/utils/lord-icon.ts
```diff
@@ -57,6 +58,7 @@ const iconNameToLottieJSON: Record<string, object> = {
   document: DocumentJson,
   api: ApiJson,
   building: BuildingJson,
+  funnnel: FunnelJson,
```
timonrey, line 61:
```suggestion
  funnel: FunnelJson,
```

timonrey (commented):
I really like the new changes! 🤩🚀 
I just have some small comments, but besides that it's good to go 🙌🏼

zbalek:
> 1 small thing, can we have the drop down closable when clicking away please ?

Also, if a user selects a filter and then clicks away -> please count this as if the user clicked 'apply filters'

FFawzy:
Final adjustments after a call with the stakeholders: 
- going back to buttons design ([check Figma here](https://www.figma.com/file/GqhHFUcEaMxBXnMM9TXdaZ/%F0%9F%9F%A2-Feedback-Board?type=design&node-id=4151-2670&mode=design&t=njYKXyN39OVNKojg-0))
- adding google analytics to track any clicks for the filters ( to understand if people are using it or not)

gabriele-ct:
Change of UI

PR #1922, opened 2024-01-23
feat: add tag search to algolia search functionality

gabriele-ct:


gabriele-ct:
@zbalek @FFawzy 
WDYT: https://docskit-ga-algolia-search-with-tags-v2.commercetools.vercel.app/docs-smoke-test/

zbalek:
Hi @gabriele-ct, can you please adjust the spacing between the buttons to match design?If user selects a box and it increases in size, I'd expect the spacing to be kept on the left and the other elements on the right to move to the right to keep the spacing.

gabriele-ct:
> Hi @gabriele-ct, can you please adjust the spacing between the buttons to match design?If user selects a box and it increases in size, I'd expect the spacing to be kept on the left and the other elements on the right to move to the right to keep the spacing.

Hi @zbalek ,
you can have a look to this implementation https://docskit-ga-algolia-search-with-tags-v2.commercetools.vercel.app/docs-smoke-test/ keeping in mind that:
- in your designs the spacing and the layout of the buttons is constant because you increased the container size (which is not possibile)
- adding the icon to the buttons increases the width of the button which, in turns, causes some form or layout change (we can agree on keeping the spacing between button constant, or to justify content as it was before, but something will inevitably shift).

Let me know if you're ok with the current solution 
Thank you

gabriele-ct:
> looks good tech wise too .. lets do a final round of cleanup like removing the icon that we don't need anymore ... etc and then lets 🚀

Ok, I leave the `LordIconButton` component as I have the feeling we're going to use it soon

PR #1930, opened 2024-01-31
docs: add algolia documentation

gabriele-ct:


FFawzy (approved):
Thanks Gabriele, looks good to me ... can you also take a copy of the crawler from Algolia and to update the repo version, Nikolaus did some changes recently there directly .. so lets add it to keep the repo file consistent

Thanks

PR #1937, opened 2024-03-01
chore: add user guiding script

gabriele-ct:


packages/gatsby-theme-docs/src/templates/homepage.js
```diff
@@ -69,6 +70,7 @@ export function Head({ data, pageContext }) {
         // eslint-disable-next-line react/prop-types
         excludeFromSearchIndex={data.contentPage.excludeFromSearchIndex}
       />
+      <UserGuiding ugId={'U4I78799B6RID'} />
```
gabriele-ct, line 73: @FFawzy any better ideas on how/where inject the script?

PR #1941, opened 2024-03-12
feat: self-learning homepage

gabriele-ct:


FFawzy (approved):
most of the changes are in the components that will move to the docs repo so even if there are some content changes that are needed can be directly done there. so approving to move it and we can run a more thorough content review there.
Thanks

PR #1947, opened 2024-03-19
[New branding] Apply design changes

timonrey:
part of https://github.com/commercetools/commercetools-docs/issues/4366

**This PR**
- implements all design changes for the kit suggested by Zoe that was described in the ticked above
- also removes the color preset functionality since it is obsolete now. (It was used to configure the homepage banner and the site icon in the upper right corner...since we decided to removed the site icons and to use only one homepage banner, it became obsolete) --> **TODO after the next release**: Adjust all the microsite configs to remove the color presets

~~FYI: One problem remains...one of the lordicons we use in the top menu was not provided in the icon folder here:
https://lordicon.com/collections?projectKey=nwuauype (it's the banner icon we use for getting started)
I'm still trying to find it somewhere or to see if I can adjust it on my own. Otherwise we would need to use a different icon.~~


packages/ui-kit/src/icons/lord-icon/building.json
```diff
@@ -2047,7 +2047,7 @@
               "nm": "Color",
               "mn": "ADBE Color Control-0001",
               "ix": 1,
-              "v": { "a": 0, "k": [0.027, 0.549, 0.875], "ix": 1 }
```
gabriele-ct, line 2050: @timonrey please ensure all the .json have the right color (I might be wrong but it looks some are missing). Also all the `../svg/lord-icon-*` need to change their color. These are the static assets used in the SSR version of the page before hydration

gabriele-ct:
@timonrey the footer of megamenu seems to be of a different color compared to the designs (TBH, I don't trust too much those designs, I might have missed some comment)

timonrey:
> @timonrey the footer of megamenu seems to be of a different color compared to the designs (TBH, I don't trust too much those designs, I might have missed some comment)

Are you sure? Double checked it now, should be the same color.

gabriele-ct:
> > @timonrey the footer of megamenu seems to be of a different color compared to the designs (TBH, I don't trust too much those designs, I might have missed some comment)
> 
> Are you sure? Double checked it now, should be the same color.

Probably it's me, disregard my comment

packages/ui-kit/src/icons/lord-icon/building.json
timonrey: @gabriele-ct I changed the svg's now too, but only the ones that we currently use in the mega menu. There are 6 other lordicons that were not in the design folder and are also not used in the mega menu atm. Do we need them for other purposes?

.changeset/quick-lies-invent.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-docs/gatsby-theme-docs': minor
```
emmenko, line 2: If you are removing functionalities, you should bump this as `major`

```suggestion
'@commercetools-docs/gatsby-theme-docs': major
```

emmenko (commented):
😇

FFawzy:
@timonrey here is how the megamenu looks like in [Figma](https://www.figma.com/file/GqhHFUcEaMxBXnMM9TXdaZ/%F0%9F%9F%A2-Feedback-Board?type=design&node-id=4804-17224&mode=design&t=XFvaKHxWNu0ArjTM-0)
vs how it looks in the preview site
Screenshot 2024-03-19 at 6 26 01 PM
Screenshot 2024-03-19 at 6 27 46 PM

.changeset/quick-lies-invent.md
gabriele-ct: @emmenko , thank you for the comment. This is going to be deployed to docs as part of an ongoing release which is already bumping a major version, technically this is a patch of the major. So no need to re-bump the major FYI @timonrey

packages/ui-kit/src/icons/lord-icon/building.json
gabriele-ct: @timonrey only the ones that are used in the mega menu needs to change color, the others are typically more complex svgs that are used in other parts of the site (mostly on the learning path homepage)

.changeset/quick-lies-invent.md
emmenko: I understand but it's still good practice to follow semantic versioning correctly.

For instance, we have the changesets that provide a great way to document changes. In particular for breaking changes it allows to document things properly (including e.g. migration steps).

Having multiple changesets with major version bumps is also not a problem, as it actually allows to better differentiate the changes in the release.

timonrey:
> @timonrey here is how the megamenu looks like in Figma
@FFawzy Right! I was looking at the archive section in the designs where it still had this color. I'll change it back to grey 👍🏼

.changeset/quick-lies-invent.md
gabriele-ct: Makes sense.

PR #1949, opened 2024-03-20
fix: expanding chapter are incorrectly expanding

gabriele-ct:


timonrey (approved):
👍🏼

timonrey (changes requested):
I checked the preview and the snapshots and it seems like there is a problem with the expandable chapters when using a deep link into a site. It doesn't open the chapter on the left...they stay closed. You can also see that in the snapshots.

gabriele-ct:
> I checked the preview and the snapshots and it seems like there is a problem with the expandable chapters when using a deep link into a site. It doesn't open the chapter on the left...they stay closed. You can also see that in the snapshots.

Thank you, I'll have a better look

gabriele-ct:
> I checked the preview and the snapshots and it seems like there is a problem with the expandable chapters when using a deep link into a site. It doesn't open the chapter on the left...they stay closed. You can also see that in the snapshots.

Thank you for spotting the issue. I didn't spotted it since it happens only when deployed (it has to do with sitePrefix). Now it should be working as expected

timonrey (approved):
Thank you!

PR #1950, opened 2024-03-20
Refactor Lordicons and apply new branding

timonrey:


gabriele-ct (approved):

FFawzy:
are we reverting the colors of the icons ? why ?

timonrey:
> are we reverting the colors of the icons ? why ?

We found a problem with the lordicons and the ways how they are being generated. @gabriele-ct and I discussed it and we agreed to revert the changes and refactor the lordicons in the docs ui-kit before we release them.

PR #1954, opened 2024-03-22
fix: disable send button for AI readonly mode

gabriele-ct:


gabriele-ct:
@timonrey I fixed the vrt issue you mentioned. You can go ahead with the reivew

timonrey (approved):
Thank you 💯

PR #1964, opened 2024-04-05
fix: add to each page a canonical tag pointing to production

gabriele-ct:


packages/gatsby-theme-docs/gatsby-browser.js
```diff
@@ -117,6 +117,8 @@ export const wrapRootElement = ({ element }, pluginOptions) => {
           aiAssistantApiBaseUrl: pluginOptions.aiAssistantApiBaseUrl,
           aiAssistantTopbarButton:
             pluginOptions.aiAssistantTopbarButton || false,
+          enableCanonicalUrls:
```
timonrey, line 120: Why not just...
```suggestion
          enableCanonicalUrls: pluginOptions.enableCanonicalUrls,
```
?

packages/gatsby-theme-docs/gatsby-ssr.js
```diff
@@ -122,6 +122,8 @@ export const wrapRootElement = ({ element }, pluginOptions) => {
         hideLogin: pluginOptions?.hideLogin || false,
         aiAssistantApiBaseUrl: pluginOptions.aiAssistantApiBaseUrl,
         aiAssistantTopbarButton: pluginOptions.aiAssistantTopbarButton || false,
+        enableCanonicalUrls:
+          pluginOptions.enableCanonicalUrls !== false ? true : false,
```
timonrey, line 126: Same here

packages/gatsby-theme-docs/src/components/config-context.ts
```diff
@@ -25,6 +26,7 @@ const ConfigContext = createContext<Config>({
   selfLearningFeatures: [],
   aiAssistantApiBaseUrl: '',
   aiAssistantTopbarButton: false,
+  enableCanonicalUrls: false,
```
timonrey, line 29: Because you defined the default here.

timonrey (approved):
Looks good, just some comments

packages/gatsby-theme-docs/gatsby-browser.js
gabriele-ct: That's because pluginOptions.enableCanonicalUrls can be `undefined` but we want to default to `true` instead of `false`. That's the reason of the weird syntax

packages/gatsby-theme-docs/src/components/config-context.ts
gabriele-ct: For some reason this default value (if set to true, as it should be) doesn't work... so I had to resort to defaulting to false and then use the logic above to actually use `true` as default. Strange but I couldn't figure out the reason

PR #1987, opened 2024-05-27
chore: remove filters from release notes list page

gabriele-ct:
This needs to be merged once we're ready to go live with new release notes filters


packages/gatsby-theme-docs/src/components/release-notes-subscribe-links.js
```diff
@@ -34,12 +27,6 @@ const linkStyles = css`
 
 const ReleaseNotesSubscribeLinks = () => (
   <SpacingsStack scale="xs">
-    <ExternalSiteLink href={withPrefix('/releases/feed.xml')} css={linkStyles}>
```
gabriele-ct, line 37: I think we should remove the RSS link too, it's the one pointing to the original feed.xml file.

timonrey (approved):

FFawzy:
@gabriele-ct do we need to adjust the template for the details page too ?
the `< All Release Notes` button

Screenshot 2024-06-12 at 11 51 57 AM

gabriele-ct:
@FFawzy I believe we should. I'll add the same logic

PR #2000, opened 2024-06-12
fix: attempt to fix release notes links

gabriele-ct:


packages/gatsby-theme-docs/src/utils/release-notes.js
```diff
@@ -1,6 +1,19 @@
 export const MIN_DATERANGE = 0;
 export const MAX_DATERANGE = 2537011284;
 
+export const isDocsKitSite = (siteTitle) =>
+  [
```
gabriele-ct, line 5: This is horrific, but I couldn't find a better idea.

websites/docs-smoke-test/src/pages/release-notes.js
```diff
@@ -0,0 +1,39 @@
+import { ThemeProvider } from '@commercetools-docs/gatsby-theme-docs';
```
gabriele-ct, line 1: This is here just to have a page to use as search placeholder. No functionalities are implemented here. To avoid 404

packages/gatsby-theme-docs/src/utils/release-notes.js
```diff
@@ -15,7 +28,7 @@ const mapSiteTitleToFacetFilter = new Map([
   ['Import API', { group: 'product', productArea: 'Import API' }],
   [
     'Merchant Center Customizations',
-    { productArea: 'Merchant Center Customizations' },
+    { group: 'product', productArea: 'Merchant Center Customizations' },
```
timonrey, line 31: If we implement it like this, it would mean that we have to add all new microsites that we want to create in the docs repo here as well...which means that we would need to go through a kit-release as part of creating a new docs microsite.
If possible, I would like to avoid that...

FFawzy (approved):
Thanks 🙏 .. much better now than the previous structure.

we can even figure out later a way to handle the "return only production link" ... there are already multiple ideas we can do simply.

PR #2026, opened 2024-06-26
Support default values for query parameters

### LINKED ISSUE #1435: Add support for default values of query parameters
We have the need to document the default value for a query parameter, but this is currently not supported by the Docs Kit.

Given this RAML spec:

```yaml
queryParameters:
      staged?:
        type: boolean
        default: false
```
The rendered page should show the same box on the respective query parameter as for default values on data types.
default-false

timonrey:
closes #1435

### DIFF
--- a/packages/gatsby-theme-api-docs/src/components/info-value.js
+++ b/packages/gatsby-theme-api-docs/src/components/info-value.js
@@ -1,0 +1,67 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/react';
+import {
+  Markdown,
+  useISO310NumberFormatter,
+  designSystem,
+} from '@commercetools-docs/ui-kit';
+
+const InfoValue = (props) => {
+  const value = props.children;
+  const valueType = props.type || typeof value;
+  const formatNumber = useISO310NumberFormatter();
+
+  const customCodeStyle = css`
+    font-size: ${designSystem.typography.fontSizes.small};
+  `;
+
+  console.log(valueType, value);
+
+  switch (valueType) {
+    case 'boolean':
+      return props.isQueryParameter ? (
+        <>
+          :{' '}
+          <Markdown.InlineCodeWithoutBox css={customCodeStyle}>
+            {value}
+          </Markdown.InlineCodeWithoutBox>
+        </>
+      ) : value ? (
+        ''
+      ) : (
+        <>
+          :{' '}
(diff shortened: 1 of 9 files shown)


packages/gatsby-theme-api-docs/src/components/info-value.js
```diff
@@ -0,0 +1,67 @@
...
+    font-size: ${designSystem.typography.fontSizes.small};
+  `;
+
+  console.log(valueType, value);
+
+  switch (valueType) {
+    case 'boolean':
+      return props.isQueryParameter ? (
+        <>
+          :{' '}
```
gabriele-ct, line 25: can't we just move this out of the component? it seems not very reusable as it's assuming a certain formatting is out of the component

packages/gatsby-theme-api-docs/src/components/resource/method/parameters.js
```diff
@@ -160,6 +162,17 @@ function ParameterRow(props) {
               <Info>{props.parameter.additionalDescription}</Info>
             </div>
           )}
+          {props.parameter.default && (
+            <div>
+              <Info>
+                Default
```
gabriele-ct, line 168: I'd add `Default: ` here and remove the `{` `}`  from the [info-value.js](https://github.com/commercetools/commercetools-docs-kit/pull/2026/files#diff-0adadd2d413fac55014359abd5a48ef13846be40584169c38e774a04af23d12a) file

gabriele-ct (approved):

packages/gatsby-theme-api-docs/src/components/info-value.js
```diff
@@ -0,0 +1,62 @@
...
+
+  const customCodeStyle = css`
+    font-size: ${designSystem.typography.fontSizes.small};
+  `;
+
+  switch (valueType) {
+    case 'boolean':
+      // We want to show the actual boolean value for the query parameter defaults.
+      return props.isQueryParameter ? (
+        <>
```
gabriele-ct, line 23: I think also this react fragment is not needed @timonrey

timonrey: Right 👍🏼

timonrey: @gabriele-ct One thing I saw now in the snapshots...we have the uniqueItems info that does not have any value...we would render it with a colon like this so I think it's better that the keep it as it was...that the component handles the colon.

Screenshot 2024-07-04 at 10 01 48

PR #2031, opened 2024-06-28
chore: display search results in scrollbar

gabriele-ct:


timonrey (approved):
I tested the change in different browsers and edge cases to see if this wasn't a bugfix. But everything looks good from what I saw. I think we can proceed with this change.

Changeset is missing, but besides that we are good to go 👍🏼

Issue #2054, opened 2024-08-19
Investigation: decide if we can use the tablist component from app-kit or look for alternative solutions


gabriele-ct:
[Solutions](https://commercetools.atlassian.net/wiki/x/YYBKP)

The goal of the ticket is decide which component to be used:
 - check if using directly the app-kit package is doable and what's the overhead in terms of bundle size
 - evaluate the dev time of "coping" the exiting component from app-kit into the docs-kit repo
 - use a different component instead: view switcher from ui-kit (possibly avoid this option)

gabriele-ct:
This task has been covered by another ticket #2055

PR #2058, opened 2024-08-22
Try out using an appkit component directly in the docs kit

timonrey:
is part of #2054 

link to the appkit component docs: (Note: the code example is only works for react-router v5)
https://docs.commercetools.com/merchant-center-customizations/api-reference/commercetools-frontend-application-components#tabularmainpage


timonrey:
We decided to not go for the appkit component.

Issue #2060, opened 2024-08-22
[exploration] define how to express in mdx


FFawzy:
write a test case /sample in the docs-kit
- check if the mdx file and the way we wrote MDX inside our component is acceptable to our authors
- write test cases in the docs-kit to try out and demonstrate to the editors what can and cannot be done.
- can we write plain markdown? (if not, total blocker, it should work)
- can we still use custom component inside markdown?

gabriele-ct:
Here are my findings.

**Proposed structure**

```
<MultiPathBlock title="some title here">
  <PathBlock language="javascript">
    // js content
  </PathBlock>
  <PathBlock language="php">
    // php content
  </PathBlock>
</MultiPathBlock>
```

**Allowed markdown**
All the standard md syntax is supported, I can't see any issue with md typically editors use.
Headings are to be avoided since the side navigation is built upon them and it will look broken if we use them inside a path block

**Custom components**
Again, most of the component seem to be working fine as you can [see here](https://docskit-ga-multiple-path-exploration.commercetools.vercel.app/)
I found that `SideBySide` component is not working, `MultiCodeBlock` should be avoided since it's creating confusion. Other than that  I could not see any other blockers. I even tested some of the self learning conditional rendering components, and they work fine.

Could you @FFawzy @nkuehn @zbalek have a look and let me know if I missed something important. Thank you

gabriele-ct:
sorry, I had posted incomplete url: https://docskit-ga-multiple-path-exploration.commercetools.vercel.app/docs-smoke-test/views/multi-path-block

FFawzy:
Nice, Thanks a lot @gabriele-ct , I'd say this is already a great start. I'd say lets continue writing some more specs for the `MultiPathBlock` component.

I'd suggest the `PathBlock` props are not bound to a `language`, since there will be usecases where it is pure content (eg, Merchant Center, Frontend Studio), I would vote for keeping the props simple to `label: string, key: string` where the label can be used for the UI and the key used for state 

wdyt?

gabriele-ct:
Definitely, I had thought about it. My only concern is about the "synchronisation" beetween the `MultiPathBlock` and the `MultiCodeBlock`. Has this been descoped? If not we would sync them only if there's an exact match between the `label` in one case and the `language`?

gabriele-ct:
Also, do we need a `title` prop? I guess it depends on the designs but I guess we don't really need it, at least if we want to mimic the app kit component.

[...]

PR #2064, opened 2024-08-27
feat: MultiPathBlock implementation

gabriele-ct:
** Description of changes **
We are implementing a new component called `MultiPathBlock` after what researched [here](https://commercetools.atlassian.net/wiki/spaces/DOCS/pages/1011515489/Path+Selector+Solution+Exploration)

** Link original ticket **
https://github.com/commercetools/commercetools-docs-kit/issues/2060

Screenshot 2024-08-30 at 14 10 54

- [x] user documentation added?
- [ ] stakeholders approve changes? (if needed)


gabriele-ct:
@zbalek would you mind validate the current UI/UX:
We also added the possibility of deep link to a specific tab (by adding a ?activePath={language}) parameter

https://docskit-ga-multiple-path-exploration.commercetools.vercel.app/docs-smoke-test/views/multi-path-block

and 
https://docskit-ga-multiple-path-exploration.commercetools.vercel.app/documentation/writing/paths

zbalek:
Looks good @gabriele-ct! I also really appreciate the the possibility of deep link to a specific tab. 

One small note: Can we please add some additional spacing (8px) between the tabs and the first line of text?  It seems a bit too squished together at the moment.

gabriele-ct:
@zbalek 
What about this:
Screenshot 2024-09-02 at 10 28 10

In the screenshot I'm using the standard 16px padding

gabriele-ct:
Hey @lukpan and @basebander and @industrian,
I think we got to a good point with the implementation. Would you like to have a look so we can merge this and you can start experimenting with this new component?

Currently we don't have a solution for the content indexing. It's tricky because not only we need to index the content, but also display the correct tab selected once the user clicks on a search result. We'd like to de-scope the content indexing from this ticket and have more time to explore possible solutions.

Also, it would be good to hear your voice about how critical is content indexing (consider that code blocks are not indexed anyway). Is it a show stopper? we can live without it? we need it, but not  on day 1

Thank you

industrian:
> Hey @lukpan and @basebander and @industrian, I think we got to a good point with the implementation. Would you like to have a look so we can merge this and you can start experimenting with this new component?
> 
> Currently we don't have a solution for the content indexing. It's tricky because not only we need to index the content, but also display the correct tab selected once the user clicks on a search result. We'd like to de-scope the content indexing from this ticket and have more time to explore possible solutions.
> 
> Also, it would be good to hear your voice about how critical is content indexing (consider that code blocks are not indexed anyway). Is it a show stopper? we can live without it? we need it, but not on day 1
> 
> Thank you

Hi @gabriele-ct , thanks again for the work on this.

From my own perspective, I can't think of any current situation I'd use this in where indexing the content was essential. For example, I don't think it would be suitable to apply this to an entire page or to contain headers within it.

If others needed indexing the content, then one possible solution to displaying the correct tabs would be to possibly add a parameter denoting the selected language (`https://docs.commercetools.com/example-microsite/example-page?code=js`). But I'm not sure of the feasibility of this idea.

gabriele-ct:
> o displaying the



> possibly add a parameter denoting the selected language

Thank you @industrian for your feedback. the feature you describe is already available (I've added it on Friday), the parameter is called `activePath` and it's documented. However it's technically not feasible for the indexer to index the content and point to the actual relevant tab (url with the correct `activePath`). We discussed it with @FFawzy and we agreed that for now the algolia search will have this limitation: the content within the component will be indexed, but the correct tab will not be automatically selected

lukpan:
> Hey @lukpan and @basebander and @industrian, I think we got to a good point with the implementation. Would you like to have a look so we can merge this and you can start experimenting with this new component?
> 
> Currently we don't have a solution for the content indexing. It's tricky because not only we need to index the content, but also display the correct tab selected once the user clicks on a search result. We'd like to de-scope the content indexing from this ticket and have more time to explore possible solutions.
> 
> Also, it would be good to hear your voice about how critical is content indexing (consider that code blocks are not indexed anyway). Is it a show stopper? we can live without it? we need it, but not on day 1
> 
> Thank you

@gabriele-ct This looks great mate, thank you. The indexing is definitely not a show stopper, I can work around it, and keep it in mind as I'm chunking information. I will let you know if it does become an issue in future. Happy for it to be released. :)

packages/ui-kit/src/hooks/use-selected-path.ts
```diff
@@ -0,0 +1,82 @@
...
+    setSelectedPath(value);
+
+    // Update the activePath query string parameter
+    const urlParams = new URLSearchParams(window.location.search);
+    urlParams.set('activePath', value);
+    const newLocation =
+      (!window.location.search && !value) || value === ''
+        ? window.location.pathname
+        : `${window.location.pathname}?${urlParams.toString()}`;
+    window.history.replaceState({}, '', newLocation);
```
ColinRosati, line 70: Instead of using history to manage url we can use the URLSearchParams API to set the new location. Can you explain why you want to manage url from history? Its not clear to me

packages/ui-kit/src/hooks/use-selected-path.ts
```diff
@@ -0,0 +1,82 @@
...
+    const value = inputValue.toLowerCase();
+    // Update the value in local storage
+    localStorage.setItem(PATH_LOCAL_STORAGE_KEY, value);
+    setSelectedPath(value);
+
+    // Update the activePath query string parameter
+    const urlParams = new URLSearchParams(window.location.search);
+    urlParams.set('activePath', value);
+    const newLocation =
+      (!window.location.search && !value) || value === ''
```
ColinRosati, line 67: For readability WDYT about moving these 2 conditions to declarect variables. 

```
const hasQueryParams = window.location.search.length !== 0
const hasInputValue = value.length !== 0;
const hasNoInputOrQueryValue = !hasQueryParams && !hasInputValue
newLocation = hasNoInputOrQueryValue ? ... :  ....
```

packages/ui-kit/src/hooks/use-selected-path.ts
```diff
@@ -0,0 +1,82 @@
...
+const PATH_LOCAL_STORAGE_KEY = 'selected_path';
+const PATH_CHANGE_EVENT = 'selectedPathChange';
+
+const useSelectedPath = () => {
+  const [selectedPath, setSelectedPath] = useState<string | null>(null);
+
+  useEffect(() => {
+    // Read the initial value from local storage
+    const urlParams = new URLSearchParams(window.location.search);
+    const activePath = urlParams.get('activePath');
```
ColinRosati, line 22: WDYT about also extracting the const string so its available to the whole file and re-use this below LOC 65

packages/ui-kit/src/hooks/use-selected-path.ts
```diff
@@ -0,0 +1,82 @@
...
+    return () => {
+      window.removeEventListener('storage', handleStorageChange);
+      window.removeEventListener(
+        PATH_CHANGE_EVENT,
+        handleCustomEvent as EventListener
+      );
+    };
+  }, []);
+
+  const updateSelectedPath = (inputValue: string) => {
```
ColinRosati, line 57: We might be able to refactor this update  a bit by making 2 single responsibility branches of behaviour. They both grab the value from their distinct places.
1. input value branch
2. url branch

For me its unclear what behaviour is unique to both paths or does all behaviour happen for both input and url?

We could even break this function into 2 so we have 1 wrapper which handles taking input or value and normalizes it to a single value. Then that value is passed to the pure update functionality

packages/ui-kit/src/components/multi-path-block.tsx
```diff
@@ -0,0 +1,212 @@
...
+              key={labelSyncItem.label}
+              label={labelSyncItem.label}
+              isActive={labelSyncItem.label === selected.label}
+              onClick={(e) => onTabHeaderClick(e)(labelSyncItem)}
+            />
+          ))}
+        </Spacings.Inline>
+      </SelectorsContainer>
+      <PathsContainer>
+        {Array.isArray(props.children) ? (
```
ColinRosati, line 189: What do you think about extracting this child/children component to a small component so we reduce the comments and logic in the markup. It might be helpful so we dont have comments in markup. 
```
const TabContentChildren = ...

LOC 189
<TabContentChildren />

```

ColinRosati (commented):
Nice feature. I added a few minor style suggestions and some questions

packages/ui-kit/src/components/multi-path-block.tsx
```diff
@@ -0,0 +1,212 @@
+import styled from '@emotion/styled';
```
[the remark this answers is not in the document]
gabriele-ct: @ColinRosati,
thank you for your review. I see what you mean here. However, we tend to extract to extract to separate files functions or components that are actually used at least in one other place in the codebase.
Both `TabHeader` and the `extractLabelSyncPair` util function are tightly coupled with the `MultiPathBlock` component implementation and probably it's more convenient to have them in the same file if any change is needed.

packages/ui-kit/src/components/multi-path-block.tsx
```diff
@@ -0,0 +1,212 @@
...
+import Spacings from '@commercetools-uikit/spacings';
+import { Theme, Interpolation, css, SerializedStyles } from '@emotion/react';
+import useSelectedPath from '../hooks/use-selected-path';
+
+type OneOrManyChildren = React.ReactElement | React.ReactElement[];
+type MultiPathBlockProps = {
+  children: OneOrManyChildren;
+};
+
+type LabelSyncPair = {
```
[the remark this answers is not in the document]
gabriele-ct: Yes, good point 👍 . I actually refactored a bit more. There's no need of having a stand alone `path-block.tsx`, as this component is super simple and it's used only in the `MultiPathBlock`

packages/ui-kit/src/components/multi-path-block.tsx
```diff
@@ -0,0 +1,212 @@
...
+      if (matchedSyncItem) {
+        setSelected(matchedSyncItem);
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedPath]);
+
+  useEffect(() => {
+    let activeIndex = Array.isArray(props.children)
+      ? props.children.findIndex(
```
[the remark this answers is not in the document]
gabriele-ct: Yes, probably it's more readable. Actually, I remember after implementing the original function I run it trough copilot asking to simplify it. I guess it was simpler in terms of line of codes but not very readable.

packages/ui-kit/src/components/multi-path-block.tsx
gabriele-ct: Yes, that makes the code more readable too

packages/ui-kit/src/hooks/use-selected-path.ts
gabriele-ct: 👍

packages/ui-kit/src/hooks/use-selected-path.ts
gabriele-ct: yes, that helps in making the condition clearer

packages/ui-kit/src/hooks/use-selected-path.ts
gabriele-ct: Thank you for the suggestion @ColinRosati .

If you have a look to LOC 65 I'm actually using the URLSearchParams api to handle the search parameters update. Using that also guarantees that any other previously exiting parameter in the url is kept while the only update happens to the `activePath` param. 

Regarding the usage of `replaceState`, AFAIK it's needed as the URLSearchParams API concern is just to manipulate the search parameters but not updating the actual browser url.

Finally have a look to the refactored function, it should guarantee that if any parameter !== 'activePath' exists in the url, remains untouched

gabriele-ct: I also added some "more meaningful" names for the `replaceState` parameters, however [the API signature itself it's pretty ugly](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState#parameters)

packages/ui-kit/src/hooks/use-selected-path.ts
gabriele-ct: I'm not sure what you mean here @ColinRosati.

The `updateSelectedPath` is part of the hook public API and allows to programmatically update the current selected path. This means performing 3 independent tasks:
1.  Update local storage `active_path`
2. Update the url `activePath` get parameter
3. Dispatch a custom event to update any other existing multi path components on the page.

Maybe I'm not getting your point here, we can discuss it in a meeting if you want

ColinRosati: Thanks for the response. That makes sense, we need all actions to happen

ColinRosati (approved):
LGTM! 👏

It would be nice to add some tests here especially for the synced across components functionality. We can do a follow up for those tests

PR #2070, opened 2024-09-05
Add support for union types in raml docs for api types

### LINKED ISSUE #1483: Add support for Union types in RAML docs
I have some union type to render from RAML that is encoded like this in the flattened RAML spec:
```yaml
properties:
  reference:
    type: CartReference | OrderReference
    (oneOf):
    - CartReference
    - OrderReference
    (builtinType): any
    description: |-
      Reference to a [Cart](ctp:api:type:Cart) or [Order](ctp:api:type:Order) that is replicated.
    required: true
    (inherited): false
 ```
I expected to see something like  "[CartReference]() or [OrderReference]()"  rendered for the `reference` property, but instead I get just "Union".
Union

Could you please add support for this for RAML types also? For query parameters we have something like that already:
union parameter

### PR DESCRIPTION
**Pull request by timonrey**
** Description of changes **
This PR refactores the pre processing of resource parameters and api props. It also adds support for union types for api types.

** Link original ticket **
closes #1483 

You can find an example here:
https://docskit-tt-refactor-type-rendering-for-types-and-endpoints.commercetools.vercel.app/api-docs-smoke-test/types/general#unionreferencetesttype

- [x] stakeholders approve changes? (if needed)

### CODE REVIEW

**Review by stmeissner — APPROVED** [review]
Very nice 👍🏻

File: packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js [hunk 1]
```diff
@@ -36,6 +73,20 @@ const NameType = (props) => {
   const isRegex = (string) =>
     string.charAt(0) === '/' && string.charAt(string.length - 1) === '/';
 
+  const isTypeUnion = (strType) => {
+    return typeof strType === 'string' && strType === 'Union';
+  };
```
gabriele-ct, line 78:
```
const isTypeUnion = (strType) => strType === 'Union'
```
It's safe to skip the typeOf check, even if it's `undefined` the check will still return `false`

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -57,11 +108,13 @@ const NameType = (props) => {
...
 
-        {isRegex(props.property.name) && !typeToRender.displayPrefix
+        {getParameterType(props.property, typeToRender.type, props.apiKey)}
+
+        {/* {isRegex(props.property.name) && !typeToRender.displayPrefix
           ? `Any ${typeof typeToRender.type} property matching this regular expression`
           : typeof typeToRender.type === 'string'
           ? typeToRender.type
-          : typeToRender.type}
+          : typeToRender.type} */}
```
gabriele-ct, line 117: I think we can safely get rid of this as it's been commented for quite a long time

gabriele-ct (changes requested):
Some little changes, but overall looks good

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -26,6 +27,42 @@ const BetaWrapper = styled.span`
...
+  const typesToRender = useTypesToRender({
+    property: props.types,
+    apiKey: props.apiKey,
+    isParameter: true,
+  });
+
+  return (
+    <>
+      Can be{' '}
+      {typesToRender.map(({ type }, idx, { length }) => {
```
ColinRosati, line 40: Minor suggestion. We can move the conditional into the body of the span so we have less repeat characters.

```
{typesToRender.map(({ type }, idx, { length }) => <span key={idx}>{length > idx + 1 ? `${type} ,` : `or ${type}`}</span>}
```

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
```diff
@@ -26,6 +27,42 @@ const BetaWrapper = styled.span`
...
+          <span key={idx}>{type}, </span>
+        ) : (
+          <span key={idx}>or {type}</span>
+        );
+      })}
+    </>
+  );
+}
+
+UnionParametersRow.propTypes = {
```
ColinRosati, line 51: Style suggestion to update this to a tsx file instead of writing more old react propTypes

packages/gatsby-transformer-raml/src/create-api-node.mjs
```diff
@@ -1,6 +1,5 @@
-import doRecursion  from "./utils/api/do-recursion.mjs";
-import parametersToArray  from "./utils/parameters-to-array.mjs";
-
+import doRecursion from './utils/api/do-recursion.mjs';
```
ColinRosati, line 1: Just a question. Why do we use mjs files? We should be safe to use .ts/.js to use modern import export syntax

ColinRosati (commented):
Looks good. Just a few minor suggestions

packages/gatsby-theme-api-docs/src/components/type/properties/rows/name-type.js
timonrey: We are currently still using js files for the api-docs components. We could definitely refactor the components to use typescript, but I would do it for the whole package and not just for a single file. Let's do this separately 👍🏼

packages/gatsby-transformer-raml/src/create-api-node.mjs
timonrey: Also an old convention that we have for our utils in the raml transformer package afaik. But in this case it probably still makes sense to use modules.

gabriele-ct: I think it's because we use ecmascript modules here, and probably the project default is to use commonjs modules. Using .mjs extension ensures node loads the files as ecmascript modules rather than commonjs modules

gabriele-ct (approved):

ColinRosati (approved):

PR #2076, opened 2024-09-10
Display multiple scopes in api endpoints in two columns

timonrey:
** Description of changes **

** Link original ticket **
closes https://github.com/commercetools/commercetools-docs/issues/4871

After:
Screenshot 2024-09-10 at 2 46 02 PM

Maintaining no column for less than 10 scopes and two columns for more than 10 scopes:
Screenshot 2024-09-10 at 2 46 10 PM

- [ ] user documentation added?
- [x] stakeholders approve changes? (if needed)

### DIFF
--- a/packages/gatsby-theme-api-docs/src/components/resource/method/scopes.tsx
+++ b/packages/gatsby-theme-api-docs/src/components/resource/method/scopes.tsx
@@ -2,13 +2,14 @@ import SpacingsStack from '@commercetools-uikit/spacings-stack';
 import { Markdown } from '@commercetools-docs/ui-kit';
 import Title from './title';
 import styled from '@emotion/styled';
+import { dimensions } from '@commercetools-docs/ui-kit/src/design-system';
 
 type ScopesProps = {
   scopes: string[];
 };
 
 const Container = styled.div`
-  @media (min-width: 768px) {
+  @media screen and (${dimensions.viewports.tablet}) {
     display: grid;
     grid-template-columns: repeat(2, 1fr);
     grid-gap: 0.5rem;
@@ -19,10 +20,12 @@ const Container = styled.div`
 `;
 
 const Text = styled(Markdown.InlineCodeWithoutBox)`
-  :not(:last-child)&::after {
-    content: ', ';
+  @media screen and (${dimensions.viewports.mobile}) {
+    :not(:last-child)&::after {
+      content: ', ';
+    }
   }
-  @media only screen and (min-width: 768px) {
+  @media screen and (${dimensions.viewports.tablet}) {
     &::after {
       content: ' ';
     }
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-api-docs/src/components/resource/method/scopes.tsx
```diff
@@ -0,0 +1,42 @@
...
+  grid-template-columns: repeat(2, 1fr);
+  grid-gap: 0.5rem;
+`;
+
+type ScopesProps = {
+  scopes: string[];
+};
+
+const Scopes = (props: ScopesProps) => {
+  const hasMultiColumnScope = props.scopes.length >= 10;
```
gabriele-ct, line 17: I'm happy with the change, however I think @nkuehn had mentioned not to use hard limits but rather use a CSS approach. Maybe he can clarify what he had in mind

nkuehn: In sloppy words, I meant "if you want columns make it columns".   It's a bit like "if you want a grid use a grid and not a table".   It only superficially works but all the layout decisions about edge case and responsive behavior will be more aligned with your intent if you use the CSS feature that is representing the intent instead of trying to emulate the behavior with other means. 

https://developer.mozilla.org/en-US/docs/Web/CSS/columns

gabriele-ct:
I think @ColinRosati  that on mobile viewports we agreed in the ticket to display the scopes in a single column
Screenshot 2024-09-11 at 07 52 34

@zbalek can you confirm this?
This is the [direct link ](https://docskit-tt-display-multiple-scopes-in-two-columns.commercetools.vercel.app/api-docs-smoke-test/endpoints/methods#post-resource)to the page

nkuehn:
FYI that was why I was asking for an actual coulumn layout.  No need for breakpoints, it automatically puts the number of columns that fit the space and distributes content so that columns are same height.

ColinRosati:
Thanks for the feedback @nkuehn. I have a mock up of using pure css columns and have some unintended limitations for the colums:
- Column cant controls the content inside the elements, so we don't guarantee a single scope is dedicated to one column and can span across multiple.
- We see poor layout for scopes with small amounts. The column align wont kick in until there is column element threshold is meant.   
- Cant control row gap. This make the scopes less readable

I explored some other config adding some additional spacing between text. But this appears to be suboptimal. 

This is the result:
Screenshot 2024-09-11 at 9 22 31 AM
Screenshot 2024-09-11 at 9 22 36 AM
Screenshot 2024-09-11 at 9 22 45 AM

The solution I propose removes the number of column controlling gap mentioned previously. The behaviour is now: 
Columns for desktops
Mobile single column (comma separated)

zbalek:
@gabriele-ct - Thanks for attaching the link. Let's not squish it on mobile - we can leave it as one column.

gabriele-ct:
Mobile view:
Screenshot 2024-09-11 at 11 44 36

Tablet/Desktop view with few elements:
Screenshot 2024-09-11 at 11 57 17


Tablet/Desktop with many elements:
Screenshot 2024-09-11 at 11 57 56

are we ok with this @zbalek

gabriele-ct (approved):
Thank you 👍

Issue #1606, opened 2023-04-05
Browser search results not highlighted in scroll bar


industrian:
I’m not sure if this is expected behavior or a bug, but I’ve noticed that browser search results (when using ctrl+f / command+f on a website) do not leave a mark in the scroll bar as they do on other websites and within Visual Studio Code.

In my opinion this is a useful feature that is missing from our docs.

gabriele-ct:
Results of the investigation:

- The search result highlights are visible only if the right scrollbar is visible. In our case we hide the y axis scrollbar by setting `overflow-y: hidden` and we leave `auto` the visibility of overflow-x.
- It turns out that the visibility of the y scrollbar is affected by the overflow-x setting meaning that we cannot have `overflow-x: hidden` and `overflow-y: auto` as it will still hide the y scrollbar.
- A solution could be to set both `overflow-x` or `overflow-y` to their defaults (auto) as I did in the linked PR. However I cannot guarantee that we won't get any sideeffects.

PR #2077, opened 2024-09-10
fix: scroll to anchor not working when navigating to a different page

gabriele-ct:
** Description of changes **
Fix the normal scroll to anchor behaviour which appeared to be broken by [this PR](https://github.com/commercetools/commercetools-docs-kit/pull/2031/files).
Reverting this change will disable the [highlighting of search results in the scrollbar](https://github.com/commercetools/commercetools-docs-kit/issues/1606).

I think at the moment is better to prioritise a core functionality such as the scrolling to anchor over the highlighting of search results in the scrollbar.

** Link original ticket **
Closes https://github.com/commercetools/commercetools-docs/issues/5484

The issue can be checked here:
https://docskit-ga-broken-anchor-scroll.commercetools.vercel.app/docs-smoke-test/views/links#link-to-another-page-on-this-site-with-a-specific-anchor-a-page-on-this-site

- [ ] user documentation added?
- [ ] stakeholders approve changes? (if needed)

### DIFF
--- a/.changeset/khaki-ghosts-know.md
+++ b/.changeset/khaki-ghosts-know.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-docs/gatsby-theme-docs': patch
+'@commercetools-website/docs-smoke-test': patch
+---
+
+Fix anchor navigation between pages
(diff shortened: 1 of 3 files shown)


FFawzy:
It sounds like you're dealing with a conflict between two behaviors: anchor scrolling and retaining the browser’s native in-page search highlights on the scrollbar. This is a tricky issue since manipulating overflow properties on the elements can directly impact both features.

maybe we should consider a hybrid approach of some sort having but to be investigated later lets get the fix out first 
Thanks @gabriele-ct

timonrey (approved):

PR #2080, opened 2024-09-12
[fix] Scopes layout patch: update import path 

ColinRosati:
**Description**
In previous PR an import path was incorrectly accessing files. We need to update the import path.


gabriele-ct (approved):
Thank you.

Issue #2065, opened 2024-08-28
Creation of three new microsites


lukpan:
The administrator learning path is going to be broken up into three new learning paths. The existing administrator content will be distributed across the three new learning paths, and new additional content added to it as well.

The names of the new learning paths are:
- Model your product catalog
- Price and discount your products
- Model your business structure

Can we therefore name the microsites:
- `learning-model-your-product-catalog`
- `learning-price-and-discount-your-products`
- `learning-model-your-business-structure`

Could we have three new microsites created for these learning paths please. They can remain hidden until we go live, but it would be good to have them living on the main branch so that we can push content to them more incrementally.

It would be great if this could happen within the next week or so, if possible...
Thank you team ET!

timonrey:
Link to the new microsite creation process documentation:
https://commercetools.atlassian.net/wiki/spaces/ET/pages/1134133271/Creating+a+new+microsite+in+the+docs

Link to the PR with the first two microsites:
https://github.com/commercetools/commercetools-docs/pull/5588

PR #2088, opened 2024-09-19
Introduce new `create-new-microsite` script and update site template

timonrey:
**Description of changes**
This PR adds a new script called `create-new-microsite` to create new microsite in the kit. It also updates the site template to bring it to the latest version of our site configuration.
Some maintenance and improvements were made as well.

**Link to original ticket**
is part of #2065


packages/gatsby-theme-docs/src/overrides/page-header-banner-area.js
```diff
@@ -1,8 +1,5 @@
 import React from 'react';
-import UserResearchBanner from '../components/user-research-banner';
 
 // A banner to be rendered in the top-right corner of a content page
-const PageHeaderBannerArea = () => {
-  return <UserResearchBanner />;
```
timonrey, line 6: Removing the user research banner since we don't need it anymore. We removed it in the docs repo a while ago already

websites/api-docs-smoke-test/gatsby-config.mjs
```diff
@@ -7,7 +7,6 @@ const config = {
   // https://www.gatsbyjs.com/docs/reference/release-notes/v2.30
   flags: {
     FAST_DEV: true,
-    PRESERVE_FILE_DOWNLOAD_CACHE: true,
```
timonrey, line 10: This flag is already included in the `FAST_DEV` flag.

websites/docs-smoke-test/gatsby-config.mjs
```diff
@@ -7,7 +7,7 @@ const shouldEnableSearch = process.env.ENABLE_SEARCH === 'true';
 
 const config = {
   flags: {
-    DEV_SSR: true,
```
timonrey, line 10: This one as well.

websites/documentation/gatsby-config.mjs
```diff
@@ -4,7 +4,6 @@ import dotenv from 'dotenv';
 dotenv.config({ path: `.env.${process.env.NODE_ENV}` });
 const config = {
   flags: {
-    DEV_SSR: true,
```
gabriele-ct, line 7: why remove the DEV_SSR?

websites/site-template/gatsby-config.mjs
```diff
@@ -13,21 +14,34 @@ const config = {
...
-    betaLink: '',
-    excludeFromSearchIndex: isProd,
-    allowWideContentLayout: true,
-    products: ['CHANGE-ME'],
-    contentType: 'CHANGE-ME',
+    betaLink: '/beta',
+    excludeFromSearchIndex: isProd && !shouldEnableSearch,
   },
   plugins: [
     ...configureThemeWithAddOns({
```
gabriele-ct, line 24: If I understand correctly you're re-writing the gatsby-config.mjs file at each microsite creation, so why even bother in adjusting this file? why not leave it how it was?

gabriele-ct (approved):
few notes, other than that it looks good

websites/site-template/gatsby-config.mjs
timonrey: Because the site template was outdated for a while already. I just adjusted the config to our current setup (like we do it in the other microsites).

websites/documentation/gatsby-config.mjs
timonrey: It's better to use the FAST_DEV flag. It includes the flags that we need
See here: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/flags.ts#L87

